PR #2, opened 2016-08-10
Fix ToJsonString

Roguelazer:
Right now, this will only encode strings up to the first escape character, and it won't encode anything at all if no escape characters are found.

That is to say, currently `"foo".to_owned().to_json_string()` returns `""`, and `"foo\" bar".to_owned().to_json_string()` returns `"foo"`.


aagahi:
Merged. ty

PR #4, opened 2016-08-11
make event fields work a little better

donhcd:
looks like sentry doesn't like loggers that have spaces in them. We should instead put the file/line in the "culprit" field. the sentry UI seems to like it when you use the culprit field, as it sticks its contents right below the name of the message.
screen shot 2016-08-11 at 2 22 14 pm


aagahi:
Merged. ty

PR #8, opened 2016-12-25
Add push_tag to Event (fix #7)

### LINKED ISSUE #1: Support for custom hosts
It seems like currently the default sentry host is hardcoded here: https://github.com/aagahi/rust-sentry/blob/master/src/lib.rs#L330

That should be an optional option 👍

### LINKED ISSUE #7: Allow specifying custom tags
Since `tags` is a private member of `Event`, and since `Event::new` doesn't have a `tags` argument, the only way to use custom tags is to manually create an instance of `Event`.

r-darwish:
This patch adds the ability to use custom tags. I wasn't sure whether I should add a new argument to `Event::new` or a new method to `Event`, so I ended up with a method. If you prefer otherwise then I'll change it. I also added a public method to posting an `Event` to Sentry and fixed the serialization of tags.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -256,6 +256,10 @@ impl Event {
             fingerprint: fingerprint.unwrap_or(vec![]),
         }
     }
+
+    pub fn push_tag(&mut self, key: String, value: String) {
+        self.tags.push((key, value));
+    }
 }
 
 impl ToJsonString for Event {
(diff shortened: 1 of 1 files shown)


aagahi:
Ok merged, seems ok, I'm going to add and perform some test since travis build seems broken on osx for unknown reason...

aagahi:
btw ty

PR #10, opened 2017-02-15
Use hyper-native-tls for https

fluxxu:
hyper 0.10 does not support HTTPS natively.
Related issue:
https://github.com/hyperium/hyper/issues/1009


aagahi:
ty

PR #11, opened 2017-03-15
Add constructor for SentryCredential from DSN

rschifflin:
This is a small ergonomic addition so that users don't have to decompose their DSN to fill in their credentials.


Cargo.toml
```diff
@@ -18,3 +18,4 @@ hyper = "0.10"
 hyper-native-tls = "0.2.2"
 time = "0.1"
 chrono = "0.2"
+regex = "0.2"
```
gevans: It might be better to use [url](https://crates.io/crates/url) for parsing DSNs instead.

aagahi:
Nice one thanks.

PR #12, opened 2017-03-15
Add settings constructor for Sentry that allows overriding Device information

rschifflin:
Provides a new constructor for Sentry that lets users be more flexible with settings. In particular this change allows the user to specify device-specific information under a new constructor, `from_settings`. Under the hood we move the Sentry settings to a struct and provide a builder for it, which defaults to reasonable values when a setting is omitted.


src/lib.rs
```diff
@@ -386,11 +405,65 @@ pub struct SentryCredential {
...
+
+pub struct SettingsBuilder {
+    server_name: Option<String>,
+    release: Option<String>,
+    environment: Option<String>,
+    device: Option<Device>
+}
+
+impl SettingsBuilder {
+    pub fn new() -> SettingsBuilder {
```
aagahi: I would have used the same approach as Device. A default function for default Settings and a simple struct with optional values. No need to have the with_ and build function I think... wdyt?

rschifflin: The builder here feels appropriate for a couple reasons. It lets us add new values to settings in the future without changing the existing public api, and it lets us have Settings objects that don't contain options so they're always fully-specified and easy to work with. Plus it'll be [familiar](https://aturon.github.io/ownership/builders.html) to rust users from other libraries that use multiple configurable and/or optional settings. (And personally I like the ergonomics of a builder over creating a mutable default config and modifying it over several statements). That's my two cents at least, we can go the other route if you prefer.

aagahi: Well I would prefer Settings that are set only once and no optionnal and default values as helper (just like Device). Settings are not changed at runtime...

aagahi:
ty

PR #13, opened 2017-04-10
update chrono & backtrace version

king6cong:


aagahi:
ok I've updated to 0.3 chrono and backtrace

PR #17, opened 2017-06-13
use error-chain for error-handling and don't panic on send

king6cong:


aagahi:
While I really like the idea of using error-chain in a project, I would prefer not to integrate in lib such as sentry to avoid having to deal with dependency issues. error-chain is cool but here I don't really see it as mandatory feature...

king6cong:
@aagahi Yeah, library should eliminate unnecessary dependencies, but error_chain is becoming the de facto error-handling mechanism in rust, while dealing and converting between different sources of errors causes lots of boilerplate code. Besides error-chain is clean on dependencies (only optional backtrace). Error handling is needed and panics should be avoided in essential libs such as sentry, please check the updated post method with two kinds of errors for the moment, what do you think?

aagahi:
Good point on the error-chain option dependency (let's hope they keep it that way).
So lgtm

king6cong:
Thanks

PR #20, opened 2017-08-07
upgrade error-chain

king6cong:


aagahi:
At least you should propose a stable version or the major/minor only. But the rc2... come on...

PR #22, opened 2017-08-29
upgrade to hyper 0.11

yjh0502:
- upgrade hyper 0.11, using async API.
 - use `serde` to serialize JSON. I know `serde` is quite huge, however I believe using standardized library is better than maintaining hand-written JSON serializer.
 - add `scheme` and `port` for `SentryCredential`, add support for `http` and custom port number.
 - remove `chrono` dependency. I believe `time::Time::to_utc()` and `tme::strfmt` can do the job.


aagahi:
ok for serde - I think it's hard to avoid it.

PR #39, opened 2018-04-28
fix: Corrently infer in_app for impl functions

jan-auer:
Impl functions do not start with the prefixes declared in `in_app_exclude`
and `in_app_include`. This PR introduces a new util `function_starts_with`
that accounts for differences in function naming.

**Example**:

Both functions will match `futures::`:

```
futures::task_impl::std::set
_<futures..task_impl..Spawn<T>>::enter::_{{closure}}
```


src/backtrace_support.rs
```diff
@@ -204,5 +204,57 @@ where
...
+pub fn function_starts_with(mut func_name: &str, pattern: &str) -> bool {
+    if func_name.starts_with("_<") {
+        func_name = &func_name[2..];
+    }
+
+    if !func_name.is_char_boundary(pattern.len()) {
+        return false;
+    }
+
+    func_name
```
jan-auer: Chose a manual zero-copy compare implementation here, but it needs `chars` which might be slow. Maybe we should use `as_bytes()` and then compare (and accept false positives)

PR #41, opened 2018-05-08
ci: Run clippy as linter

jan-auer:
Current nightly can't compiler one of our dependencies (or clippy).

### DIFF
--- a/src/scope/real.rs
+++ b/src/scope/real.rs
@@ -1,9 +1,9 @@
-use std::mem;
-use std::fmt;
-use std::thread;
-use std::cell::RefCell;
-use std::sync::{Arc, RwLock};
 use std::borrow::Cow;
+use std::cell::RefCell;
+use std::fmt;
+use std::mem;
+use std::sync::{Arc, RwLock};
+use std::thread;
 
 use api::protocol::{Breadcrumb, Context, User, Value};
 use client::Client;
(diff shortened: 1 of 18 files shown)


src/scope/real.rs
```diff
@@ -344,6 +344,7 @@ impl Scope {
     }
 
     /// Sets a tag to a specific value.
+    #[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
     pub fn set_tag<V: ToString>(&mut self, key: &str, value: V) {
```
jan-auer: Didn't want to change a public-facing API, but I think this suggestion is valid (as it will always copy).

src/transport.rs
```diff
@@ -85,20 +84,21 @@ impl Transport {
     pub fn new(dsn: Dsn, user_agent: String) -> Transport {
         let (sender, receiver) = sync_channel(30);
         let drain_signal = Arc::new(Condvar::new());
+        #[cfg_attr(feature = "cargo-clippy", allow(mutex_atomic))]
         let queue_size = Arc::new(Mutex::new(0));
```
jan-auer: Did not change this `Mutex`, as I'm pretty sure we want to block around `signal.notify_all()`

jan-auer:
@mitsuhiko This is ready for review now.

PR #45, opened 2018-06-04
feat: Move default contexts into client

mitsuhiko:


src/client/real.rs
```diff
@@ -296,13 +311,42 @@ impl Client {
...
+                        None => {
+                            match key.as_str() {
+                                "os" => add_os = false,
+                                "rust" => add_rust = false,
+                                "device" => add_device = false,
+                                _ => {},
+                            }
+                        }
+                        Some(ref value) => {
+                            event.contexts.insert((*key).clone(), (*value).clone());
```
jan-auer: This should not override existing contexts in the event.

jan-auer: Never mind, just saw that map does not override when calling `insert`. You could simplify this for the three other contexts then and just "attempt" an insert, instead of keeping three flags...

mitsuhiko: We can change that. Current behavior was already to override.

jan-auer:
@mitsuhiko do you care about that? Decided for simpler Code over efficiency here (duh), but was split in the first place. If you like, I can refactor

mitsuhiko:
Merge it.

PR #86, opened 2018-09-18
fix: Add new patterns for backtrace cleaning

jan-auer:
This PR adds new patterns for backtrace cleanup. Apparently, rust has changed the way it calls functions:

```diff
- _<sentry..integrations..log..Logger as log..Log>::log
+  <sentry::integrations::log::Logger as log::Log>::log

- _<futures..task_impl..Spawn<T>>::enter::_{{closure}}
+  <futures::task_impl::Spawn<T>>::enter::{{closure}}
```

Also, removed the `"_<"` matcher for `WELL_KNOWN_SYS_MODULES`, since this would either match with `"std::"` already (see the implementation of `is_sys_function`) or default to false in the end.

### DIFF
--- a/src/backtrace_support.rs
+++ b/src/backtrace_support.rs
@@ -20,9 +20,7 @@ lazy_static!{
             "sentry::",
             "sentry_types::",
             // these are not modules but things like __rust_maybe_catch_panic
-            // or _<T as core..convert..Into<U>>::into
             "__rust_",
-            "_<",
         ];
         #[cfg(feature = "with_failure")] {
             rv.push("failure::");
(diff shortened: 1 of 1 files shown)


jan-auer:
Also, regarding `<T as Into<U>` where U is in_app: This means that the implementing code is not in_app and therefore not interesting. That's why we can completely rely on the prefix matching we do right now.

jan-auer:
@mitsuhiko Also using `function_starts_with` for system frame detection now. As you can see, the implementation got more complicated so that we can match two cases:

 - If the pattern has leading `<` we also want to match `_<`
 - Otherwise, we just look for the namespace of the first name

I'm wondering if we should just drop support for old backtrace versions. I'd be surprised if there is still any code running that uses us but has old backtrace pinned.

PR #119, opened 2019-01-10
Combined panic/failure integration

Diggsey:
It would be useful to have a way to turn a `failure::Error` into a panic without losing the backtrace.

This is very common when you receive an error from further down the call-stack, and you don't have a good way to handle it, but currently you only get a callstack to the panic location, you lose the underlying error's backtrace.

This is just one possible way to implement the functionality.

### DIFF
--- a/src/integrations/failure.rs
+++ b/src/integrations/failure.rs
@@ -30,6 +30,7 @@
 use failure;
 use failure::{Error, Fail};
 use regex::Regex;
+use std::fmt::Display;
 
 use api::protocol::{Event, Exception, Frame, Level, Stacktrace};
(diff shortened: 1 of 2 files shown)


Diggsey:
@jan-auer not sure if you realised, but this was merged into `feature/detect-code`, not master.

jan-auer:
@Diggsey Thanks, I did not notice :) 
Did you want to add this to something or can I merge this in?

Diggsey:
@jan-auer I'm happy for you to do whatever with it - this is just our custom fork. If both features get merged in we can stop using the custom fork entirely.

PR #132, opened 2019-02-28
#130 Possibility to inject Reqwest client to Sentry through configuration.

mixalturek:
The new config option can be partially considered as a duplication of `transport` that allows to set a custom `TransportFactory`.


mixalturek:
There are two implementations and pull requests for this feature. Please read the following first.
https://github.com/getsentry/sentry-rust/issues/130#issuecomment-468278410

mixalturek:
#131 is the right approach.

PR #142, opened 2019-05-23
fix: SDK identifier

HazAT:


HazAT:
Credit goes to @lobsterkatie

PR #146, opened 2019-07-12
add custom_event_from_record

woodgear:


src/integrations/log.rs
```diff
@@ -182,7 +186,24 @@ impl log::Log for Logger {
     fn log(&self, record: &log::Record<'_>) {
         if self.options.create_issue_for_record(record) {
             Hub::with_active(|hub| {
-                hub.capture_event(event_from_record(record, self.options.attach_stacktraces))
+                let event = if let Some(custom_event_from_record) =
+                    &self.options.custom_event_from_record
+                {
+                    let mut event = custom_event_from_record(record);
+                    if self.options.attach_stacktraces && event.exception.is_empty() {
+                        event.exception = vec![Exception {
```
Swatinem, line 194: would be nice to extract this into a common function thats shared with `event_from_record`.
With that exposed, there is also no need to special case this here, we can just rely on the custom callback to attach the stacktrace.

src/integrations/log.rs
```diff
@@ -50,6 +51,8 @@ pub struct LoggerOptions {
     /// If set to `true` current stacktrace will be resolved and attached
     /// to each event. (expensive, defaults to `true`)
     pub attach_stacktraces: bool,
+    /// allow you to custom event,will respect attach_stacktraces is exception is empty
```
Swatinem, line 54:
```suggestion
    /// Allow you to customize event creation from a log `Record`.
```

Swatinem:
this is now possible with the redesigned log integration

PR #148, opened 2019-07-12
clippy: Switch to "cognitive_complexity" lint name

rye:
This was renamed upstream and the name `cognitive_complexity` can be used instead.  I think this will resolve the build errors that permeate the project at the moment.


jan-auer:
Hi, this was done on `master` in the meanwhile. Still, many thanks for the PR!

PR #170, opened 2019-11-27
Add an integration for std::error::Error

aiwenar:
The integration is based on `sentry::integrations::failure`, as modern standard error is very similar to `Fail`. The integration is gated behind `with_std_error` feature.

Currently there is no support for backtraces, as `std::backtrace` and `Error::backtrace()` are still unstable, but it could be implemented behind a feature gate: https://github.com/katalysteducation/sentry-rust/commit/665afb22693899a9c930a3af1d79411bd9439524. I decided not to include it in this PR because I don't know the policy Sentry has on unstable features, can't get it to work with the test harness (namely, `cargo +stable clippy --all-features ...`), and because I'm not sure how correct the backtrace parser is.


Swatinem:
closing, as this was added in #209

PR #174, opened 2020-01-07
write the panics section for functions that may panic

KeenS:
Functions that may panic should have the panics section


KeenS:
CI is failing but all of them are irrelevant to this PR...

PR #181, opened 2020-02-11
Fix clippy lint that is breaking CI

kornelski:


kornelski:
@mitsuhiko ping? The old pre-std-futures reqwest version is problematic and this project needs your attention.

jan-auer:
Thanks, including this in https://github.com/getsentry/sentry-rust/pull/183

PR #186, opened 2020-03-24
fix: Move reqwest init off the main thread

### LINKED ISSUE #184: 0.18.0 Release Issue on OSX
Get the following error:
```
thread 'main' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.', /Users/anthony.mittaz/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.11/src/runtime/enter.rs:19:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
 ERROR reqwest::blocking::client > Failed to communicate successful startup: Ok(())
```
I had no issue with the previous version: 0.17.0

Cargo file:

```
[package]
name = "segment-recover-bot"
version = "0.1.0"
authors = ["Anthony Mittaz <sync@me.com>"]
edition = "2018"
exclude = ["test/*"]

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
reqwest = { version = "0.10", features = ["json"] }
tokio = { version = "0.2", features = ["full"] }
lazy_static = "1.4.0"
url = { version = "2.1.1" }
serde = { version = "1.0", features = ["derive"] }
serde_json = "1.0"
regex = "1"
err-derive = "0.2"
sentry = "0.18.0"
chrono = { version = "0.4", features = ["serde"] }
either = { version = "1.5", features = ["serde"] }
futures-timer = "3.0"

warp = "0.2"
strum = "0.17.1"
strum_macros = "0.17.1"
log = "0.4"
pretty_env_logger = "0.4"
bytes = "0.5"
github_webhook_message_validator = "0.1.2"
hex = "0.4.2"
dotenv = "0.15.0"
graphql_client = "0.8.0"

[dev-dependencies]
httpmock = "0.3.4"
env_logger = "0.7"
```

Swatinem:
fixes #184

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,6 +67,7 @@ rustc_version = { version = "0.2.3", optional = true }
 failure_derive = "0.1.6"
 pretty_env_logger = "0.4.0"
 actix-web = { version = "0.7.19", default-features = false }
+tokio = { version = "0.2", features = ["full"] }
 
 [[example]]
 name = "error-chain-demo"
--- a/src/transport.rs
+++ b/src/transport.rs
@@ -11,11 +11,7 @@ use std::io::Cursor;
 use httpdate::parse_http_date;
 
 #[cfg(feature = "with_reqwest_transport")]
-use reqwest::{
-    blocking::{Client, ClientBuilder},
-    header::RETRY_AFTER,
-    Proxy,
-};
+use reqwest::{blocking::Client, header::RETRY_AFTER, Proxy};
 
 #[cfg(feature = "with_curl_transport")]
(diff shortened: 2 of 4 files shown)


Swatinem:
> I don't like the tokio `full` dependency, but since we depend on most of it already for reqwest, it's not too big of an issue.

Its a dev-dependency anyway, so not really such big of a deal. The bigger problem is actix pulling in a duplicate (older) version of tokio, and other dependencies -_-

Issue #188, opened 2020-04-02
Random segmentation fault when a panic is called in a tide request handler.


qdequele:
I had a segmentation fault when I tried to add Sentry to my project. https://github.com/meilisearch/meilisearch

I have noticed that the segmentation fault appears when I call a `panic!` into one of my [tide](https://github.com/http-rs/tide) request handler. 

To simplify the debug, I have created a new repository to create a minimal reproducible project. https://github.com/qdequele/test-sentry I also add an [lldb backtrace file](https://github.com/qdequele/test-sentry/blob/master/export-lldb.txt), but I explain in the readme all the step to reproduce the segmentation fault.

Apparently, it's from using the backtrace library.  I'll let you be the judge of that, but if I can help, I'd be happy to.

Swatinem:
Now that `backtrace` is using gimli by default, I think we can close this?

PR #189, opened 2020-04-03
update backtrace dependency and add gimli-symbolize feature flag

qdequele:
I have updated the backtrace dependency and add `gimli-symbolize` feature flag.
This feature flag seems to fix the issue https://github.com/getsentry/sentry-rust/issues/188.
I found how to fix it in the issue https://github.com/rust-lang/backtrace-rs/issues/150.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ with_rustls = ["reqwest/rustls-tls"]
 with_native_tls = ["reqwest/default-tls"]
 
 [dependencies]
-backtrace = { version = "0.3.44", optional = true }
+backtrace = { version = "0.3.46", optional = true, features = ["gimli-symbolize"] }
 url = { version = "2.1.1", optional = true }
 failure = { version = "0.1.6", optional = true }
 log = { version = "0.4.8", optional = true, features = ["std"] }


qdequele:
Friendly reminder ;) don't hesitate to tell me if I must modify something.

Swatinem:
@jan-auer had some concerns regarding the `gimli-symbolize` feature, care to explain?

jan-auer:
I'm afraid that gimli might not yet be fully ready for replacing `backtrace`, so I'd just wait for https://github.com/rust-lang/backtrace-rs/issues/189 to play out. Apart from that, it still adds quite some weight, so I'm not sure if I'd go with this by default.

We're in the process of splitting `sentry` up into multiple smaller crates. One of those might be backtrace support, and I think it may be a great idea to expose that as a feature. Alternatively, one can always include `backtrace = { features = "gimli-symbolize", ... }` in their Cargo.toml.

Swatinem:
> Alternatively, one can always include `backtrace = { features = "gimli-symbolize", ... }` in their Cargo.toml.

I will close this for now, as you can just enforce this feature on all other copies of that library.

PR #191, opened 2020-04-14
ref: Move sentry into a workspace member

Swatinem:


Makefile
```diff
@@ -20,17 +20,17 @@ check: style lint test
...
 format:
 	@rustup component add rustfmt --toolchain stable 2> /dev/null
-	cargo +stable fmt --all
+	cargo +stable fmt
 .PHONY: format
 
 lint:
 	@rustup component add clippy --toolchain stable 2> /dev/null
-	cargo +stable clippy --all-features --tests --all --examples -- -D clippy::all
+	cargo +stable clippy --all-features --tests --examples -- -D clippy::all
```
jan-auer, line 33: This should still contain `--all`.

Makefile
```diff
@@ -40,12 +40,12 @@ test: checkall testall
 
 testfast:
 	@echo 'TESTSUITE'
-	cargo test --features=with_test_support
+	cd sentry && cargo test --features=with_test_support
```
jan-auer, line 43: Can you use `cargo test -p sentry` here, instead? It would be nice to just run `cargo test` and have all workspace members tested, rather than selecting a package, because otherwise it'll be hard to grow this list later.

Makefile
```diff
@@ -40,12 +40,12 @@ test: checkall testall
...
 .PHONY: testfast
 
 testall:
 	@echo 'TESTSUITE'
-	cargo test --all-features --all
+	cargo test --all-features
```
jan-auer, line 48: Does this activate all features of all workspace members? This was intended to test all targets.

Swatinem:
I hope this answers all your questions at once:

* The root is now what cargo calls a *virtual manifest*, it only has workspace members but is not a package in itself.
* Cargo will implicitly use `--workspace` (which replaced the deprecated `--all`) for all commands
* Clippy inherits this implicit `--all` from cargo, so does rustfmt (though rustfmt does still use `--all` over `--workspace`)
* Specifying `--package` with either `--no-default-features` or `--features` raises the following error: 
  ```
  > cargo test --package sentry --no-default-features
  error: --no-default-features is not allowed in the root of a virtual workspace
  note: while this was previously accepted, it didn't actually do anything
  ```
  I think the reason is that you can use multiple `--package` options, and its hard to know where any feature belongs, I guess.

PR #192, opened 2020-04-15
ref: Create a sentry-core crate with hubs and scopes

Swatinem:


jan-auer:

PR #193, opened 2020-04-16
ref: Create the sentry-core crate

Swatinem:
This is a different try to #192, as this starts with a completely empty implementation, and explores the API more freely.

So far this is just a sketch of the public static and Hub API, with the following changes:
* `Client`/`Hub`/`Scope` are not explicitly wrappen in `Arc`, the plan instead is to use internal refcounting, or ownership transfer in case of `Scope`.
* Made `Uuid` return values consistent.
* `Hub::run` now works on `&self`.

### DIFF
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "sentry-core"
+version = "0.1.0"
+authors = ["Sentry <hello@sentry.io>"]
+license = "Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/getsentry/sentry-rust"
+homepage = "https://github.com/getsentry/sentry-rust"
+documentation = "https://getsentry.github.io/sentry-rust"
+description = """
+Core sentry library used for instrumentation and integration development.
+"""
+edition = "2018"
+
+[dependencies]
+sentry-types = "0.15.0"
--- a/sentry-core/src/api.rs
+++ b/sentry-core/src/api.rs
@@ -12,15 +12,15 @@ use crate::{Event, Hub, IntoBreadcrumbs, Level, Scope, Uuid};
 ///
 /// # Example
 ///
-/// ```
-/// use sentry_core::{Event, Level};
+/// ```should_panic
+/// use sentry_core::{Event, Level, Uuid};
 ///
 /// let uuid = Uuid::new_v4();
 /// let event = Event {
 ///     event_id: uuid,
 ///     message: Some("Hello World!".into()),
 ///     level: Level::Info,
-///     ..Default::default(),
+///     ..Default::default()
 /// };
 ///
 /// assert_eq!(sentry_core::capture_event(event.clone()), None);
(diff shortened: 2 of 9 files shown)


sentry-core/src/api.rs
```diff
@@ -0,0 +1,162 @@
...
+/// This function takes two arguments: the first is a callback that is passed
+/// a scope and can reconfigure it. The second is a callback that then executes
+/// in the context of that scope.
+///
+/// This is useful when extra data should be send with a single capture call,
+/// for instance a different level or tags:
+///
+/// # Example
+///
+/// ```compile_fail
```
jan-auer, line 130: What's the compile failure here?

sentry-core/src/hub.rs
```diff
@@ -0,0 +1,163 @@
...
+/// This can be used to capture events and manage the scope. This object is
+/// internally synchronized so it can be used from multiple threads if needed.
+/// The default hub that is available automatically is thread local.
+///
+/// See the
+/// [Unified API](https://docs.sentry.io/development/sdk-dev/unified-api/#hub)
+/// documentation for further details.
+///
+/// In most situations developers do not need to interface with the Hub
+/// directly. Instead, toplevel convenience functions are exposed that will
```
jan-auer, line 14: > In most situations developers do not need to interface with the Hub directly.

I've come to learn that this is not true. Especially with async/await or concurrent software, this is even the common case.

sentry-core/src/hub.rs
```diff
@@ -0,0 +1,163 @@
...
+/// automatically dispatch to the thread local (`Hub::current`) Hub. In some
+/// situations this might not be possible, in which case it might become
+/// necessary to manually work with the Hub. This is for instance the case when
+/// working with async code.
+///
+/// Hubs that are wrapped in `Arc`s can be bound to the current thread with
+/// the `run` static method.
```
jan-auer, line 21: I'm assuming that this no longer holds.

sentry-core/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+repository = "https://github.com/getsentry/sentry-rust"
+homepage = "https://github.com/getsentry/sentry-rust"
+documentation = "https://getsentry.github.io/sentry-rust"
+description = """
+Core sentry library used for instrumentation and integration development.
+"""
+edition = "2018"
+
+[dependencies]
+sentry-types = "0.15.0"
```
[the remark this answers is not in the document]
Swatinem: That would actually make a lot of sense!

sentry-core/src/api.rs
Swatinem: the Scope API does not exist yet

sentry-core/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "sentry-core"
+version = "0.1.0"
```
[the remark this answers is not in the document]
Swatinem: I still think that `sentry-core` and `sentry-types` for that matter should get to 1.0, and stay there. So that integration authors can rely on that version.
Most users will rather use the `sentry` facade, which might have more breaking changes. But I have no strong objections to synchronize the versions either.

jan-auer:

PR #194, opened 2020-04-17
ref: Add a Scope to sentry-core

Swatinem:
This is the next incremental step on top of #193


jan-auer:

PR #195, opened 2020-04-17
meta: Import sentry-types into the monorepo

Swatinem:
This imports the whole sentry-types repo, including all its history.
This was done using:
```
git remote add types https://github.com/getsentry/rust-sentry-types.git
git fetch types master

… move things into the sentry-types directory, delete everything unrelated …

git commit --all
git checkout -b types-head
git checkout master
git merge types-head --allow-unrelated-histories
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,4 +2,5 @@
 members = [
     "sentry",
     "sentry-actix",
+    "sentry-types",
 ]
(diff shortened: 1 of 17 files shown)


Swatinem:
> How would you think about bumping the version number of sentry-types to match the others? Also, since the changelog only contains one entry, we can probably remove it.

I think if we version everything at once, we only need one top-level changelog, and can remove all the rest.

PR #198, opened 2020-04-21
ref: Move the init function to sentry crate

Swatinem:
This is in preparation of moving the default options to the sentry
crate as well, which is needed to move the default transport and
integrations out of `sentry-core`.

Along with this, it removes some deprecated functions.

### DIFF
--- a/sentry/src/lib.rs
+++ b/sentry/src/lib.rs
@@ -111,5 +111,26 @@
 //! * `with_test_support`: Enables the test support module.
 #![warn(missing_docs)]
 
+#[macro_use]
+extern crate sentry_core;
+
+#[cfg(feature = "with_client_implementation")]
+mod init;
+
 #[doc(inline)]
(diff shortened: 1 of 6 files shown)


sentry/src/lib.rs
```diff
@@ -111,5 +111,26 @@
 //! * `with_test_support`: Enables the test support module.
 #![warn(missing_docs)]
 
+#[macro_use]
+extern crate sentry_core;
```
jan-auer, line 115: Instead of the macro use, can you just use a qualified import to `sentry_core::sentry_debug`? That may not work due to the inner macro invocation, so you may need to qualify this with `$crate::`.

Swatinem: nice! Still had to export that macro though, and also pull in `log` as another dependency. I would like to overhaul the complete logging at some point either way.

PR #201, opened 2020-04-28
 ref: Move transports out of sentry-core 

Swatinem:
This is on top of #199 and gets rid of some feature flags in sentry_core \o/


jan-auer:
Would be nice to make each transport it's own member crate.

PR #202, opened 2020-04-28
feat: Create a simple Integration trait

Swatinem:
This takes a lot of inspiration from, and obsoletes #99, and is built on top of #199

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -1,5 +1,7 @@
+use std::any::TypeId;
 use std::borrow::Cow;
 use std::fmt;
+use std::panic::RefUnwindSafe;
 use std::sync::Arc;
 use std::sync::RwLock;
 use std::time::Duration;
@@ -9,8 +11,9 @@ use rand::random;
 use crate::backtrace_support::process_event_stacktrace;
 pub use crate::clientoptions::ClientOptions;
 use crate::constants::SDK_INFO;
+use crate::integrations::Integration;
 use crate::internals::{Dsn, Uuid};
-use crate::protocol::{DebugMeta, Event};
+use crate::protocol::{ClientSdkInfo, DebugMeta, Event};
 use crate::scope::Scope;
 use crate::transport::Transport;
 use crate::utils;
@@ -20,10 +23,13 @@ impl<T: Into<ClientOptions>> From<T> for Client {
         Client::with_options(o.into())
     }
 }
+
 /// The Sentry client object.
 pub struct Client {
     options: ClientOptions,
     transport: RwLock<Option<Arc<dyn Transport>>>,
+    integrations: Vec<(TypeId, Arc<dyn Integration>)>,
+    sdk_info: ClientSdkInfo,
 }
 
 impl fmt::Debug for Client {
@@ -40,6 +46,8 @@ impl Clone for Client {
         Client {
             options: self.options.clone(),
             transport: RwLock::new(self.transport.read().unwrap().clone()),
+            integrations: self.integrations.clone(),
+            sdk_info: self.sdk_info.clone(),
         }
     }
 }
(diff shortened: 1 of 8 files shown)


sentry-core/src/integrations/mod.rs
```diff
@@ -15,3 +22,37 @@ pub mod env_logger;
...
+/// An Integration in sentry has two primary purposes.
+/// It can act as an *Event Source*, which will capture new events;
+/// or as an *Event Processor*, which can modify every `Event` flowing through
+/// the pipeline.
+// NOTE: we need `Any` here so that the `TypeId` machinery works correctly.
+pub trait Integration: Sync + Send + Any {
+    /// Name of this integration
+    ///
+    /// This will be added to the SDK information sent to sentry
+    fn name(&self) -> Option<&str> {
```
jan-auer, line 37: Can this be `&'static str`?

sentry-core/src/integrations/mod.rs
```diff
@@ -15,3 +22,37 @@ pub mod env_logger;
...
+        None
+    }
+
+    /// Called once the first time the integration is attached to a client
```
jan-auer, line 41: _nit_: In all of these, could you add a period at the end of the sentence?

sentry-core/src/integrations/mod.rs
jan-auer, line 37: What if we make this `fn name(&self) -> &'static str { std::any::type_name<Self>() }`

sentry-core/src/client.rs
```diff
@@ -57,12 +68,46 @@ impl Client {
...
+    where
+        I: Integration,
+    {
+        let integrations = self.integrations.read().unwrap();
+        let integration = integrations.get(&TypeId::of::<I>())?;
+        // We can‘t use the following code, because it requires a `'static`
+        // reference, and the reference has the same lifetime as `self`.
+        // (&integration.as_ref() as &dyn Any).downcast_ref()
+        // SAFETY: we just checked for the concrete type via TypeId
+        let integration = unsafe { &*(&**integration as *const dyn Integration as *const I) };
```
jan-auer, line 109: Potential solution for this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7dc36186ce572191e5872717b18f898b

sentry-core/src/integrations/mod.rs
```diff
@@ -15,3 +22,49 @@ pub mod env_logger;
...
+    fn name(&self) -> &str {
+        type_name::<Self>()
+    }
+
+    /// Called once the first time the integration is attached to a client.
+    fn setup_once(&self) {}
+
+    /// Called whenever the integration is attached to a Client.
+    #[cfg(feature = "with_client_implementation")]
+    fn setup(&self, options: &mut ClientOptions) {
```
jan-auer, line 46: This probably should be implemented even if `with_client_implementation` is not there. Otherwise, external implementors also need to add their own feature flag and always keep it in sync.

sentry-core/src/integrations/mod.rs
```diff
@@ -15,3 +22,49 @@ pub mod env_logger;
...
+pub trait Integration: Sync + Send + Any + AsAny {
+    /// Name of this integration.
+    ///
+    /// This will be added to the SDK information sent to sentry.
+    fn name(&self) -> &str {
+        type_name::<Self>()
+    }
+
+    /// Called once the first time the integration is attached to a client.
+    fn setup_once(&self) {}
```
jan-auer, line 42: Should we also pass client options here conservatively?

sentry-core/src/client.rs
```diff
@@ -73,15 +86,48 @@ impl Client {
...
-        Client { options, transport }
+        let mut sdk_info = SDK_INFO.clone();
+
+        let integrations: HashMap<_, _> = options
+            .integrations
+            .iter()
+            .map(|integration| (integration.as_ref().type_id(), integration.clone()))
+            .collect();
+
+        let mut global = GLOBAL_INTEGRATIONS.lock().unwrap();
```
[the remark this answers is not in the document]
Swatinem: valid point. originally, I was thinking to just push the responsibility into each of the integrations themselves, so they would maintain their own atomic bool, and just use `setup` alone.

sentry-core/src/integrations/mod.rs
Swatinem: I was thinking about just ripping out all this `with_client_implementation` nonsense. The problem here is with `ClientOptions`, but those seem to be independent of the client itself, so might as well expose that unconditionally.

sentry-core/src/clientoptions.rs
```diff
@@ -46,7 +47,11 @@ pub struct ClientOptions {
     pub in_app_include: Vec<&'static str>,
     /// Module prefixes that are never "in_app".
     pub in_app_exclude: Vec<&'static str>,
-    // TODO: Integration options
+    // Integration options
+    /// A list of integrations to enable.
+    pub integrations: Vec<Arc<dyn Integration>>,
```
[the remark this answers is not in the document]
Swatinem: hm, thats why we have the additional boolean flag, also as discussed with daniel, the default integrations are added on top inside of `init`.

I do have one concern regarding the ordering. We should have some way to guarantee the order of integrations, so that the *attach a stacktrace to events* integration runs before the *trim and flag stackframes* integration.

sentry-core/src/integrations/mod.rs
```diff
@@ -15,3 +22,46 @@ pub mod env_logger;
...
+/// An Integration in sentry has two primary purposes.
+/// It can act as an *Event Source*, which will capture new events;
+/// or as an *Event Processor*, which can modify every `Event` flowing through
+/// the pipeline.
+// NOTE: we need `Any` here so that the `TypeId` machinery works correctly.
+pub trait Integration: Sync + Send + Any + AsAny {
+    /// Name of this integration.
+    ///
+    /// This will be added to the SDK information sent to sentry.
+    fn name(&self) -> &str {
```
jan-auer, line 37: This is still `&str`, can we make this `&'static str` please? This is a little less constraining on users of this method.

sentry-core/src/integrations/mod.rs
jan-auer: Sounds good, let's remove the feature flag, then.

PR #203, opened 2020-04-28
ref: Move log integration into separate crate

Swatinem:
This is on top of #202 although it is independent from that PR.


sentry-core/src/client.rs
```diff
@@ -121,11 +115,28 @@ impl Client {
...
+            event = match integration.process_event(event) {
+                Some(event) => event,
+                None => {
+                    sentry_debug!("integration dropped event {:?}", id);
+                    return None;
+                }
+            }
+        }
+
+        if let Some(scope) = scope {
```
jan-auer, line 133: Hmm, I think this was intentionally before the other assignments to ensure that we override potentially invalid information. @HazAT what's the general order of applying integrations, scopes, internal stuff, etc.?

sentry-core/src/hub.rs
```diff
@@ -222,6 +224,49 @@ impl Hub {
...
+        I: Integration,
+        F: FnOnce(&I) -> R,
+        R: Default,
+    {
+        with_client_impl! {{
+            if let Some(client) = self.client() {
+                let integrations = client.integrations.read().unwrap();
+                if let Some(integration) = integrations.get(&TypeId::of::<I>()) {
+                    // SAFETY: we just checked for the concrete type via TypeId
+                    let integration = unsafe { &*(&**integration as *const dyn Integration as *const I) };
```
jan-auer, line 262: Since we checked, there's really no way this should fail in practice, so it would be great to avoid the `unsafe` block and just do a downcast. Gives us another `if let Some()` but at least it's safe.

sentry-core/src/client.rs
```diff
@@ -20,10 +23,12 @@ impl<T: Into<ClientOptions>> From<T> for Client {
         Client::with_options(o.into())
     }
 }
+
 /// The Sentry client object.
 pub struct Client {
     options: ClientOptions,
     transport: RwLock<Option<Arc<Box<dyn Transport>>>>,
+    pub(crate) integrations: RwLock<HashMap<TypeId, Arc<dyn Integration>>>,
```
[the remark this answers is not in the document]
jan-auer: For the record, I'm fine with not introducing another dependency like `anymap`, since this is really just a small shim.

sentry-log/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+documentation = "https://getsentry.github.io/sentry-rust"
+description = """
+Sentry Integration for log and env_logger crates
+"""
+edition = "2018"
+
+[dependencies]
+sentry-core = { version = "0.18.0", path = "../sentry-core", features = ["with_backtrace"] }
+log = { version = "0.4.8", features = ["std"] }
+env_logger = { version = "0.7.1", optional = true }
```
jan-auer, line 18: Should we make `env_logger` a separate thing? Otherwise, we'd have one log integration again with all different loggers as features.

sentry-core/src/client.rs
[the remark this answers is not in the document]
Swatinem: I think this is just used for the lookup on hub. I can make the lookup a separate `pub(crate)` method, and keep the member private. there is no mutation going on, so no need for add/remove.

sentry-core/src/client.rs
HazAT: Event > Scope > Integrations > Internal

Swatinem: Makes sense. It will break the assumption of the code comment above though. Do we want to look at debug meta in scope processors? I want to move those to integrations, so the order would be switched.

sentry-core/src/client.rs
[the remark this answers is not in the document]
jan-auer: > I can make the lookup a separate `pub(crate)` method

That sounds great!

sentry-core/src/clientoptions.rs
```diff
@@ -1,50 +1,68 @@
...
     /// The release to be sent with events.
     pub release: Option<Cow<'static, str>>,
     /// The environment to be sent with events.
     pub environment: Option<Cow<'static, str>>,
+    /// The sample rate for event submission. (0.0 - 1.0, defaults to 1.0)
+    pub sample_rate: f32,
+    /// Maximum number of breadcrumbs. (defaults to 100)
+    pub max_breadcrumbs: usize,
+    /// Attaches stacktraces to messages.
+    pub attach_stacktrace: bool,
```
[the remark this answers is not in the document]
Swatinem: > That could become an integration, maybe? Same with `trim_backtraces`.

I plan to do that ;-)
However, those options are documented in the official docs, and the backtrace processor would be a default integration anyway.

sentry-core/src/integrations/mod.rs
```diff
@@ -1,17 +1,48 @@
...
 #[cfg(feature = "with_panic")]
 pub mod panic;
+
+/// Integration abstraction.
+///
+/// An Integration in sentry has two primary purposes.
+/// It can act as an *Event Source*, which will capture new events;
+/// or as an *Event Processor*, which can modify every `Event` flowing through
+/// the pipeline.
+pub trait Integration: Debug + Sync + Send + 'static {
```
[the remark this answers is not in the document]
Swatinem: The Client / Options have a debug impl, which infects everything. We could make integrations opaque, or use the `name`, but inspecting any integration specific options provides real value. WDYT? I could think of a few workarounds, but they are not really pretty :-(

sentry-log/Cargo.toml
Swatinem: hm… `log` and `env_logger` are pretty tight AFAICS.

sentry-log/src/integration.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+}
+
+pub fn with_integration<F, R>(f: F) -> R
+where
+    F: FnOnce(&LogIntegration) -> R,
+    R: Default,
+{
+    Hub::with_active(|hub| hub.with_integration(f))
```
jan-auer, line 125: This looks like most integrations will need this -- should we add a shorthand to `Hub` for this?

Swatinem: I think I will do that in a followup, yes

sentry-log/src/logger.rs
```diff
@@ -4,14 +4,9 @@ use crate::converters::{breadcrumb_from_record, event_from_record};
 use crate::integration::with_integration;
 
 /// Provides a dispatching logger.
+#[derive(Default)]
 pub struct Logger;
 
-impl Logger {
-    pub fn new() -> Self {
```
jan-auer: TBH, having `new` even on types implement `Default` is not a bad idea. See `Vec` or `HashMap`.

Swatinem: 👍 I wrote that down on my list of additional cleanup tasks that I want to do after all the refactorings

PR #205, opened 2020-05-04
ref: Move Backtrace parsing out of failure integration

Swatinem:
In preparation to split things up further.


Swatinem:
> How would you feel about making this a crate again?

that will be the next incremental step ;-)
the backtrace trimming/classification still has the feature flags in it, thats why that needs to wait for other progress first.

PR #206, opened 2020-05-05
ref: Move error-chain integration out of core

Swatinem:


sentry-core/src/backtrace_support.rs
```diff
@@ -108,6 +105,7 @@ static ref SECONDARY_BORDER_FRAMES: Vec<(&'static str, &'static str)> = {
     "#).unwrap();
 }
 
+/// Parses a backtrace string into a Sentry Stacktrace
```
jan-auer, line 108:
```suggestion
/// Parses a backtrace string into a Sentry `Stacktrace`.
```

sentry-core/src/backtrace_support.rs
```diff
@@ -53,19 +54,15 @@ static ref WELL_KNOWN_BORDER_FRAMES: Vec<&'static str> = {
...
 };
 
 static ref SECONDARY_BORDER_FRAMES: Vec<(&'static str, &'static str)> = {
     #![allow(unused_mut)]
     let mut rv = Vec::new();
-    #[cfg(feature = "with_error_chain")] {
-        rv.push(("error_chain::make_backtrace", "<T as core::convert::Into<U>>::into"));
-    }
-    {rv}
+    // TODO: figure out how to expose these secondary frames to be configured
```
jan-auer, line 63: We could do this by adding an event processor that scans for the frame and then removes it. Basically make the internal logic reusable and move it to a processor instead of having it in the client.

sentry-core/src/backtrace_support.rs
```diff
@@ -159,6 +157,10 @@ fn parse_crate_name(func_name: &str) -> Option<String> {
         .map(|cr| cr.as_str().into())
 }
 
+/// Processes a stacktrace.
```
jan-auer, line 160: Same below. Either:
```suggestion
/// Processes a `Stacktrace`.
```

Or:
```suggestion
/// Processes a stack trace.
```

sentry-core/src/backtrace_support.rs
```diff
@@ -1,3 +1,7 @@
+//! Backtrace utilities for sentry
+//!
+//! Exposes functions to capture, process and convert/parse stacktraces
```
jan-auer, line 3: _nit_: Would be dope to convert the example into a doc test, WDYT? That way, we have a more self-explaining docs.rs page and it's automatically tested to compile etc as well. For now, there don't have to be any assertions.

sentry-core/src/backtrace_support.rs
Swatinem: Hm… maybe have a function which optionally trims just one match, then an integration can easily just chain two calls together. that would make things quite easy I think… Lets do that as a followup ;-)

sentry-core/src/backtrace_support.rs
Swatinem: I plan to move this out into its separate crate anyway, once all dependants are also moved out. I will keep this in mind ;-)

PR #207, opened 2020-05-07
ref: Move failure and panic integration out of core

Swatinem:
This also makes the PanicIntegration configurable with additional
extractors.


sentry-failure/src/lib.rs
```diff
@@ -49,6 +50,33 @@ fn fail_typename<F: Fail + ?Sized>(f: &F) -> (Option<String>, String) {
     }
 }
 
+#[derive(Default)]
+pub struct FailureIntegration;
```
jan-auer, line 54: Can we give all public types and functions a doc comment, and potentially a doc test? Also applies to the `sentry-panic` crate. We should specifically mention that this is a default integration.

sentry-failure/src/lib.rs
```diff
@@ -75,18 +103,20 @@ pub fn exception_from_single_fail<F: Fail + ?Sized>(
...
+        .enumerate()
+        .map(|(idx, cause)| {
+            let bt = match cause.backtrace() {
+                Some(bt) => Some(bt),
+                None if idx == 0 => Some(err.backtrace()),
+                None => None,
+            };
+            exception_from_single_fail(cause, bt)
+        })
+        .collect();
```
jan-auer, line 117: IMHO it was more readable before, but either way is fine.

sentry-panic/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+//!
+//! The panic integration can be configured with an additional extractor, which
+//! might optionally create a sentry `Event` out of a `PanicInfo`.
+//!
+//! ```
+//! let integration = sentry_panic::PanicIntegration::default()
+//!     .add_extractor(|info| None);
+//! ```
+//!
+//! Additionally panics are forwarded to the previously registered panic hook.
```
jan-auer, line 16:
```suggestion
//! Additionally, panics are forwarded to the previously registered panic hook.
```

sentry-panic/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::panic::{self, PanicInfo};
+use std::sync::Once;
+
+use sentry_core::backtrace_support::current_stacktrace;
+use sentry_core::protocol::{Event, Exception, Level};
+use sentry_core::{ClientOptions, Hub, Integration};
+
+/// A panic handler that sends to Sentry.
+///
+/// This panic handler report panics to Sentry. It also attempts to prevent
```
jan-auer, line 27:
```suggestion
/// This panic handler reports panics to Sentry. It also attempts to prevent
```

sentry-failure/src/lib.rs
jan-auer, line 54: Could give all Integrations a `new` constructor and favor that over using `default` when instantiating? See [`C-CTOR`](https://rust-lang.github.io/api-guidelines/predictability.html#constructors-are-static-inherent-methods-c-ctor):

> The name new should generally be used for the primary method of instantiating a type.

sentry-failure/src/lib.rs
```diff
@@ -49,6 +50,33 @@ fn fail_typename<F: Fail + ?Sized>(f: &F) -> (Option<String>, String) {
...
+        cfg.extra_border_frames.extend_from_slice(&[
+            "failure::error_message::err_msg",
+            "failure::backtrace::Backtrace::new",
+            "failure::backtrace::internal::InternalBacktrace::new",
+            "failure::Fail::context",
+        ]);
+    }
+}
+
+pub fn panic_extractor(info: &PanicInfo<'_>) -> Option<Event<'static>> {
```
jan-auer, line 72: For this method, a specific note that this is useful with the `PanicIntegration` would be great, along with some usage examples in a doc comment.

sentry-failure/src/lib.rs
Swatinem: I was trying to avoid the temporary `Vec`, but looks like the source iterator is not exact-sized, so it has no `.rev()` :-(

sentry-failure/src/lib.rs
```diff
@@ -69,6 +81,18 @@ impl Integration for FailureIntegration {
     }
 }
 
+/// Extracts a Sentry `Event` from a `PanicInfo`.
+///
+/// Creates a new Sentry `Event` when the panic has a `failure::Error` payload.
+/// This is for use with the `sentry-panic` integration, and iss enabled by
```
jan-auer, line 87:
```suggestion
/// This is for use with the `sentry-panic` integration, and is enabled by
```

sentry-panic/src/lib.rs
```diff
@@ -1,7 +1,9 @@
-//! Panic handler support.
+//! The Sentry Panic handler Integration.
 //!
-//! A panic handler can be installed that will automatically dispatch all errors
-//! to Sentry that are caused by a panic.
+//! The `PanicIntegration`, which is enabled by default in `sentry` installs a
```
jan-auer, line 3:
```suggestion
//! The `PanicIntegration`, which is enabled by default in `sentry`, installs a
```

PR #208, opened 2020-05-07
ref: Move backtrace support out of sentry-core

Swatinem:
Moves the `sentry_core::backtrace_support` into its own crate, and creates two new integrations, one that is responsible for attaching thread backtraces to events, the other one to trim and classify backtraces.


sentry-backtrace/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+mod utils;
+
+pub use crate::integration::{
+    current_thread, AttachStacktraceIntegration, ProcessStacktraceIntegration,
+};
+pub use crate::parse::parse_stacktrace;
+pub use crate::process::{backtrace_to_stacktrace, process_event_stacktrace};
+pub use crate::trim::trim_stacktrace;
+pub use crate::utils::error_typename;
+pub use sentry_core::protocol::{Frame, Stacktrace};
```
jan-auer, line 21: Out of curiosity, why are you re-exporting this? Users should always get this from `sentry::protocol`

Swatinem: Just for convenience, so the crate exposes all the non-std types it consumes/returns. Same reason that `sentry-types` re-exports a bunch of crono/uuid types I guess.

PR #209, opened 2020-05-11
feat: Add support for `std::error::Error`

### LINKED ISSUE #151: Lack of support for `std::error::Error`
I'm surprised that there's a convenience function for `failure::Error`, but not for the standard library equivalent.

`std::error` doesn't have a backtrace, but it does have a chain of `source()` errors. And the `source()` method supports downcasting, so it could automatically pick up `failure::Error` and others in the chain.

Swatinem:
This adds support directly to `sentry-core` instead of an integration.

fixes #151


sentry-core/src/error.rs
```diff
@@ -0,0 +1,61 @@
...
+    while let Some(err) = source {
+        exceptions.push(exception_from_error(err));
+        source = err.source();
+    }
+
+    exceptions.reverse();
+    Event {
+        exception: exceptions.into(),
+        level: Level::Error,
+        ..Default::default()
```
jan-auer, line 45: Can we capture a backtrace here on nightly?
See how [anyhow does it](https://github.com/dtolnay/anyhow/blob/2aa0cc1aba82a1ea513ba9e94d0077fd4d62479d/src/lib.rs#L201)

sentry-core/src/hub.rs
```diff
@@ -284,23 +286,20 @@ impl Hub {
         }}
     }
 
+    /// Capture any `std::error::Error`.
+    pub fn capture_error<E: Error + ?Sized>(&self, error: &E) -> Uuid {
+        let event = event_from_error(error);
+        self.capture_event(event)
+    }
```
jan-auer, line 293: Can we move this into `error.rs`, to keep all the error-related functions together? That way the file would look similar to integrations.

sentry-core/src/hub.rs
```diff
@@ -284,23 +286,20 @@ impl Hub {
...
-                } else {
-                    Uuid::nil()
-                }
-            })
-        }}
+        let mut event = Event {
+            message: Some(msg.to_string()),
+            level,
+            ..Default::default()
+        };
```
jan-auer, line 301: This runs now even if the client is disabled. Is this intended? Ideally, we would skip this if there is no client bound.

sentry-core/src/hub.rs
jan-auer: Can we guard this in a hub call or client to avoid work if the client is disabled? I'm realizing that this also applies to `capture_anyhow` and `capture_fail`.

sentry-core/src/utils.rs
```diff
@@ -2,6 +2,121 @@
...
+/// assert_eq!(parsed, (None, "Some".into()));
+/// ```
+pub fn parse_type_name_from_debug<D: std::fmt::Debug + ?Sized>(d: &D) -> (Option<String>, String) {
+    let dbg = format!("{:#?}", d);
+    parse_type_name(
+        dbg.split(&['(', '{', '\r', '\n'][..])
+            .next()
+            .unwrap()
+            .trim(),
+    )
```
jan-auer, line 100: Probably no need to pass this into `parse_type_name`, since `Debug` implementations never have module paths. You can just return the slice directly and can even avoid the allocation. If the caller needs a `String`, they can still call `to_owned`.

sentry-core/src/utils.rs
```diff
@@ -2,6 +2,121 @@
...
+///
+/// # Examples
+///
+/// ```
+/// use sentry_core::utils::parse_type_name;
+///
+/// let parsed = parse_type_name(std::any::type_name::<Vec<Option<usize>>>());
+/// assert_eq!(parsed, (Some("alloc::vec".into()), "Vec<core::option::Option<usize>>".into()));
+/// ```
+pub fn parse_type_name(mut type_name: &str) -> (Option<String>, String) {
```
jan-auer, line 15: Can this return borrowed strings? 

A few thoughts on this:

- Do we need the generics? IIRC this was just used to get the error name, in which case I'd even prefer without generics. 
- It seems that we strip off the module path, but not module paths of the generics. This seems inconsistent.
- If we really need generics, do we need `dyn`?
- Ideally this would just slice out the relevant part(s), but not allocate. If we need to allocate we could do that in the caller.

sentry-core/src/error.rs
Swatinem: not right now. I moved backtrace parsing out of core. I need to rethink how all this should fit together.

jan-auer: Alright, this is an important point to reconsider, then. Eventually you will want to also add backtraces, so maybe this needs to be changed considering that it's a valid use case to capture errors just with `sentry_core`, and expect backtraces.

sentry-core/src/utils.rs
```diff
@@ -2,119 +2,41 @@
...
-    parse_type_name(
-        dbg.split(&['(', '{', '\r', '\n'][..])
-            .next()
-            .unwrap()
-            .trim(),
-    )
+
+    dbg.split(&[' ', '(', '{', '\r', '\n'][..])
+        .next()
+        .unwrap_or(&dbg)
```
jan-auer, line 20: You can safely `unwrap` here, since split will always yield at least one element. If not, it might be more clear by signalling `unwrap_or("")`. Semantically, we would never want the full debug representation as the name.

Can we change the type signature to return a borrowed string and allocate in the caller?

sentry-core/src/utils.rs
```diff
@@ -2,119 +2,41 @@
...
-///
-/// ```
-/// use sentry_core::utils::parse_type_name_from_debug;
-///
-/// let parsed = parse_type_name_from_debug(&Some(1));
-/// assert_eq!(parsed, (None, "Some".into()));
-/// ```
-pub fn parse_type_name_from_debug<D: std::fmt::Debug + ?Sized>(d: &D) -> (Option<String>, String) {
+pub fn parse_type_from_debug<D: std::fmt::Debug + ?Sized>(d: &D) -> String {
     let dbg = format!("{:#?}", d);
```
jan-auer, line 16: This should probably use the standard debug print, not the alternate form. Otherwise, you might format the entire backtrace for an error (in the case of `failure`, for instance), which is considerably slower.

PR #212, opened 2020-05-14
feat: Add anyhow Integration

Swatinem:
Unfortunately, it does require a separate integration/method, since
`anyhow::Error` does not impl `std::error::Error`.

### DIFF
--- a/sentry-anyhow/src/lib.rs
+++ b/sentry-anyhow/src/lib.rs
@@ -1,0 +1,69 @@
+//! Adds support for capturing Sentry errors from `anyhow::Error`.
+//!
+//! # Example
+//!
+//! ```no_run
+//! # fn function_that_might_fail() -> anyhow::Result<()> { Ok(()) }
+//! use sentry_anyhow::capture_anyhow;
+//! # fn test() -> anyhow::Result<()> {
+//! let result = match function_that_might_fail() {
+//!     Ok(result) => result,
+//!     Err(err) => {
+//!         capture_anyhow(&err);
+//!         return Err(err);
+//!     }
+//! };
+//! # Ok(()) }
+//! ```
+
+#![deny(missing_docs)]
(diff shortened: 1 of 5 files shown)


sentry/examples/anyhow-demo.rs
```diff
@@ -0,0 +1,19 @@
+fn execute() -> anyhow::Result<usize> {
+    let parsed = "NaN".parse()?;
+    Ok(parsed)
+}
+
+fn main() {
+    let _sentry = sentry::init((
+        "https://a94ae32be2584e0bbd7a4cbb95971fee@sentry.io/1041156",
```
flub, line 8: Which DSN is this?

jan-auer: Let's actually back out all DSNs. We could configure via env_var if we want to test this, or just put a placeholder in.

sentry-anyhow/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use sentry_core::internals::Uuid;
+use sentry_core::Hub;
+
+/// Captures an error chain.
+pub fn capture_anyhow(e: &anyhow::Error) -> Uuid {
+    Hub::with_active(|hub| hub.capture_anyhow(e))
+}
+
+/// Hub extension methods for working with anyhow.
```
jan-auer, line 34: _nit_: Put `anyhow` in code quotes.

```suggestion
/// Hub extension methods for working with `anyhow`.
```

sentry/examples/anyhow-demo.rs
Swatinem: Let me remove that from all other examples as a followup.

jan-auer:
What's going on with CI?

Swatinem:
> What's going on with CI?

The base is a different PR, and builds don‘t run for those -_-

PR #213, opened 2020-05-15
ref: Move debug_meta into a separate package

Swatinem:
This creates a new `DebugImagesIntegration`, which can be configured
with an optional filter. By default, *all* events will get a list of
loaded libraries attached, depending on OS support.


sentry-debug-images/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+#![deny(missing_docs)]
+#![deny(unsafe_code)]
+
+use sentry_core::protocol::DebugImage;
+
+#[cfg(unix)]
+mod unix;
+
+#[cfg(unix)]
+use unix::find_shlibs;
```
jan-auer, line 28: Do you mean to `use debug_images` here?

sentry-debug-images/src/integration.rs
```diff
@@ -0,0 +1,52 @@
+use std::borrow::Cow;
+
+use sentry_core::protocol::{DebugMeta, Event};
+use sentry_core::{ClientOptions, Integration};
+
+use crate::debug_images;
```
jan-auer, line 6: _nit_: Rather than importing, can you just use the qualified path at call site? We usually don't import functions.

sentry-debug-images/src/lib.rs
Swatinem: 🤦

PR #214, opened 2020-05-15
feat: Add support for std::future::Future

Swatinem:
Adds a new `SentryFuture` type, which activates a hub for the duration
of a `poll`.


sentry-core/src/futures.rs
```diff
@@ -0,0 +1,96 @@
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::{Context, Poll};
+
+use crate::Hub;
+
+/// A future that bind a `Hub` to its execution.
```
flub, line 8:
```suggestion
/// A future that binds a `Hub` to its execution.
```

sentry-core/src/futures.rs
```diff
@@ -0,0 +1,96 @@
+use std::future::Future;
```
flub, line 1: Might be worth adding a module-level docstring with the common example of using `FutureExt::bind_hub`?

sentry-core/src/futures.rs
```diff
@@ -0,0 +1,96 @@
...
+impl<F> Future for SentryFuture<F>
+where
+    F: Future,
+{
+    type Output = F::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        let hub = self.hub.clone();
+        // https://doc.rust-lang.org/std/pin/index.html#pinning-is-structural-for-field
+        let future = unsafe { self.map_unchecked_mut(|s| &mut s.future) };
```
jan-auer, line 34: At first I was thinking it would be nice to use [`futures::pin_mut`](https://docs.rs/futures/0.3.5/futures/macro.pin_mut.html), but since we don't have a dependency on `futures` yet, it's probably not worth it.

Swatinem: I don‘t think its worth pulling in the dependency for that. Also, I remember I read about not being able to essentially implement future combinators without unsafe code before. But I’m surprised there is no better safe alternative apparently?

sentry-core/src/futures.rs
Swatinem: I hope I can clean up all the docs etc in a followup.

PR #215, opened 2020-05-15
ref: Clean up the feature flags a bit

Swatinem:
Cleans up the feature flags of `sentry-core`, but tries to keep the existing feature flags of `sentry`, for backwards compatibility. Not quite sure if we want to do that, or rather just remove the non-default features, and rename the others.

### DIFF
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -16,10 +16,10 @@ edition = "2018"
 all-features = true
 
 [features]
-default = ["with_client_implementation"]
-with_client_implementation = ["im", "url", "rand"]
-with_debug_to_log = ["log"]
-with_test_support = []
+default = []
+client = ["im", "url", "rand"]
+debug-to-log = ["log"]
+test = []
 
 [dependencies]
 url = { version = "2.1.1", optional = true }
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -18,7 +18,9 @@ all-features = true
 [features]
 default = []
 client = ["im", "rand"]
-debug-to-log = ["log"]
+# I would love to just have a `log` feature, but this is used inside a macro,
+# and macros actually expand features where they are used!
+debug-logs = ["log"]
 test = []
 
 [dependencies]
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -17,23 +17,19 @@ all-features = true
 
 [features]
 default = []
-client = ["im", "url", "rand"]
+client = ["im", "rand"]
 debug-to-log = ["log"]
 test = []
 
 [dependencies]
-url = { version = "2.1.1", optional = true }
-log = { version = "0.4.8", optional = true, features = ["std"] }
-sentry-types = { path = "../sentry-types", version = "0.15.0" }
-env_logger = { version = "0.7.1", optional = true }
+sentry-types = { version = "0.15.0", path = "../sentry-types" }
 lazy_static = "1.4.0"
 im = { version = "14.2.0", optional = true }
 rand = { version = "0.7.3", optional = true }
+log = { version = "0.4.8", optional = true, features = ["std"] }
 
 [dev-dependencies]
-pretty_env_logger = "0.4.0"
 thiserror = "1.0.15"
 anyhow = "1.0.30"
 failure = "0.1.8"
-sentry-failure = { version = "0.18.0", path = "../sentry-failure" }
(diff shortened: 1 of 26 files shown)


sentry-core/Cargo.toml
```diff
@@ -16,23 +16,19 @@ edition = "2018"
...
 rand = { version = "0.7.3", optional = true }
+log = { version = "0.4.8", optional = true, features = ["std"] }
 
 [dev-dependencies]
-pretty_env_logger = "0.4.0"
 thiserror = "1.0.15"
 anyhow = "1.0.30"
-failure = "0.1.8"
-sentry-failure = { version = "0.18.0", path = "../sentry-failure" }
+failure = "0.1.8"
```
jan-auer, line 34: _nit_: Add a newline.

sentry-core/src/macros.rs
```diff
@@ -55,10 +44,10 @@ macro_rules! with_client_impl {
 macro_rules! sentry_debug {
     ($($arg:tt)*) => {
         $crate::with_client_impl! {{
-            #[cfg(feature = "with_debug_to_log")] {
+            #[cfg(feature = "log")] {
```
jan-auer, line 47: Shouldn't this be `debug-to-log`?

sentry-core/Cargo.toml
```diff
@@ -16,23 +16,19 @@ edition = "2018"
 all-features = true
 
 [features]
-default = ["with_client_implementation"]
-with_client_implementation = ["im", "url", "rand"]
-with_debug_to_log = ["log"]
-with_test_support = []
+default = []
+client = ["im", "rand"]
+debug-to-log = ["log"]
```
jan-auer, line 21: _nit_: Maybe even shorter `debug-logs`? :)

I think we could remove this feature and just use the `log` feature from the optional dependency. The only reason not to do that would be in the top-level `sentry` crate, if we want to distinguish between `log` support and the internal debug logging. 

In the rest of the code base, it seems that you're mostly using the `log` feature for this already.

sentry-core/src/api.rs
```diff
@@ -167,9 +148,5 @@ where
 
 /// Returns the last event ID captured.
 pub fn last_event_id() -> Option<internals::Uuid> {
```
jan-auer, line 150: _nit_: Unrelated to this change, could you clarify in the docs that this returns the last event id from the current execution context, i.e. the active Hub?

sentry-core/src/lib.rs
```diff
@@ -79,37 +79,31 @@
...
 mod constants;
 mod error;
+mod hub;
+mod integration;
 mod intodsn;
-mod transport;
-
-#[cfg(any(test, feature = "with_test_support"))]
+mod scope;
+#[cfg(any(test, feature = "test"))]
```
jan-auer, line 104: Would you mind adding the blank line again and moving the `test` module into its own group? Otherwise, it's too easy to mess up the feature flags. Especially, since it's also a public module, it should go in its own group.

sentry/src/defaults.rs
```diff
@@ -33,18 +33,33 @@ pub fn apply_defaults(mut opts: ClientOptions) -> ClientOptions {
...
             ));
         }
-        integrations.push(Arc::new(sentry_contexts::ContextIntegration::default()));
+        #[cfg(feature = "debug-images")]
+        {
+            integrations.push(Arc::new(
+                sentry_debug_images::DebugImagesIntegration::default(),
+            ))
+        }
+        #[cfg(feature = "sentry-error-chain")]
```
jan-auer, line 48: I think it would be nicer to have just an `error-chain` feature here.

sentry-core/Cargo.toml
Swatinem: This is a bit of a 🤦‍♂️ moment. cfg is not really hygienic in macros, which means those features need to have the same name across core and sentry.

jan-auer: I see, in that case, probably having one consistent feature with a name distinct from `log` would be better.

sentry/Cargo.toml
```diff
@@ -17,27 +17,61 @@ autoexamples = true
 all-features = true
 
 [features]
-default = ["with_client_implementation", "with_default_transport", "with_panic", "with_failure"]
+default = ["client", "backtrace", "contexts", "failure", "panic", "transport"]
```
flub, line 20: failure being default kind of surprises me a little.  But I don't mind really.

Swatinem: It is currently the default. I don’t want to break that just now, rather wait another release cycle to change the defaults.

PR #216, opened 2020-05-18
ref: Restructure and document the public core API

Swatinem:
The `sentry-core` exports have been a bit restructured, other things in
`sentry` have been marked as deprecated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,3 @@ members = [
     "sentry-slog",
     "sentry-types",
 ]
-
-[profile.dev.build-override]
-opt-level = 0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,6 @@ members = [
     "sentry-slog",
     "sentry-types",
 ]
+
+[profile.dev.build-override]
+opt-level = 0
--- a/sentry-core/src/api.rs
+++ b/sentry-core/src/api.rs
@@ -43,6 +43,19 @@ pub fn capture_event(event: Event<'static>) -> Uuid {
 ///
 /// This creates an event from the given message and sends it via
 /// [`capture_event`](fn.capture_event.html).
+///
+/// # Examples
+///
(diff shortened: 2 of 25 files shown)


Cargo.toml
```diff
@@ -13,3 +13,6 @@ members = [
     "sentry-panic",
     "sentry-types",
 ]
+
+[profile.dev.build-override]
+opt-level = 0
```
flub, line 18: m: This is for all devel builds?  why?

sentry-core/src/api.rs
```diff
@@ -1,44 +1,70 @@
...
+/// assert_eq!(sentry::capture_event(event.clone()), Uuid::nil());
+///
+/// let events = sentry::test::with_captured_events(|| {
+///     assert_eq!(sentry::capture_event(event), uuid);
 /// });
+/// assert_eq!(events.len(), 1);
 /// ```
-pub fn capture_event(event: Event<'static>) -> internals::Uuid {
+///
+/// [`capture_message`]: fn.capture_message.html
```
flub, line 36: Pretty sure you don't need to link to html anymore, perhaps shouldn't even.
https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html#additions-to-the-documentation-syntax and the next "Implied Shortcut Reference Links" section detail this a bit better.  But basically type a valid rust path to the thing you refer between the square brackets.

flub: oh, ignore this.  you've just told me this is nightly-only :(

Cargo.toml
Swatinem: https://doc.rust-lang.org/cargo/reference/profiles.html#overrides
I haven’t actually measured this, but its supposed to speed up building all the proc-macro stuff. On the other hand, I haven’t actually done any `--release` builds anyway, so I have no idea how slow we are for those.

PR #218, opened 2020-05-21
feat: Add support for surf HTTP client

feymartynov:
This PR provides support of [surf](https://github.com/http-rs/surf) HTTP client as an alternative to reqwest and curl.

### DIFF
--- a/sentry/src/transport.rs
+++ b/sentry/src/transport.rs
@@ -4,7 +4,11 @@ use std::sync::{Arc, Condvar, Mutex};
 use std::thread::{self, JoinHandle};
 use std::time::{Duration, SystemTime};
 
-#[cfg(any(feature = "with_reqwest_transport", feature = "with_curl_transport"))]
+#[cfg(any(
+    feature = "with_reqwest_transport",
+    feature = "with_curl_transport",
+    feature = "with_surf_transport"
+))]
 use httpdate::parse_http_date;
 
 #[cfg(feature = "with_curl_transport")]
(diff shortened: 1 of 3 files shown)


feymartynov:
> Your build failures are coming from a duplicated Client import due to the feature flag mess.
I see. Will fix that.

> how heavy is surf in terms of dependencies & compile times?
It's comparable to reqwest. The difference is that reqwest is based on tokio while surf is based on async-std.

By the way it would be nice to switch to async/await someday because reqwest supports it since 0.10 and surf is initially async (currently I had to wrap it into `block_on` which is so-so).

Swatinem:
> it would be nice to switch to async/await someday

Yes, I would very much love to. However:

* We don’t want to force that on our users, so that would need to be restricted to a background thread either way.
* I’m not quite sure if some of the requests need a specific ordering. I will check with other team members.
* We would want to add a `flush` method to the transport trait in the future. Just to keep in mind that we would have to make it work somehow. With sync code, you would just create a oneshot channel, send it to the background thread and wait for the answer with a timeout. Or I think that should work quite straight forward ;-)
* On the other hand async makes the "forcibly close the transport" case much easier, since you can just stop polling the futures and return right away.

Just some food for thought ;-)

In any case, I really appreciate this! So to make this more future proof, maybe at least move it into a different file for now, and we can decide to create a sub-crate for it later on maybe.

sentry/src/transport.rs
```diff
@@ -376,15 +402,130 @@ implement_http_transport! {
...
+        let dsn = options.dsn.clone().unwrap();
+        let user_agent = options.user_agent.to_string();
+        let mut disabled = None::<SystemTime>;
+
+        thread::Builder::new()
+            .name("sentry-transport".to_string())
+            .spawn(move || {
+                sentry_debug!("spawning surf transport");
+                let http_client = http_client;
+                let url = dsn.store_api_url().to_string();
```
Swatinem, line 429: You are doing a `to_string()` here, and a `as_str()` later on. But I see the other transports do that as well.

Issue #220, opened 2020-05-27
Panic + deadlock when using LLD 10.0.0


Jake-Shadle:
I'm not sure when this was changed, so it might apply to more LLD's than just 10.0.0, but it appears that LLD now uses xxhash to compute the build ID stored in `.note.gnu.build-id` instead of ld's usual sha1. This causes a panic [here](https://github.com/getsentry/sentry-rust/blob/b66352a0a30a5da5f6aef0f25e29f8024e7bc3e0/sentry-debug-images/src/unix.rs#L17), as the section is only 8 bytes instead of 20, so the slice panics, and since this code is usually already being called from a panic handler, we end up in a deadlock waiting for the `Once` in `Client::capture_event`.

Jake-Shadle:
For now I'm working around this by forcing LLD to use sha1 when calculating the build id, as this probably won't be a simple fix.

```ini
[target.x86_64-unknown-linux-gnu]
rustflags = [
    "-C", "link-arg=-fuse-ld=lld",
    "-C", "link-arg=-Wl,--build-id=sha1",
]
```

Swatinem:
Thats interesting. We basically do the same on native:

https://github.com/getsentry/sentry-native/blob/9871bf8109a23850d29ff20dfb97c5bffbaf9446/src/modulefinder/sentry_modulefinder_linux.c#L287

Just with the difference that we do an unsafe read, because yolo, its C…

I think @jan-auer knows more about how to handle this, since our whole pipeline would need to handle these IDs.

Jake-Shadle:
It might be that everyone else is correctly [reading your docs](https://docs.sentry.io/workflow/debug-files/#gnu-build-identifiers) and explicitly specifying the flag when using clang/lld, and I just got lucky. :wink:

Swatinem:
Either way, we shouldn’t be panicking (or reading invalid memory) when the ID in the header is <16 bytes. Rather pad it with 0 or something. Not sure what the best idea would be in this case.

jan-auer:
It needs to be padded with 0 to comply with breakpad. I can get a fix for this going.

Jake-Shadle:
Thanks!

PR #221, opened 2020-05-28
fix(debug-images): Compute correct DebugId for ELF images

jan-auer:
This ports over the logic from [`symbolic`](https://github.com/getsentry/symbolic/blob/4d509d112f385daa1ec11a78e4b815e54b0dd3b6/debuginfo/src/elf.rs#L383) to compute `DebugId` from the GNU build identifier. This fixes two distinct bugs:

 1. A panic if the identifier is smaller than 16 bytes (Fixes https://github.com/getsentry/sentry-rust/issues/220)
 2. Incorrect endianness on big-endian targets.

### DIFF
--- a/sentry-debug-images/src/unix.rs
+++ b/sentry-debug-images/src/unix.rs
@@ -1,9 +1,37 @@
 use std::env;
 
-use findshlibs::{Segment, SharedLibrary, SharedLibraryId, TargetSharedLibrary, TARGET_SUPPORTED};
 use sentry_core::protocol::debugid::DebugId;
 use sentry_core::protocol::{DebugImage, SymbolicDebugImage};
 use sentry_core::types::Uuid;
+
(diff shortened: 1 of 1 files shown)


sentry-debug-images/src/unix.rs
```diff
@@ -1,10 +1,38 @@
...
+
+/// Converts an ELF object identifier into a `DebugId`.
+///
+/// The identifier data is first truncated or extended to match 16 byte size of
+/// Uuids. If the data is declared in little endian, the first three Uuid fields
+/// are flipped to match the big endian expected by the breakpad processor.
+///
+/// The `DebugId::appendix` field is always `0` for ELF.
+fn debug_id_from_build_id(build_id: &[u8]) -> Option<DebugId> {
+    let mut data = [0 as u8; UUID_SIZE];
```
jan-auer, line 19: Ideally, we could use `uuid::Bytes` here, and get rid of the constant, but that's not exposed from `sentry-types`.

PR #222, opened 2020-05-28
fix(debug-images): Use first PT_LOAD to obtain load address

jan-auer:
Aligns the Rust SDK with the fix from https://github.com/getsentry/symbolic/pull/148. The first `PT_LOAD` program header declares the load address of an image, regardless of whether it is marked as executable or not.


jan-auer:
@Swatinem if you choose to backport the other fix to `0.18`, then this is also a good candidate.

Swatinem:
> then this is also a good candidate.

👍 lets see how far we can get with pushing out the new release.

jan-auer:
Hold off, I think this logic is still broken...

jan-auer:
The proper fix has to go into `findshlibs`, since currently there are not enough functions exposed to do this properly. See https://github.com/gimli-rs/findshlibs/pull/55

Jake-Shadle:
Version 0.8.0 of `findshlibs` was just published that includes the PR made by @jan-auer, would be great to get this in and released to get correclty resolved stack traces in sentry.

jan-auer:
Thanks @Jake-Shadle for following up on this! @Swatinem will run the release.

PR #224, opened 2020-06-02
fix: Better document default integrations

Swatinem:
This also exposes a new global `with_integration` function which is also
used in existing integrations.

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -103,18 +103,13 @@ impl Client {
 
         let mut sdk_info = SDK_INFO.clone();
 
-        let mut integrations = vec![];
-        for integration in &options.integrations {
-            let id = integration.as_ref().type_id();
-            match integrations.iter().position(|(iid, _)| *iid == id) {
-                Some(idx) => {
-                    integrations[idx].1 = integration.clone();
-                }
-                None => {
-                    integrations.push((id, integration.clone()));
-                }
-            }
-        }
+        // NOTE: We do not filter out duplicate integrations based on their
+        // TypeId.
+        let integrations: Vec<_> = options
+            .integrations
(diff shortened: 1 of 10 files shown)


sentry-core/src/client.rs
```diff
@@ -103,13 +103,18 @@ impl Client {
...
-        // TypeId.
-        let integrations: Vec<_> = options
-            .integrations
-            .iter()
-            .map(|integration| (integration.as_ref().type_id(), integration.clone()))
-            .collect();
+        let mut integrations = vec![];
+        for integration in &options.integrations {
+            let id = integration.as_ref().type_id();
+            match integrations.iter().position(|(iid, _)| *iid == id) {
```
jan-auer, line 109: This is the point where creating a dedicated `Integrations` type would pay off. Internally, this could use a `BTreeSet` or `BTreeMap`.

Swatinem:
> Of course, `with_integration` would then only return one of them, but potentially there is no need to access such integrations later on.

One of the obvious usecases is to customize the default `PanicIntegration` with a custom extractor. Since that relies on the `with_integration` ordering, it would never use the custom extractor, unless you explicitly use `default_integrations: false`, and manually add all the defaults.

On the other hand, it totally makes sense to have multiple instances of event-processor-like integrations.

I think I will revert the unique-ness check, and rather better document this case, and also the default integrations.

jan-auer:
> unless you explicitly use default_integrations: false, and manually add all the defaults

But isn't that exactly what the unified spec says? My understanding was that you need to disable default integrations if you want to customize one of them.

Swatinem:
> But isn't that exactly what the unified spec says?

Hm, it doesn’t really say anything about integrations :-D

Anyhow, I added some more docs here.

PR #229, opened 2020-06-18
meta: Update publishing configuration/scripts

Swatinem:


Swatinem:
> You don't like perl? ;)

:-D Don’t really care either way. I copied this from here:

https://github.com/getsentry/symbolic/blob/67ceb271ad7d7fc659ad976293fca1e3cb72507f/scripts/bump-version#L12-L13

Its actually incredible that these scripts are completely unique across all sentry projects that I have looked at. And ideally craft would just do automatically depending on the type of project. One can still dream, right? ;-)

jan-auer:
Please revert this. The bump-version script does not work on Linux. Either run with PERL, or do this instead:
https://github.com/getsentry/relay/blob/63d16b971ba0466aa7998bf1da2f269f4d4f60b0/scripts/bump-version.sh#L14-L16

PR #232, opened 2020-06-27
Removed not existing feature from log crate that made methods not accessible

tglman:
The method `with_env_logger_dest` is not accessible by public API because is behind a not existing feature.

### DIFF
--- a/sentry-log/src/integration.rs
+++ b/sentry-log/src/integration.rs
@@ -62,7 +62,6 @@ impl Default for LogIntegration {
 
 impl LogIntegration {
     /// Initializes an env logger as destination target.
-    #[cfg(feature = "env_logger")]
     pub fn with_env_logger_dest(mut self, logger: Option<env_logger::Logger>) -> Self {
         let logger = logger
             .unwrap_or_else(|| env_logger::Builder::from_env(env_logger::Env::default()).build());


Swatinem:
https://github.com/getsentry/sentry-rust/blob/301fcb7188b8ebe4a4c8d13e31ee0892cbc04b5f/sentry-log/Cargo.toml#L21

The dependency is optional, which means there is an implicit feature with that name. Unless I have something completely wrong, this PR is invalid?

tglman:
Hi,
I come to this conclusion because i was trying to add this package as dependency trough:
`sentry = { version ="0.19", features=["log"]}` and i could not see this method in my code, as well I noticed that there was no feature acutally named "env_logger" (didn't know you could use a dependecy name as feature), so yes i just need to figure out how to activate transitively activate this feature if possible and than this can be just closed.

Regards

tglman:
Hi, 
I could work around this doing:
```
sentry = { version ="0.19", features=["log"]}
sentry-log = { version ="0.19", features=["env_logger"]}
```
not  sure sentry is supposed to be used like that, but for now it works, closing this.

Regards

Swatinem:
Ah yes, maybe putting an explicit `env_logger` feature on `sentry` itself, which then pulls in the correct dependency would be a good idea.

PR #234, opened 2020-07-03
meta: Use cargo-readme to generate each crates README

Swatinem:
TODO as a followup: improve the crate docs of some sub-crates.

### DIFF
--- a/scripts/update-readme.sh
+++ b/scripts/update-readme.sh
@@ -4,4 +4,6 @@ set -euo pipefail
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $SCRIPT_DIR/..
 
-find sentry* -name Cargo.toml -execdir cargo readme --output README.md \;
+NEW_VERSION="${1}"
+
+find sentry* -name Cargo.toml -execdir ../scripts/generate-readme.sh $NEW_VERSION \;
--- a/scripts/update-readme.sh
+++ b/scripts/update-readme.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $SCRIPT_DIR/..
+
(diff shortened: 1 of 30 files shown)


scripts/update-readme.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $SCRIPT_DIR/..
+
+find sentry* -name Cargo.toml -execdir cargo readme --output README.md \;
```
flub, line 7: no newline at end of a shell file can all to often be a trap

Swatinem:
> Could we add a CI check to ensure the READMEs are never out of date?

Hm, I rather not, I think it might make reviews harder. Also that would mean we would need to create a commit hook for it as well. I rather have a TODO to hook it up to the versioning script, so the README will be updated at the time of a release.

flub:
> > Could we add a CI check to ensure the READMEs are never out of date?
> 
> Hm, I rather not, I think it might make reviews harder.

How so?  There's some generated files that can be ignored when reviewing, that's what you get for committing generated files but nothing that makes review much harder does it?

> Also that would mean we would need to create a commit hook for it as well.

Do we have a standard commit hook format?  I know of pre-commit but I didn't know if we were using it.

> I rather have a TODO to hook it up to the versioning script, so the README will be updated at the time of a release.

That can be confusing while there's no release can it not?

Swatinem:
> Do we have a standard commit hook format? I know of pre-commit but I didn't know if we were using it.

I don’t think so. Also, I would rather not want to pull in either python or node dependencies.

Also, this is only relevant for the per-crate README, which shows up on crates.io, but I doubt anyone wants to directly navigate there on Github.

jan-auer:
What if we run `cargo-readme` as part of the `bump-version.sh` script? That way it's always up-to-date when we release.

Swatinem:
```#![doc(html_favicon_url = "https://example.com/favicon.ico")]``` is still missing, will do it as a separate PR.

PR #236, opened 2020-07-20
feat: Convert Transport to deal with Envelopes

Swatinem:


sentry-core/src/envelope.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    fn to_buf(envelope: Envelope) -> std::io::Result<Vec<u8>> {
+        let mut vec = vec![];
+        envelope.to_writer(&mut vec)?;
+        Ok(vec)
```
flub, line 119: Since writing to `Vec` is infallible you could remove the `Result` here and unwrap right away.  (Additionally you're just unwrapping this function at al callers anyway afaik)

sentry-core/src/envelope.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    #[test]
+    fn test_empty() {
+        assert_eq!(to_buf(Envelope::new()).unwrap(), b"{}\n");
+    }
+
+    #[test]
+    fn test_event() {
+        let event_id = Uuid::parse_str("22d00b3f-d1b1-4b5d-8d20-49d138cd8a9c").unwrap();
```
flub, line 129: Does it make sense to add a test with the event_id set to None as well?

sentry-core/src/envelope.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// A Sentry Envelope.
+///
+/// An Envelope is the data format that Sentry uses for Ingestion. It can contain
+/// multiple Items, some of which are related, such as Events, and Event Attachments.
+/// Other Items, such as Sessions are independant.
+///
+/// See the [documentation on Envelopes](https://develop.sentry.dev/sdk/envelopes/)
+/// for more details.
+#[derive(Default, Debug, PartialEq)]
```
flub, line 30: Event is clonable, should Envelope also be clonable?

sentry-core/src/envelope.rs
```diff
@@ -0,0 +1,148 @@
+use std::io::Write;
+
+use crate::protocol::Event;
```
flub, line 3: Should envelope not also live there?

sentry-log/src/integration.rs
```diff
@@ -42,7 +42,10 @@ impl Integration for LogIntegration {
             log::set_max_level(filter);
         }
 
-        INIT.call_once(|| log::set_boxed_logger(Box::new(Logger::default())).unwrap());
+        INIT.call_once(|| {
+            // silently ignore errors here, as running tests will hit this condition for some reason.
+            log::set_boxed_logger(Box::new(Logger::default())).ok();
+        });
```
flub, line 48: So tests end up running this more than once?  I'm not convinced this is the right work around in this case

sentry/src/transport.rs
```diff
@@ -222,9 +222,12 @@ implement_http_transport! {
                         }
                     }
 
+                    let mut body = vec![];
```
flub, line 225: Out of curiosity, I tend to write this as `Vec::new()` but you use the macro, are there any advantages to using the macro?  (my choice is purely because I like code with fewer macros)

sentry/src/transport.rs
```diff
@@ -222,9 +222,12 @@ implement_http_transport! {
...
+                    envelope.to_writer(&mut body).unwrap();
```
flub, line 226: The `envelope.to_writer()` call also ends up calling some serde stuff that returns results instead of only writing into the vec (which is infallible).  Are you sure it's safe to unwrap those errors?

flub: (Yes, I am aware that differs from my comment on the test code of the envelope, but this isn't test code i gues)

sentry/src/transport.rs
```diff
@@ -325,7 +328,10 @@ implement_http_transport! {
                     _ => {}
                 }
 
-                let mut body = Cursor::new(serde_json::to_vec(&event).unwrap());
+                let mut body = vec![];
+                envelope.to_writer(&mut body).unwrap();
```
flub, line 332: same about errors i think

sentry/src/transport.rs
```diff
@@ -440,25 +446,16 @@ implement_http_transport! {
                         }
                     }
 
-                    let req_result = http_client
+                    let mut body = vec![];
+                    envelope.to_writer(&mut body).unwrap();
```
flub, line 450: same again i think

sentry-core/src/envelope.rs
Swatinem: The `event_id` is not nullable. I can add more tests once I add session support.

sentry-core/src/envelope.rs
Swatinem: you mean the `protocol` module? Not really sure, but we can move it there later on as well.

sentry/src/transport.rs
Swatinem: Actually there is so much unwrapping going on in the transports, I don’t think its worth caring about this I think.

sentry-log/src/integration.rs
Swatinem: So this was super weird. Some of my local testruns were crashing hard. The global logger is not thread safe, so I suspect thats what was causing things to fail, though not sure.

sentry-core/src/envelope.rs
flub: yeah, meant the protocol module.  I guess need to decide on this before releasing it though

PR #240, opened 2020-08-14
feat: Add basic session support

Swatinem:
This adds release health sessions to the SDK, with some things that require special attention:

* sessions live on the scope.
* sessions are forward-inherited, which means that creating a new Hub from another one, and pushing scopes will use the exact same session.
* however, sessions are not backward-inherited, which means that starting a new session will not propagate backwards to any parent scopes/hubs.

So far, intermediate session updates are sent along with normal events, however session end updates are sent on their own and not yet batched together. This means that request-mode sessions are not yet as efficient as we would like them to be.

### DIFF
--- a/sentry-core/src/api.rs
+++ b/sentry-core/src/api.rs
@@ -1,6 +1,6 @@
 use crate::protocol::{Event, Level};
 use crate::types::Uuid;
-use crate::{Hub, Integration, IntoBreadcrumbs, Scope, SessionGuard};
+use crate::{Hub, Integration, IntoBreadcrumbs, Scope};
 
 /// Captures an event on the currently active client if any.
 ///
@@ -264,18 +264,21 @@ pub fn last_event_id() -> Option<Uuid> {
 
 /// Start a new session for Release Health.
 ///
-/// Returns a guard that ends the session on drop and thus determines the
-/// lifetime of the session.
-///
 /// # Examples
 ///
 /// ```
-/// let session = sentry::start_session();
+/// sentry::start_session();
 ///
-/// // capturing any event / error here will update the sessions `errors` count.
+/// // capturing any event / error here will update the sessions `errors` count,
+/// // up until we call `sentry::end_session`.
 ///
-/// drop(session);
+/// sentry::end_session();
 /// ```
-pub fn start_session() -> SessionGuard {
+pub fn start_session() {
     Hub::with_active(|hub| hub.start_session())
+}
+
+/// End the current Release Health Session.
+pub fn end_session() {
+    Hub::with_active(|hub| hub.end_session())
 }
--- a/sentry-core/src/api.rs
+++ b/sentry-core/src/api.rs
@@ -1,6 +1,6 @@
 use crate::protocol::{Event, Level};
 use crate::types::Uuid;
-use crate::{Hub, Integration, IntoBreadcrumbs, Scope};
+use crate::{Hub, Integration, IntoBreadcrumbs, Scope, SessionGuard};
 
 /// Captures an event on the currently active client if any.
 ///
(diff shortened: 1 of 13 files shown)


sentry-core/src/session.rs
```diff
@@ -0,0 +1,316 @@
...
+pub enum SessionUpdate {
+    NeedsFlushing(Session),
+    Unchanged,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Session {
+    session_id: Uuid,
+    status: SessionStatus,
+    errors: usize,
```
flub, line 32: why tie this to pointer size instead of u64?  more efficient?

sentry-core/src/session.rs
```diff
@@ -0,0 +1,316 @@
...
+    user: Option<Arc<User>>,
+    release: Cow<'static, str>,
+    environment: Option<Cow<'static, str>>,
+    started: Instant,
+    started_utc: DateTime<Utc>,
+    duration: Option<f64>,
```
flub, line 38: I had to look up the api docs to find out this is "seconds as float".  Either documenting this or just keeping it here as a `Duration` until it's serialised might be better?  I think I'd vote for `Duration`.

sentry-core/src/session.rs
```diff
@@ -0,0 +1,316 @@
...
+        let mut is_crash = false;
+        for exc in &event.exception.values {
+            has_error = true;
+            if let Some(mechanism) = &exc.mechanism {
+                if let Some(false) = mechanism.handled {
+                    is_crash = true;
+                    break;
+                }
+            }
+        }
```
flub, line 77: Are these panics in the rust sdk?  or could they be something else as well?

sentry-core/src/session.rs
```diff
@@ -0,0 +1,316 @@
...
+
+        if is_crash {
+            self.status = SessionStatus::Crashed;
```
flub, line 80: If this is different from a previous status should it not be considered dirty?

sentry-core/src/client.rs
```diff
@@ -239,14 +240,33 @@ impl Client {
...
+                    let session = scope.and_then(|scope| {
+                        if let SessionUpdate::NeedsFlushing(session) =
+                            scope.update_session_from_event(&event)
+                        {
+                            Some(session)
+                        } else {
+                            None
+                        }
+                    });
+                    let mut envelope: Envelope = event.into();
```
flub, line 252: If you move this to before the `let session` you don't need the if statement on the next line and can and it right away when you get `NeedsFlushing` (and don't need the else branch there either)

sentry-core/src/scope/real.rs
```diff
@@ -114,6 +118,36 @@ impl Stack {
...
+    }
+}
+
+impl Drop for SessionGuard {
+    fn drop(&mut self) {
+        if let Some((client, scope)) = self.0.take() {
+            if let Some(session) = scope.session.lock().unwrap().take() {
+                if let SessionUpdate::NeedsFlushing(session) = session.close() {
+                    let mut envelope = Envelope::new();
+                    envelope.add(session.into());
```
flub, line 143: Worth changing the `envelope::add()` to accept `impl Into<...>`?

sentry-core/src/session.rs
Swatinem: Yes, I recently changed the panic handler to emit the correct `mechanism` for this.

sentry-core/src/session.rs
Swatinem: doesn’t really matter I guess, upstream does not dictate any type, and for those cases `usize` is the most appropriate I would say.

sentry-core/src/session.rs
Swatinem: it might not be totally obvious, but `is_crash` implies `has_error`, which makes it `dirty`.

sentry-core/src/client.rs
Swatinem: `Scope::update_session_from_event` needs a `&Event`, but `event.into()` consumes it. The second `if let` is unfortunate, but we do need it because of lifetimes.

sentry-core/src/api.rs
```diff
@@ -264,18 +264,21 @@ pub fn last_event_id() -> Option<Uuid> {
...
 ///
-/// // capturing any event / error here will update the sessions `errors` count.
+/// // capturing any event / error here will update the sessions `errors` count,
+/// // up until we call `sentry::end_session`.
 ///
-/// drop(session);
+/// sentry::end_session();
 /// ```
-pub fn start_session() -> SessionGuard {
+pub fn start_session() {
```
jan-auer, line 277: It would be great to denote that this is an experimental API and not ready for a high volume of sessions yet.

sentry-core/src/session.rs
```diff
@@ -40,10 +41,23 @@ pub struct Session {
...
+    }
+}
+
 impl Session {
     pub fn from_stack(stack: &StackLayer) -> Option<Self> {
-        let options = stack.client.as_ref()?.options();
+        let client = stack.client.as_ref()?;
+        let options = client.options();
         Some(Self {
+            client: client.clone(),
```
jan-auer:
```suggestion
        let client = stack.client.clone()?;
        let options = client.options();
        Some(Self {
            client,
```

jan-auer (approved):

sentry-core/src/session.rs
Swatinem: nope ;-) you can’t access anything from `options` *after* you have moved `client` into the new struct :-D

PR #243, opened 2020-08-21
feat: Use Envelope Endpoint and add envelope test support

Swatinem:


sentry-types/src/dsn.rs
```diff
@@ -82,17 +82,33 @@ impl Dsn {
...
+            self.project_id(),
+            endpoint
+        )
+        .unwrap();
         Url::parse(&buf).unwrap()
     }
 
+    /// Returns the submission API URL.
+    pub fn store_api_url(&self) -> Url {
+        self.api_url("store")
```
flub, line 104: This seems a bit weird.  I wonder if it might be better off storing the base `Url` struct in the DSN and then use its `join()` method for these.

flub (approved):

sentry-types/src/dsn.rs
flub: lol, the `Url::join` implementation invokes the whole parse machinery anyway so it really doesn't matter.  this is probably cleaner from the dsn's point of view then

PR #244, opened 2020-08-21
fix: Add an unhandled mechanism to panics

Swatinem:
The "handled" prop of the exception mechanism is used in some places, for example the upcoming release health feature that will flag sessions as "crashed" if they are not handled.


jan-auer (approved):
That is great!

Issue #239, opened 2020-08-07
Client doesn't log errors


WhyNotHugo:
My code is fairly simply, and copying the panic example from the docs doesn't log anything (nothing shows up on Sentry, and it doesn't seem to pause to send the events).

```rust
//...

#[tokio::main]
async fn main() -> Result<()> {
    let version = env!("VERSION").to_string();
    info!("Version {}", version);

    let sentry = sentry::init((
        "https://XXXXXX@o58074.ingest.sentry.io/YYYYY",
        sentry::ClientOptions {
            release: Some(format!("sync@{}", version).into()),
            ..Default::default()
        },
    ));

    configure_logger();

    if sentry.is_enabled() {
        debug!("Sentry is enabled");
    }

    //...

    panic!("Everything is on fire!");
//...
```

```
[2020-08-07 22:42:57] [DEBUG][sync_tool] Sentry is enabled
thread 'main' panicked at 'Everything is on fire!', src/main.rs:165:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

No event on sentry.

Any hints? This is pretty close to a copy-paste from the docs on all sides.

GeorgeHahn:
I observe the same issue on versions 19.0 and 0d9d1d0. I've traced through the panic handler and can see the sentry handler run, but the event doesn't show up in the web interface.

ErichDonGubler:
This hasn't been resolved with 0.20.0, has it?

Swatinem:
I had hoped that it has. Can you provide any more Info on this?

ErichDonGubler:
I unfortunately am not familiar with usage of `sentry` as a crate, though I'm now responsible for a codebase that uses it with @NZXTCorp. I would defer information gathering to @GeorgeHahn and @WhyNotHugo, since I haven't reproduced the issue myself yet -- I'm just assessing risk with issues for my team right now. :)

Mastermindaxe:
Can confirm that this is still happening even with the minimal example that is stated in the docs. Using version `0.20.1`. I'm trying to see if I can fix it by going back a few versions

jan-auer:
@Mastermindaxe Up to SDK version `0.18`, you have to register the panic handler using [`sentry::integrations::panic::register_panic_handler`](https://docs.rs/sentry/0.18.1/sentry/integrations/panic/fn.register_panic_handler.html). 

[...]

PR #245, opened 2020-08-24
fix: Deal with concurrent hub access better

### LINKED ISSUE #237: Bug: Using sentry_debug on a non-main thread before init may break Hub::main
Referenced code: [`hub.rs`]

**TL;DR**: Using `sentry_debug` may break `Hub::main` if it happens too early on a non-main thread. Using the `debug-log` feature may avoid this behavior.

[`Hub::main`] reads from a lazy-static [`PROCESS_HUB`] value. When `PROCESS_HUB` is first used, it is initialized with the thread ID of the current thread. It is assumed that the first usage of `PROCESS_HUB` is in the main function (via `sentry::init`), but this is not always the case.

There are also two thread-local values, [`THREAD_HUB`] and [`USE_PROCESS_HUB`]. `USE_PROCESS_HUB` holds a boolean which decides if the current thread should use the main thread's hub, aka `PROCESS_HUB`. These thread-locals are used by [`Hub::with`] to figure out if the hub should be taken from `PROCESS_HUB` or `THREAD_HUB`.

The issue occurs if the first usage of `PROCESS_HUB` is outside of the main thread. This initializes its thread ID to the current thread ID, which is not the main thread's ID. Future calls to `Hub::with` on the main thread will use `THREAD_HUB` instead of `PROCESS_HUB`. Since the Sentry client is initialized on the main thread, and the main thread is not using `PROCESS_HUB`, the initialized client is inaccessible to other threads.

The concrete way that this occurs in my code is when I provide a non-default transport in the client options (Edit: happens with the default reqwest transport as well):
```rust
let curl_transport_factory = |options: &sentry::ClientOptions| {
    Arc::new(sentry::transports::CurlHttpTransport::new(options)) as Arc<dyn sentry::Transport>
};
let options = sentry::ClientOptions {
    transport: Some(Arc::new(curl_transport_factory)),
    release: sentry::release_name!(),
    ..sentry::ClientOptions::default()
};
let client = sentry::init(options);
```

[The transport factory method is called before the client is bound to the hub][create_transport]. [`CurlHttpTransport::new`] spawns a thread which immediately calls [`sentry_debug`]. `sentry_debug` uses `log::debug` when the `debug-logs` feature is enabled (a non-default feature), but otherwise uses `Hub::with`. As long as the new thread reaches the `Hub::with` call before the main thread accesses `PROCESS_HUB`, this bug is triggered.

[`hub.rs`]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry-core/src/hub.rs
[`Hub::main`]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry-core/src/hub.rs#L132
[`Hub::with`]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry-core/src/hub.rs#L141
[`PROCESS_HUB`]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry-core/src/hub.rs#L19
[`THREAD_HUB`]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry-core/src/hub.rs#L27
[`USE_PROCESS_HUB`]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry-core/src/hub.rs#L29
[create_transport]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry-core/src/client.rs#L102
[`CurlHttpTransport::new`]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry/src/transport.rs#L289
[`sentry_debug`]: https://github.com/getsentry/sentry-rust/blob/36218ea5e815d871c073ee15d7f71ee01dab69c7/sentry-core/src/macros.rs#L56

Swatinem:
This should fix #237, and we should backport this to 0.19 as well.

### DIFF
--- a/sentry-core/src/hub.rs
+++ b/sentry-core/src/hub.rs
@@ -52,10 +52,12 @@ impl HubImpl {
             Err(err) => err.into_inner(),
             Ok(guard) => guard,
         };
-        if let Some(client) = guard.top().client.as_ref() {
-            return client.is_enabled();
-        }
-        false
+
+        guard
+            .top()
+            .client
+            .as_ref()
+            .map_or(false, |c| c.is_enabled())
     }
 }
 
--- a/sentry-core/src/hub.rs
+++ b/sentry-core/src/hub.rs
@@ -48,12 +48,14 @@ impl HubImpl {
     }
 
     fn is_active_and_usage_safe(&self) -> bool {
-        let guard = match self.stack.try_read() {
-            Err(TryLockError::Poisoned(err)) => err.into_inner(),
-            Err(TryLockError::WouldBlock) => return false,
+        let guard = match self.stack.read() {
+            Err(err) => err.into_inner(),
             Ok(guard) => guard,
         };
-        guard.top().client.is_some()
+        if let Some(client) = guard.top().client.as_ref() {
+            return client.is_enabled();
+        }
+        false
     }
 }
 
--- a/sentry-log/src/integration.rs
+++ b/sentry-log/src/integration.rs
@@ -43,8 +43,7 @@ impl Integration for LogIntegration {
         }
 
         INIT.call_once(|| {
-            // silently ignore errors here, as running tests will hit this condition for some reason.
-            log::set_boxed_logger(Box::new(Logger::default())).ok();
+            log::set_boxed_logger(Box::new(Logger::default())).unwrap();
         });
     }
 }
--- a/sentry-log/src/integration.rs
+++ b/sentry-log/src/integration.rs
@@ -43,7 +43,7 @@ impl Integration for LogIntegration {
         }
 
         INIT.call_once(|| {
-            log::set_boxed_logger(Box::new(Logger::default())).unwrap();
+            log::set_boxed_logger(Box::new(Logger::default())).ok();
         });
     }
 }
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -11,7 +11,7 @@ use rand::random;
 use crate::constants::SDK_INFO;
 use crate::protocol::{ClientSdkInfo, Event};
 use crate::types::{Dsn, Uuid};
-use crate::{ClientOptions, Integration, Scope, Transport};
+use crate::{ClientOptions, Hub, Integration, Scope, Transport};
 
 impl<T: Into<ClientOptions>> From<T> for Client {
     fn from(o: T) -> Client {
@@ -93,6 +93,10 @@ impl Client {
     /// If the DSN on the options is set to `None` the client will be entirely
     /// disabled.
     pub fn with_options(mut options: ClientOptions) -> Client {
+        // Create the main hub eagerly to avoid problems with the background thread
+        // See https://github.com/getsentry/sentry-rust/issues/237
+        Hub::with(|_| {});
+
         let create_transport = || {
             options.dsn.as_ref()?;
             let factory = options.transport.as_ref()?;


sentry-log/src/integration.rs
```diff
@@ -43,8 +43,7 @@ impl Integration for LogIntegration {
         }
 
         INIT.call_once(|| {
-            // silently ignore errors here, as running tests will hit this condition for some reason.
-            log::set_boxed_logger(Box::new(Logger::default())).ok();
+            log::set_boxed_logger(Box::new(Logger::default())).unwrap();
```
jan-auer, line 46: Just in case someone already set the logger, should we still `.ok()` here, or even log the error? Even if this is a static programming error, it's a bit harsh to panic here.

sentry-core/src/hub.rs
```diff
@@ -48,12 +48,14 @@ impl HubImpl {
...
     fn is_active_and_usage_safe(&self) -> bool {
-        let guard = match self.stack.try_read() {
-            Err(TryLockError::Poisoned(err)) => err.into_inner(),
-            Err(TryLockError::WouldBlock) => return false,
+        let guard = match self.stack.read() {
+            Err(err) => err.into_inner(),
             Ok(guard) => guard,
         };
-        guard.top().client.is_some()
+        if let Some(client) = guard.top().client.as_ref() {
```
jan-auer, line 55: You can shorten this to:

```suggestion
        if let Some(ref client) = guard.top().client {
```

A match:

```suggestion
        match guard.top().client {
            Some(ref client) => client.is_enabled(),
            None => false,
        }
```

Or even a one-liner:

```suggestion
        guard.top().client.as_ref().map_or(false, |c| c.is_enabled())
```

jan-auer (approved):

jan-auer:
@Swatinem does this also fix https://github.com/getsentry/sentry-rust/issues/239?

sentry-log/src/integration.rs
Swatinem: We had the `unwrap` here originally, which I changed just recently: https://github.com/getsentry/sentry-rust/commit/97b90064a464dc0e1a01b081091b5d2e1d4f6941#diff-8c955be4e92f6bf80501c6237c378f5e

After looking into this with @mitsuhiko we think it was rather a problem of libbacktrace segfaulting on us. Since we should rather fail hard in case of misconfiguration rather than silently not do anything, I think its okay to `unwrap` here, in the case the user has really configured the logger twice.

sentry-core/src/hub.rs
Swatinem: ahhh, the expressive power of rust ;-)

Swatinem:
> does this also fix #239?

I will do some more tests also on windows where I was also seeing this problem.

sentry-log/src/integration.rs
jan-auer: Well, the original `log::set_boxed_logger` also doesn't panic and rather errors. The problem is that it's much harder to recover from a panic than from an  error, and our SDK should never crash user apps, even when misconfigured. I'd rather log an error and disable the integration than make the user application crash. Imagine this is only triggered with production configuration for some customers and then hits them where it hurts them most.

PR #248, opened 2020-09-02
feat: Add per-request sessions to actix integration

Swatinem:
Built on top of #240 , this has to be opt-in for now, as the actix middleware currently has no way to disable the *user-mode* session which is now the default on the main Hub.

### DIFF
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -9,6 +9,41 @@
 //! # Example
 //!
 //! ```no_run
+//! use actix_web::{server, App, Error, HttpRequest};
+//! use sentry_actix::SentryMiddleware;
+//!
+//! fn failing(_req: &HttpRequest) -> Result<String, Error> {
+//!     Err(std::io::Error::new(std::io::ErrorKind::Other, "An error happens here").into())
+//! }
+//!
+//! fn main() {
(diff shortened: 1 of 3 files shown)


sentry-actix/src/lib.rs
```diff
@@ -116,13 +157,21 @@ impl SentryMiddlewareBuilder {
...
+
+    /// Enables or disables per-request Session tracking.
+    pub fn track_session(mut self, val: bool) -> Self {
+        self.middleware.track_session = val;
+        self
+    }
 }
 
 /// Reports certain failures to sentry.
+//#[derive(Clone)]
```
flub, line 169: Not related to this PR really, but we don't make this `Debug`?

sentry-actix/src/lib.rs
```diff
@@ -211,6 +261,9 @@ impl<S: 'static> Middleware<S> for SentryMiddleware {
         let cached_data = Arc::new(Mutex::new(None));
 
         let root_scope = hub.push_scope();
+        if self.track_session {
+            hub.start_session();
+        }
```
flub, line 266: How does it stop?

Swatinem: It stops when the scope goes out of scope :-D

Swatinem:
old actix is gone

PR #252, opened 2020-09-07
ci: Remove Appveyor and Travis CI

Swatinem:


.craft.yml
```diff
@@ -7,7 +7,6 @@ changelogPolicy: auto
 targets:
   - name: crates
     noDevDeps: true
-  - name: gh-pages
```
jan-auer, line 10: Is that because the artifact is missing or are we removing GH pages entirely?

jan-auer (approved):

.craft.yml
Swatinem: That stuff was previously built in a travis job. Its basically duplicated with docs.rs: https://getsentry.github.io/sentry-rust/sentry/

So I think we should remove it in favor of that.

jan-auer: Right. I believe we introduced this slightly before docs.rs was a thing but kept it alive. I'm fine with turning off completely 👍

PR #254, opened 2020-09-08
chore: Update to use `im 15.0.0`

### LINKED ISSUE #253: cargo audit flagging `sized-chunks 0.5.3`
```
error: Vulnerable crates found!

ID:       RUSTSEC-2020-0041
Crate:    sized-chunks
Version:  0.5.3
Date:     2020-09-06
URL:      https://rustsec.org/advisories/RUSTSEC-2020-0041
Title:    Multiple soundness issues in Chunk and InlineArray
Solution:  No safe upgrade is available!
Dependency tree:
sized-chunks 0.5.3
└── im 14.3.0
    └── sentry-core 0.19.1
        ├── sentry-panic 0.19.1
```

Not sure if updating `im` to 15.0.0 will fix it (or isn't a huge hassle).

jrconlin:
(that fix was simple and not at all obvious)

Closes #253

### DIFF
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -234,10 +234,10 @@ impl Scope {
             }
         }
 
-        event.breadcrumbs.extend(self.breadcrumbs.iter().cloned());
-        event.extra.extend(self.extra.iter().cloned());
-        event.tags.extend(self.tags.iter().cloned());
-        event.contexts.extend(self.contexts.iter().cloned());
+        event.breadcrumbs.extend(self.breadcrumbs.clone().into_iter());
+        event.extra.extend(self.extra.clone().into_iter());
+        event.tags.extend(self.tags.clone().into_iter());
+        event.contexts.extend(self.contexts.clone().into_iter());
 
         if event.transaction.is_none() {
             if let Some(ref txn) = self.transaction {
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -234,7 +234,9 @@ impl Scope {
             }
         }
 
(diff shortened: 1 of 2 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -234,10 +234,10 @@ impl Scope {
             }
         }
 
-        event.breadcrumbs.extend(self.breadcrumbs.iter().cloned());
-        event.extra.extend(self.extra.iter().cloned());
-        event.tags.extend(self.tags.iter().cloned());
-        event.contexts.extend(self.contexts.iter().cloned());
+        event.breadcrumbs.extend(self.breadcrumbs.clone().into_iter());
```
Swatinem, line 237: Just run rustfmt on this:
```
-        event.breadcrumbs.extend(self.breadcrumbs.clone().into_iter());
+        event
+            .breadcrumbs
+            .extend(self.breadcrumbs.clone().into_iter());
```

But can you explain why you need this? Is this an API change in `im`?

jrconlin: Sure thing. (Sorry, I always forget to `fmt`)

Yeah, this looks to be a change required by `im`. 
`extend()` changed from accepting a reference for a tuple to a tuple of references.

```
error[E0271]: type mismatch resolving `<im::hashmap::Iter<'_, std::string::String, sentry_types::protocol::v7::Context> as std::iter::Iterator>::Item == &_`
   --> sentry-core/src/scope/real.rs:243:52
    |
243 |         event.contexts.extend(self.contexts.iter().cloned());
    |                                                    ^^^^^^ expected tuple, found reference
    |
    = note:  expected tuple `(&std::string::String, &sentry_types::protocol::v7::Context)`
            found reference `&_`
```

PR #255, opened 2020-09-11
Fix typo in README.md

bbqsrc:


Swatinem:
These are actually auto-generated from rustdoc, which I recently fixed: https://github.com/getsentry/sentry-rust/commit/4901164653eade6affe49016240509e3408a25af
We re-generate the README on release, so this will magically solve itself on the next release ;-)

PR #259, opened 2020-09-14
fix(transport): Consume responses from Sentry

jan-auer:


Swatinem:
#260 does the same as well

PR #260, opened 2020-09-14
fix: Improve transport synchronization and consume response

Swatinem:
The synchronization fixes have been proposed in #172 originally

### DIFF
--- a/sentry/src/transport.rs
+++ b/sentry/src/transport.rs
@@ -78,7 +78,7 @@ macro_rules! implement_http_transport {
             shutdown_signal: Arc<Condvar>,
             shutdown_immediately: Arc<AtomicBool>,
             queue_size: Arc<Mutex<usize>>,
-            _handle: Option<JoinHandle<()>>,
+            handle: Option<JoinHandle<()>>,
         }
 
         impl $typename {
@@ -104,7 +104,7 @@ macro_rules! implement_http_transport {
                 #[allow(clippy::mutex_atomic)]
                 let queue_size = Arc::new(Mutex::new(0));
                 let http_client = http_client(options, $hc_client);
-                let _handle = Some(spawn(
+                let handle = Some(spawn(
                     options,
                     receiver,
                     shutdown_signal.clone(),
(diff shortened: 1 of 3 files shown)


sentry/src/transport.rs
```diff
@@ -242,9 +252,10 @@ implement_http_transport! {
                                     disabled = Some(retry_after);
                                 }
                             }
+                            sentry_debug!("Got response: {:?}", resp.text());
```
jan-auer, line 255: Two points here:
- The response must be consumed unconditionally, but at the moment it is conditional based on whether debug logs are emitted.
- If possible, can we avoid decoding to increase performance? You can still decode in the log, but not outside.

jan-auer (changes requested):
@Swatinem I assume that this makes https://github.com/getsentry/sentry-rust/pull/259 redundant?

Swatinem:
> @Swatinem I assume that this makes #259 redundant?

yes. And I updated your comments about the response log.

sentry/src/transport.rs
```diff
@@ -252,7 +251,10 @@ implement_http_transport! {
                                     disabled = Some(retry_after);
                                 }
                             }
-                            sentry_debug!("Got response: {:?}", resp.text());
+                            match resp.text() {
+                                Err(err) => { sentry_debug!("Failed to read sentry response: {}", err); },
+                                Ok(text) => { sentry_debug!("Get response: `{}`", text); },
```
jan-auer, line 256:
```suggestion
                                Err(err) => sentry_debug!("Failed to read sentry response: {}", err),
                                Ok(text) => sentry_debug!("Get response: `{}`", text),
```

jan-auer (approved):

sentry/src/transport.rs
jan-auer: I'm wondering how this does not fail rustfmt.

Swatinem: its inside the macro, so none of that stuff is rustfmt-ed, also, I needed to wrap this in a block because of the way the `sentry_debug` macro is built :-(

jan-auer: I see, thanks. At least the comment after the `}` should go usually.

PR #261, opened 2020-09-14
fix: Make the Logger completely self-contained

Swatinem:
Instead of using `with_integration`, which can potentially lock on every call, we rather unsafely move the `dest_log` into our own Logger.

### DIFF
--- a/sentry-log/src/integration.rs
+++ b/sentry-log/src/integration.rs
@@ -1,8 +1,9 @@
 use std::sync::Once;
 
-use log::{Level, LevelFilter, Record};
+use log::LevelFilter;
 use sentry_core::{ClientOptions, Integration};
 
+use crate::filters::Filters;
 use crate::logger::Logger;
 
 /// Logger specific options.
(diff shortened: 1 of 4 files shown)


sentry-log/src/integration.rs
```diff
@@ -37,13 +38,27 @@ impl Integration for LogIntegration {
...
+            // The way the current log-integration code is structured, we have
+            // no way to move the `dest_logger` from the integration to our own
+            // log instance without breaking the API.
+            // The `Once` here makes sure that we only ever do this unsafe
+            // `take` once, and there are no other pieces of code that read from
+            // from this integration instance.
+            let dest_log = unsafe {
+                let const_ptr = &self.dest_log as *const Option<Box<dyn log::Log>>;
+                let mut_ptr = const_ptr as *mut Option<Box<dyn log::Log>>;
+                (&mut *mut_ptr).take()
```
jan-auer, line 58: Okay, I know that I may have suggested this in a meeting, but we need to find another way to do this. If we end up breaking all integration interfaces to make them builders, I guess that is fine and a good lesson on why we should not expose fields publicly :) 

To quote https://doc.rust-lang.org/nightly/nomicon/transmutes.html (UB = undefined behavior):

> - Transmuting an & to &mut is _always_ UB.
> - No you can't do it.
> - No you're not special.

Swatinem:
closing this in favor of #268

PR #266, opened 2020-09-21
ref: Make the fields of ClientOptions private

Swatinem:


szokeasaurusrex:
Closing, as this appears to be out of date. I've made an issue to do this (#995) as it is still a good idea

PR #267, opened 2020-09-23
ref: Deprecate public fields on Integrations

Swatinem:
Not really sure if I should deprecate these first, or just break them right away. I don’t think people use these integrations directly a lot.

### DIFF
--- a/sentry-anyhow/src/lib.rs
+++ b/sentry-anyhow/src/lib.rs
@@ -12,7 +12,7 @@
 //! let _sentry =
 //!     sentry::init(sentry::ClientOptions::new().add_integration(AnyhowIntegration));
 //!
-//! if let Err(err) = match function_that_might_fail() {
+//! if let Err(err) = function_that_might_fail() {
 //!     capture_anyhow(&err);
 //! }
 //! ```
--- a/sentry-anyhow/src/lib.rs
+++ b/sentry-anyhow/src/lib.rs
@@ -4,8 +4,11 @@
 //!
 //! ```no_run
 //! # fn function_that_might_fail() -> anyhow::Result<()> { Ok(()) }
-//! use sentry_anyhow::capture_anyhow;
+//! use sentry_anyhow::{capture_anyhow, AnyhowIntegration};
 //! # fn test() -> anyhow::Result<()> {
+//! let _sentry =
+//!     sentry::init(sentry::ClientOptions::new().add_integration(AnyhowIntegration));
+//!
 //! let result = match function_that_might_fail() {
 //!     Ok(result) => result,
 //!     Err(err) => {
@@ -21,9 +24,6 @@
 #![warn(missing_docs)]
 #![deny(unsafe_code)]
 
-use std::error::Error;
-use std::fmt;
-
 use sentry_core::types::Uuid;
 use sentry_core::{ClientOptions, Hub, Integration};
 
@@ -32,7 +32,7 @@ use sentry_core::{ClientOptions, Hub, Integration};
 pub struct AnyhowIntegration;
 
 impl AnyhowIntegration {
-    /// Creates a new Failure Integration.
+    /// Creates a new anyhow Integration.
     pub fn new() -> Self {
         Self::default()
     }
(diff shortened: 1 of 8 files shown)


jan-auer:
> Not really sure if I should deprecate these first, or just break them right away.

I think since you're not removing functionality but just changing its signature, it's fine to remove right away.

sentry-debug-images/src/lib.rs
```diff
@@ -1,20 +1,23 @@
...
 //! use sentry_core::Level;
-//! let integration = sentry_debug_images::DebugImagesIntegration {
-//!     filter: Box::new(|event| event.level >= Level::Warning),
-//!     ..Default::default()
-//! };
+//! let integration = sentry_debug_images::DebugImagesIntegration::new()
+//!     .filter(|event| event.level >= Level::Warning);
 //! ```
+//!
+//! [`Event`]: sentry_core::Event
```
jan-auer, line 20: This link does not actually work. If I `cargo doc` and click it, it sends me to: `sentry_debug_images/sentry_core::Event`. Can you double check this and other links that we have?

jan-auer: I compiled with nightly. The other link for `DebugImagesIntegration` does work.

Swatinem:
```
warning: unresolved link to `sentry_core::Event`
  --> sentry-debug-images\src\lib.rs:20:16
   |
20 | //! [`Event`]: sentry_core::Event
   |                ^^^^^^^^^^^^^^^^^^ the crate `sentry_core` has no inner item named `Event`
   |
   = note: `#[warn(broken_intra_doc_links)]` on by default
```

Wow, this is awesome!

Also, I fixed some imports in sentry-core which error when using `no-default-features`, which apparently happens when you document a single one of these crates, lol.

sentry-anyhow/src/lib.rs
```diff
@@ -4,8 +4,11 @@
 //!
 //! ```no_run
 //! # fn function_that_might_fail() -> anyhow::Result<()> { Ok(()) }
-//! use sentry_anyhow::capture_anyhow;
+//! use sentry_anyhow::{capture_anyhow, AnyhowIntegration};
```
jan-auer, line 7: _nit_: Consider changing this example to include the function and have it return an error. Then you can simply capture the error and report it, no more return:


```rust
fn function_that_might_fail() -> anyhow::Result<()> { 
    Err("this is an error".into()) 
}

let _sentry = sentry::init(sentry::ClientOptions::new().add_integration(AnyhowIntegration));
if let Err(error) = match function_that_might_fail() {
    capture_anyhow(&err);
}
```

sentry-anyhow/src/lib.rs
```diff
@@ -57,36 +57,13 @@ pub fn capture_anyhow(e: &anyhow::Error) -> Uuid {
...
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        self.0.fmt(fmt)
-    }
-}
-
-impl Error for AnyhowError<'_> {
-    fn source(&self) -> Option<&(dyn Error + 'static)> {
-        self.0.source()
+        let e: &dyn std::error::Error = e.as_ref();
+        self.capture_error(e)
```
jan-auer, line 67: Just curious, why do you need to annotate types here?

sentry-contexts/src/integration.rs
```diff
@@ -36,6 +33,23 @@ impl ContextIntegration {
     pub fn new() -> Self {
         Self::default()
     }
+
+    /// Add `os` context, enabled by default.
+    pub fn add_os(mut self, add_os: bool) -> Self {
+        self.add_os = add_os;
+        self
+    }
+    /// Add `rust` context, enabled by default.
```
jan-auer, line 42: _nit_: Add an empty line. 

The `add_` prefixes seem a bit redundant, but since they are speaking they are fine.

sentry-debug-images/src/integration.rs
```diff
@@ -5,15 +5,25 @@ use sentry_core::{ClientOptions, Integration};
...
     pub fn new() -> Self {
         Self::default()
     }
+
+    /// Sets a custom filter function.
+    ///
+    /// The filter specified which [`Event`]s should get debug images.
+    pub fn filter<F>(mut self, filter: F) -> Self
+    where
+        F: Fn(&Event<'static>) -> bool + Send + Sync + 'static,
```
jan-auer, line 22: Since this event is immutable, you can make it `&Event<'_>`

jan-auer (approved):

sentry-anyhow/src/lib.rs
Swatinem: can’t remember exactly, but I needed it to make rust happy.

PR #268, opened 2020-09-23
ref: Rearchitect the log and slog Integrations

Swatinem:
In hindsight, putting all the filters/mappers on the Hub/Client-bound Integration was a mistake, and all that should be inside a lightweight logger.


sentry-log/src/lib.rs
```diff
@@ -34,6 +32,6 @@ mod converters;
 mod integration;
 mod logger;
 
-pub use converters::{breadcrumb_from_record, event_from_record};
+pub use converters::*;
 pub use integration::LogIntegration;
-pub use logger::Logger;
+pub use logger::{NoopLogger, SentryLogger};
```
jan-auer, line 37: Exports for some of the types are missing, namely `LevelFilter` and `RecordMapping`. There are no doc tests that check for this, either.

sentry-log/src/logger.rs
```diff
@@ -1,47 +1,158 @@
-use crate::converters::{breadcrumb_from_record, event_from_record};
-use crate::LogIntegration;
+use log::Record;
+use sentry_core::protocol::{Breadcrumb, Event};
+
+use crate::converters::{breadcrumb_from_record, event_from_record, exception_from_record};
+
+/// The Action that Sentry should perform for a [`log::Metadata`].
+#[derive(Debug)]
+pub enum LevelFilter {
```
jan-auer, line 8: Can we give this a different name so that this can be imported alongside `log::LevelFilter`?

sentry-slog/src/integration.rs
```diff
@@ -1,129 +1,14 @@
...
-        #[derive(Debug)]
-        struct SlogMapper;
-        let mapper = self.mapper.as_ref().map(|_| SlogMapper);
-        f.debug_struct("SlogIntegration")
-            .field("mapper", &mapper)
-            .finish()
-    }
-}
+#[derive(Debug, Default)]
+pub struct SlogIntegration {}
```
jan-auer, line 5: I'm starting to wonder if this integration makes sense, since it has no logic. Is there no need for an `slog::` in app filter, at least?

Since this is so basic, and probably also useful even without the integration, what if we remove the integration, just expose the logger, and add the in_app filters to `sentry_core` by default? Same goes for `sentry-log::LogIntegration`.

jan-auer (commented):
Two general points:

1. As opposed to before, the user now has to call `log::set_max_level_filter` manually. We need to clearly document that in the migration notes (Changelog), as otherwise suddenly logs will just be suppressed.

2. What's now not really possible is to create breadcrumbs with a higher level than the inner max level. Following example: We may want breadcrumbs on TRACE level, but only log INFO to stderr. Do you think that is something we may be able to achieve? My initial hunch says this is too much of an edge case and can be solved in the log stash, too.

sentry-slog/src/integration.rs
Swatinem: Hm, I’m wondering how useful these "just add myself to the integrations list" kind of integrations are, and if we should just remove those altogether.

jan-auer: So I've been thinking a little more about it. If the integration is expected to perform more work during initialization and setup, then I think they do make sense. 

But if they really only add something that could be added by default, then I'd just go ahead and put it in `sentry_core`. The point about the integration is to hook into the lifecycle, but if we don't need that, then we can as well just remove it. 

A good counter example is the panic integration. It registers something during setup. 

cc @HazAT do you have any opinion on this?

Swatinem: I think we can make the decision later on if we want to rather remove all these "noop" integrations.

sentry-log/src/lib.rs
```diff
@@ -41,6 +41,7 @@
 #![doc(html_favicon_url = "https://sentry-brand.storage.googleapis.com/favicon.ico")]
 #![doc(html_logo_url = "https://sentry-brand.storage.googleapis.com/sentry-glyph-black.png")]
 #![warn(missing_docs)]
+#![allow(clippy::match_like_matches_macro)]
```
jan-auer, line 44: Instead of allowing this, can we just fix this?

sentry-log/src/lib.rs
```diff
@@ -8,32 +8,45 @@
 //! # Examples
 //!
 //! ```
+//! let mut log_builder = pretty_env_logger::formatted_builder();
+//! log_builder.parse_filters("info");
+//! let logger = sentry_log::SentryLogger::with_dest(log_builder.build());
+//!
+//! log::set_boxed_logger(Box::new(logger))
+//!     .map(|()| log::set_max_level(log::LevelFilter::Info))
+//!     .unwrap();
```
jan-auer, line 17: Can we simplify this?


```suggestion
//! log::set_boxed_logger(Box::new(logger)).unwrap();
//! log::set_max_level(log::LevelFilter::Info);
```

sentry-slog/src/drain.rs
```diff
@@ -1,15 +1,87 @@
-use crate::SlogIntegration;
+use sentry_core::protocol::{Breadcrumb, Event};
 use slog::{Drain, OwnedKVList, Record};
 
-/// A Drain which passes all Records to sentry.
+use crate::{breadcrumb_from_record, event_from_record, exception_from_record};
+
+/// The Action that Sentry should perform for a [`slog::Level`].
```
jan-auer, line 6:
```suggestion
/// The action that Sentry should perform for a [`slog::Level`].
```

sentry-slog/src/drain.rs
```diff
@@ -1,15 +1,87 @@
...
+    /// Create a message [`Event`] from this [`Record`].
+    Event,
+    /// Create an exception [`Event`] from this [`Record`].
+    Exception,
+}
+
+/// The type of Data Sentry should ingest for a [`slog::Record`].
+#[allow(clippy::large_enum_variant)]
+pub enum RecordMapping {
+    /// Ignore the [`Record`]
```
jan-auer, line 22:
```suggestion
    /// Ignore the [`Record`].
```

sentry-slog/src/drain.rs
```diff
@@ -1,15 +1,87 @@
...
+        F: Fn(slog::Level) -> LevelFilter + Send + Sync + 'static,
+    {
+        self.filter = Box::new(filter);
+        self
+    }
+
+    /// Sets a custom mapper function.
+    ///
+    /// The mapper is responsible for creating either breadcrumbs or events
+    /// from [`Record`]s.
```
jan-auer, line 78: Here, it would be great to give a code example why and how to do this.

sentry-slog/src/integration.rs
jan-auer: Do you mean after this release? Since that would create another case where users have to update their code.

I do not view this as blocking, by the way. I think you can merge as-is, too.

jan-auer (approved):

sentry-log/src/lib.rs
Swatinem: not unless we break the MSRV, since `matches!` is a 1.42 feature.

sentry-slog/src/integration.rs
Swatinem: we can do it in a followup PR, still in the same release. I think one of the integrations was only recently added anyway.

PR #273, opened 2020-09-29
ci: Make testfast actually fast

Swatinem:
turns out, changing RUSTFLAGS invalidates all the caches, thats why the `checkfast/fastfast` job was not using it properly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,9 @@ jobs:
     name: checkall using ${{ matrix.rust }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
+    env:
+      RUSTFLAGS: -Dwarnings
+
     steps:
       - name: Checkout sources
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -120,7 +126,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v1
         with:
-          key: fast-MSRV
+          key: v0-fast-MSRV
```
jan-auer, line 129: Would it be enough to just wait for the next time we touch Cargo.toml or Cargo.lock, which should be within the next few days?

Swatinem: sure. I just want to verify this is actually working. I can revert that change

Swatinem:
lol, all the other tests are broken by https://github.com/http-rs/surf/issues/244 right now. So yes, we will definitely change our `Cargo.toml` soon, lol.

jan-auer (approved):

PR #276, opened 2020-10-03
Preliminary work to integrate Performance Monitoring 

aramperes:
Hello! I'm the lead developer at [Wavy Labs Inc.](https://github.com/WavyLabs/) We're integrating Rust more and more into our stack and we've been using Sentry for over a year. We'd love to start integrating Performance Monitoring, but it seems this SDK doesn't support it yet.

I'd like to kickstart the discussion on this and contribute some developer time for this. I've been reading the [Guidelines for Tracing Support](https://develop.sentry.dev/sdk/unified-api/tracing/) and peeking into the Python SDK for reference. I've started some preliminary work on adding the new types (Span and Transaction), but I'll need some guidance on integrating with Hub. Feel free to start reviewing at any time!

### DIFF
--- a/sentry-types/src/protocol/v7.rs
+++ b/sentry-types/src/protocol/v7.rs
@@ -1273,14 +1273,6 @@ mod event {
     }
 }
 
-/// Represents the type of `Event`
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
-pub enum EventType {
-    /// A `transaction` event, used for Performance Monitoring (tracing)
-    #[serde(rename = "transaction")]
-    Transaction,
-}
-
 /// Represents a full event for Sentry.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct Event<'a> {
@@ -1377,32 +1369,6 @@ pub struct Event<'a> {
     /// SDK metadata
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub sdk: Option<Cow<'a, ClientSdkInfo>>,
-    /// Type of event. Omitted for regular events.
-    #[serde(default, skip_serializing_if = "Option::is_none", rename = "type")]
-    pub _type: Option<EventType>,
-    /// A timestamp representing when the measuring started.
-    ///
-    /// Only used for tracing (`Transaction`) events.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub start_timestamp: Option<DateTime<Utc>>,
-    /// The ID of the trace.
-    ///
-    /// Only used for tracing (`Transaction`) events.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub trace_id: Option<Uuid>,
-    /// Longer description of the transaction's operation.
-    ///
-    /// Only used for tracing (`Transaction`) events.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub description: Option<String>,
-    /// Describes the status of the transaction.
-    ///
-    /// Only used for tracing (`Transaction`) events.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub status: Option<String>,
-    // TODO: Document this field
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub parent_span_id: Option<String>,
 }
 
 impl<'a> Default for Event<'a> {
@@ -1435,12 +1401,6 @@ impl<'a> Default for Event<'a> {
             extra: Default::default(),
             debug_meta: Default::default(),
             sdk: Default::default(),
-            _type: Default::default(),
-            start_timestamp: Default::default(),
-            trace_id: Default::default(),
-            description: Default::default(),
-            status: Default::default(),
-            parent_span_id: Default::default(),
         }
     }
 }
@@ -1486,12 +1446,6 @@ impl<'a> Event<'a> {
             extra: self.extra,
             debug_meta: Cow::Owned(self.debug_meta.into_owned()),
             sdk: self.sdk.map(|x| Cow::Owned(x.into_owned())),
-            _type: self._type,
-            start_timestamp: self.start_timestamp,
-            trace_id: self.trace_id,
-            description: self.description,
-            status: self.status,
-            parent_span_id: self.parent_span_id,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


sentry-types/src/protocol/v7.rs
```diff
@@ -1454,3 +1501,96 @@ impl<'a> fmt::Display for Event<'a> {
...
+        Utc::now()
+    }
+}
+
+/// Represents a tracing span.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct Span {
+    #[serde(default = "event::default_id", serialize_with = "event::serialize_id")]
+    pub event_id: Uuid,
+    #[serde(default = "event::default_id", serialize_with = "event::serialize_id")]
```
Swatinem, line 1522:
```suggestion
    #[serde(default = "span::default_id", serialize_with = "event::serialize_id")]
```
or rather remove the separate module?

Swatinem (commented):
I really appreciate you taking first steps here! Tracing is on our radar, we just wanted to gather more experience with our other SDKs before integrating it into Rust together will all the learnings from the other SDKs. That being said, I haven’t really looked at the implementation from other SDKs, or how it would best fit into Rust.

Nevertheless, looking at your draft and the Guidelines, I think that the `Transaction` type should rather live in `sentry-core`, as it should be able to "submit itself" to the client/transport. So it needs to have at least an `Arc<Client>` on itself.

Instead of "extending" the existing `Event` type, it should rather use composition, as in:

```rust
struct Transaction {
event: Event,
span: Span,
client: Arc<Client>,
}
```

It can then flatten all the event/span properties when serializing.
It can also have a `DerefMut` impl for both `Event` and `Span`.

I would also give it its own `EnvelopeItem` type.

Roping in @HazAT and @rhcarvalho as they should have more experience with the tracing API.

aramperes:
Glad to help! I understand that this may not have been on your team's roadmap in the immediate future. Nevertheless we'll try to contribute as much as possible.

Since `EnvelopeItem` lives in sentry-types, and `Client` lives in sentry-core, I don't think it's possible to include a client reference in `Transaction`.

This seems more like the sort of thing that could use a trait and `Hub::current`, e.g.

```rust
// In `sentry-types`

#[derive(Clone, Debug, Serialize)]
pub struct Transaction<'a> {
    event: Event<'a>,
    span: Span,
}
```

```rust
// In `sentry-core`

// Trait in case other types of envelopes can 'submit themselves'
pub trait Sendable { // [or another name]
	pub fn send();
}

impl<'a> Sendable for Transaction<'a> {
    fn send() {
        Hub::current().send_transaction(self)
    }
}
```

Swatinem (commented):
This looks much better now.

Ah yes, I moved all the `Envelope`-related things into `sentry-types`. We had a similar case with `Session`, and decided to split the POD-bits into the types crate, and a small wrapper into core. See https://github.com/getsentry/sentry-rust/pull/251
I think something similar can be done here as well.

This PR on its own looks very promising already, and I would be happy to land this incrementally, meaning landing the bits from types even if they are not yet fully usable.

aramperes:
Sounds good!

How about I finish this PR by:
* [ ] implementing TraceContext
* [ ] documenting the new structs and fields
* [ ] adding unit tests for serializing Transaction and Span to make sure everything is proper
* [ ] reverting the changes in `Scope` to be done in a later PR

Swatinem:
Yes that sounds very good

rhcarvalho:
FWIW we've updated the developer docs to include the `tracesSampler` option that gives more control over the sampling decision. Not a must have for a first pass, but something to keep in mind.

https://develop.sentry.dev/sdk/unified-api/tracing#sdk-configuration

(might be easier to read the diff -- https://github.com/getsentry/develop/pull/165)

aramperes:
@rhcarvalho Thanks for the update! I went through it and it's much clearer than before.

Just to double-check that this is essentially what the JSON structure of the envelope looks like for a Transcation event:

```jsonc
// Envelope Headers
{"event_id": "...", "sent_at": "..."}

// Item Headers
{"type": "transaction", ...}

// Item payload 
{
    // Transaction struct
    "type": "transaction",
    "transaction": "...",
    "tags": { ... },
    "timestamp": "...",
    "start_timestamp": "...",
    "spans": [
        { // Span struct
            "trace_id": "...",
            "span_id": "...",
            "parent_span_id": "...",
            "op": "...",
            "description": "...",
            "timestamp": "...",
            "start_timestamp": "...",
            "same_process_as_parent": false,
            "status": "...", // Optional
            "tags": { ... }, // Optional
            "data": { ... }, // Optional
        },
        // ...
    ],
    "contexts": {
        "trace": { // TraceContext struct
            "trace_id": "...",
            "span_id": "...",
            "parent_span_id": "...",
            "op": "...",
            "description": "...",
            "status": "..." // Optional
        }
    }
}
```

Swatinem:
> It also has a `type` field always set to `"transaction"`

I’m actually not sure if that is supposed to live on the transaction type itself, or if its meant to be just the envelope header.

aramperes:
I'll exclude it, but it does appear to be sent in other SDKs. It is used to identity the type of event in SDKs that don't have strict typing, e.g. in Python 'Event' is just a Dict, so this field is used to send the envelope differently. It wouldn't be necessary in Rust since Transaction has a different struct.

https://github.com/getsentry/sentry-python/blob/91c7a8fcb8e94b37e7dba74e66f7d0992f3cf145/sentry_sdk/client.py#L337-L352

It's possible that it gets ignored on the server-side however.

sentry-types/src/protocol/v7.rs
```diff
@@ -1454,3 +1482,186 @@ impl<'a> fmt::Display for Event<'a> {
...
+}
+
+impl Span {
+    /// Creates a new span with the current timestamp and random id.
+    pub fn new() -> Span {
+        Default::default()
+    }
+
+    /// Creates a fully owned version of the span.
+    pub fn into_owned(self) -> Span {
```
Swatinem, line 1550: what do we need this for? This is basically just `self -> Self`?

sentry-types/src/protocol/v7.rs
```diff
@@ -1454,3 +1482,186 @@ impl<'a> fmt::Display for Event<'a> {
...
+
+/// Represents a tracing transaction.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct Transaction<'a> {
+    /// The ID of the event
+    #[serde(default = "event::default_id", serialize_with = "event::serialize_id")]
+    pub event_id: Uuid,
+    /// The transaction name of the event.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub transaction: Option<String>,
```
Swatinem, line 1590: shouldn’t this be `name`?

Swatinem (commented):
after talking to @rhcarvalho, it looks like we need a `type` on the transaction object after all (at least when its serialized to json) I will research this myself to know how this should look like

Swatinem (approved):
Sorry for the delay here. Please rebase this on master again, the clippy error should be gone then.

This looks good from my perspective, I will ask @rhcarvalho for another review however, as he should know all the data structures related to tracing a bit better than me.

Swatinem:
Thanks a lot for doing this!

I’m wondering if we should maybe gate all of this behind an `unstable` feature flag, to indicate that we would want to iterate on these types a bit more before we consider them to be "stable" before we put them in a release.

PR #279, opened 2020-10-13
ref: Introduce a SessionFlusher

Swatinem:
The flusher maintains a queue that is flushed every minute.
Sessions are batched into Envelopes at this point, but in the future
might be pre-aggregated.

### DIFF
--- a/sentry-core/src/session.rs
+++ b/sentry-core/src/session.rs
@@ -2,9 +2,11 @@
 //!
 //! https://develop.sentry.dev/sdk/sessions/
 
-use std::sync::Arc;
-use std::time::Instant;
+use std::sync::{Arc, Condvar, Mutex};
+use std::thread::JoinHandle;
+use std::time::{Duration, Instant};
 
+use crate::client::TransportArc;
 use crate::protocol::{
     EnvelopeItem, Event, Level, SessionAttributes, SessionStatus, SessionUpdate,
 };
@@ -23,10 +25,8 @@ pub struct Session {
 impl Drop for Session {
     fn drop(&mut self) {
         self.close();
-        if let Some(item) = self.create_envelope_item() {
-            let mut envelope = Envelope::new();
-            envelope.add_item(item);
-            self.client.capture_envelope(envelope);
+        if self.dirty {
+            self.client.enqueue_session(self.session_update.clone());
         }
     }
 }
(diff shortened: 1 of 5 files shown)


sentry-core/src/session.rs
```diff
@@ -114,8 +114,109 @@ impl Session {
     }
 }
 
+// as defined here: https://develop.sentry.dev/sdk/envelopes/#size-limits
+const MAX_SESSION_ITEMS: usize = 100;
+const FLUSH_INTERVAL: u64 = 60;
```
jan-auer, line 119: _nit_: You could specify this as `Duration::from_secs` directly to avoid any misunderstanding.

sentry-core/src/session.rs
```diff
@@ -114,8 +114,109 @@ impl Session {
...
+
+type SessionQueue = Arc<Mutex<Vec<SessionUpdate<'static>>>>;
+
+pub(crate) struct SessionFlusher {
```
jan-auer, line 123: _nit_: Even though this is not public, it would be great to have doc comments explaining the high-level behavior of this type and its most important methods.

jan-auer (commented):
This looks good, thanks! 

One overall suggestion: Instead of always waiting for the flush interval, could we flush as soon as we reach the maximum item count?

sentry-core/src/session.rs
```diff
@@ -169,10 +178,24 @@ impl SessionFlusher {
...
+    /// When the queue is full, it will be flushed immediately.
     pub fn enqueue(&self, session_update: SessionUpdate<'static>) {
-        self.queue.lock().unwrap().push(session_update)
+        let flush_immediately = {
+            let mut queue = self.queue.lock().unwrap();
+            queue.push(session_update);
+            queue.len() >= MAX_SESSION_ITEMS
+        };
+        if flush_immediately {
+            SessionFlusher::flush(&self.queue, &self.transport);
```
jan-auer, line 191: This is potentially racy. Between checking the queue length and calling `flush`, another thread could enqueue and flush, or a flush could happen. 

To avoid this, consider keeping the lock and passing a lock guard into `flush`.

jan-auer (commented):
I'm realizing one more thing, although this is probably fine: Even if the application is idle or does not log any sessions, your background thread runs once every minute to check an empty queue. Since it is only once per minute, this is not that big of a deal.

The ideal solution would probably be to stop spinning if the queue is empty and wait for a CondVar which is triggered by the `enqueue` code. Since there's little benefit to this, I'm fine with keeping it this way.

jan-auer (approved):

Issue #277, opened 2020-10-07
sentry 0.20.x client doesn't log events


pjenvey:
With the alternative CurlHttpTransport, 0.20.x of the client doesn't log events. Reverting back to 0.19 solves the issue.

I'm unable to easily try the default transport in this app (likely due to it linking boringssl which clashes w/ openssl, causing seg faults). With no other reports of 0.20.x failing in this way, I suspect it may only be an issue w/ the non standard transport.

pjenvey:
debug output for 0.19 (successful):
```
[sentry] curl: > POST /api/.../store/ HTTP/1.1
Host: ...
Accept: */*
X-Sentry-Auth: Sentry sentry_key=..., sentry_version=7, sentry_timestamp=..., sentry_client=sentry.rust/0.19.1
Content-Type: application/json
Content-Length: 39285
[sentry] curl: {"event_id":"...",...}
[sentry] curl: < HTTP/1.1 200 OK
[sentry] curl: < Server: nginx
[sentry] curl: < Date: Fri, 06 Nov 2020 23:43:45 GMT
[sentry] curl: < Content-Type: application/json
[sentry] curl: < Content-Length: 41
[sentry] curl: < Vary: Accept-Encoding
[sentry] curl: < X-XSS-Protection: 1; mode=block
[sentry] curl: < X-Content-Type-Options: nosniff
[sentry] curl: < Content-Language: en
[sentry] curl: < Expires: Fri, 06 Nov 2020 23:43:45 GMT
[sentry] curl: < Vary: Accept-Language, Cookie
[sentry] curl: < Last-Modified: Fri, 06 Nov 2020 23:43:45 GMT
[sentry] curl: < Cache-Control: max-age=0
[sentry] curl: < X-Frame-Options: deny
[sentry] curl: < Strict-Transport-Security: max-age=31536000
[sentry] curl: < Via: 1.1 google
[sentry] curl: < Alt-Svc: clear
[sentry] curl: < 
```

debug output for 0.20's (failure):
```
[sentry] curl: > POST /api/.../envelope/ HTTP/1.1
Host: ...
Accept: */*
X-Sentry-Auth: Sentry sentry_key=..., sentry_version=7, sentry_timestamp=..., sentry_client=sentry.rust/0.20.1
Content-Type: application/json
Content-Length: 39372
[sentry] curl: {"event_id":"..."}
{"type":"event","length":39287}
{...}
[sentry] curl: < HTTP/1.1 403 OK
[...]

PR #282, opened 2020-10-17
feat: actix-web 3 support

### LINKED ISSUE #143: actix-web 3.0
Hey! First of all, thanks a lot for all these popular integrations. Really makes sentry a breeze to use in rust.

Been trying the [actix-web 1.0 rc](https://docs.rs/crate/actix-web/1.0.0-rc) that recently came up. Judging by their [1.0 migration guide](https://github.com/actix/actix-web/blob/master/MIGRATION.md#10) middleware should just be mounted slightly differently:

> instead of
```rust
    let app = App::new()
         .middleware(middleware::Logger::default())
```
> use .wrap() method
```rust
    let app = App::new()
         .wrap(middleware::Logger::default())
```

but there does appear to be more to it under the hood because the sentry middleware is the last thing for me that fails during the migration:

```rust
    .wrap(sentry_actix::SentryMiddleware::new())
```

>    |             .wrap(sentry_actix::SentryMiddleware::new())
>    |              ^^^^ the trait `actix_service::transform::IntoTransform<_, <impl actix_service::NewService as actix_service::NewService>::Service>` is not implemented for `sentry_actix::SentryMiddleware`

Haven't looked into what's actually needed to fix this here, but hopefully it's an easy impl :pray:

aramperes:
I refactored the middleware by basing it on 'official' v3 middleware (logger, cors, etc.) Some of the Sentry implementation was re-used from https://github.com/mozilla-services/autopush-rs/blob/9b93817776bb7c8f27383185a1b5ce08607a6b35/autoendpoint/src/middleware/sentry.rs (v2.0 middleware, adapted for v3.0 and `wrap`-style middleware).

Example is updated and tested locally.

Bumped minimum Rust version to 1.42.0 because that's what actix-web 3.0 targets.

Resolves #143

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [1.40.0]
+        rust: [1.42.0]
 
     name: checkfast/testfast using ${{ matrix.rust }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
--- a/sentry-actix/Cargo.toml
+++ b/sentry-actix/Cargo.toml
@@ -24,6 +24,7 @@ with_sentry_default = [
 [dependencies]
 sentry = { version = "0.20.1", path = "../sentry", default-features = false }
 sentry-failure = { version = "0.20.1", path = "../sentry-failure" }
-actix-web = { version = "0.7", default-features = false }
+actix-web = { version = "3", default-features = false }
 failure = "0.1.3"
+futures-util = "0.3.5"
 fragile = "0.3.0"
--- a/sentry-actix/Cargo.toml
+++ b/sentry-actix/Cargo.toml
@@ -27,5 +27,6 @@ actix-web = { version = "3", default-features = false }
 futures-util = "0.3.5"
 
 [dev-dependencies]
-sentry = { version = "0.20.1", path = "../sentry", default-features = false }
+sentry = { version = "0.20.1", path = "../sentry", default-features = false, features = ["test"] }
 actix-rt = "1.1.1"
+futures = "0.3"
--- a/sentry-actix/Cargo.toml
+++ b/sentry-actix/Cargo.toml
@@ -28,3 +28,4 @@ futures-util = "0.3.5"
 
 [dev-dependencies]
 sentry = { version = "0.20.1", path = "../sentry", default-features = false }
+actix-rt = "1.1.1"
--- a/sentry-actix/Cargo.toml
+++ b/sentry-actix/Cargo.toml
@@ -22,9 +22,9 @@ with_sentry_default = [
 ]
 
 [dependencies]
-sentry = { version = "0.20.1", path = "../sentry", default-features = false }
-sentry-failure = { version = "0.20.1", path = "../sentry-failure" }
+sentry-core = { version = "0.20.1", path = "../sentry-core", default-features = false }
 actix-web = { version = "3", default-features = false }
-failure = "0.1.3"
 futures-util = "0.3.5"
-fragile = "0.3.0"
+
+[dev-dependencies]
+sentry = { version = "0.20.1", path = "../sentry", default-features = false }
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -306,15 +306,11 @@ mod tests {
 
     use actix_web::test::{call_service, init_service, TestRequest};
     use actix_web::{get, web, App, HttpRequest, HttpResponse};
+    use futures::executor::block_on;
 
-    use sentry::{ClientInitGuard, Level};
+    use sentry::Level;
 
     use super::*;
-
-    fn _init() -> ClientInitGuard {
-        // Initialize with a 'valid' DSN so errors are captured
-        sentry::init("https://public@sentry.invalid/1")
-    }
 
     fn _assert_hub_no_events() {
         if Hub::current().last_event_id().is_some() {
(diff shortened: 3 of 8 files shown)


aramperes:
I think the build is failing because `actix-utils` (for actix-web 3.0) uses the `matches!` macro, which was only stabilised in 1.42.0. Is it okay to increment the minimum stable version in this case?

OskarPersson:
I've tried this branch with the provided example locally on my machine and can't get it to produce an issue on Sentry. The error message is rendered as expected in the browser but no issue is created.

onelson:
@OskarPersson I wonder if there could be any overlap with #239 or #277

aramperes:
@OskarPersson @onelson 

I ran the `basic.rs` file (after changing the DSN to a real sentry.io one), and the event/issue is created properly:

image

Can you provide a counter-example? Maybe your machine specs, Sentry version, Rust version, etc?

sentry-actix/Cargo.toml
```diff
@@ -24,6 +24,7 @@ with_sentry_default = [
 [dependencies]
 sentry = { version = "0.20.1", path = "../sentry", default-features = false }
 sentry-failure = { version = "0.20.1", path = "../sentry-failure" }
-actix-web = { version = "0.7", default-features = false }
+actix-web = { version = "3", default-features = false }
 failure = "0.1.3"
+futures-util = "0.3.5"
 fragile = "0.3.0"
```
Swatinem, line 30: this is also unused now

sentry-actix/src/lib.rs
```diff
@@ -68,221 +71,267 @@
...
-/// Reports certain failures to sentry.
-pub struct SentryMiddleware {
+struct Inner {
     hub: Option<Arc<Hub>>,
     emit_header: bool,
     capture_server_errors: bool,
 }
 
+/// Reports certain failures to Sentry.
+pub struct Sentry(Rc<Inner>);
```
Swatinem, line 139: Its just super weird seeing `Rc` in async code… I think its fine because of `Pin`?
Otherwise, I think its fine to just derive `Clone` here instead of using the Rc.

sentry-actix/src/lib.rs
```diff
@@ -68,221 +71,267 @@
...
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: ServiceRequest) -> Self::Future {
+        let inner = self.inner.clone();
+        let hub = Arc::new(Hub::new_from_top(Hub::main()));
```
Swatinem, line 217: I wonder if this has been broken all along. Why does the builder have options to set a specific Hub if we derive from the Main Hub anyway…

sentry-actix/src/lib.rs
```diff
@@ -68,221 +71,267 @@
...
+        let guard = hub.push_scope();
+
+        let (tx, sentry_req) = sentry_request_from_http(&req, with_pii);
+        hub.configure_scope(|scope| {
+            scope.add_event_processor(Box::new(move |event| {
+                process_event(event, tx.clone(), &sentry_req)
+            }))
+        });
+
+        req.extensions_mut().insert(HubWrapper { hub: hub.clone() });
```
Swatinem, line 231: Since sentry has futures support now anyway, it might be a good idea to use that, and bind the Hub to the future.
If I understand correctly, the new actix is completely based on Futures, right? That would mean that people can use `Hub::current/with` etc and get the correct hub, instead of having an actix specific method to get the Hub.

sentry-actix/src/lib.rs
```diff
@@ -68,221 +71,267 @@
...
+
+        let fut = self.service.call(req);
+
+        async move {
+            // Service errors
+            let mut res: Self::Response = match fut.await {
+                Ok(res) => res,
+                Err(e) => {
+                    if inner.capture_server_errors {
+                        hub.capture_event(sentry::event_from_error(&e));
```
Swatinem, line 241: Is `actix_web::Error` a `std::error::Error`? If so, you can just use `sentry::capture_error()` here?

sentry-actix/src/lib.rs
```diff
@@ -315,34 +364,6 @@ impl ActixWebHubExt for Hub {
...
-            ptr = fail.cause();
-            idx += 1;
-        }
-        exceptions.reverse();
-        self.capture_event(Event {
-            exception: exceptions.into(),
-            level: Level::Error,
-            ..Default::default()
-        })
+        self.capture_event(sentry::event_from_error(err))
```
Swatinem, line 367: related to the question above, if these are `std::error::Error` anyway, we can also remove this method. (Along with the complete `ActixWebHubExt` if we switch to the builtin futures support)

sentry-actix/Cargo.toml
```diff
@@ -24,6 +24,7 @@ with_sentry_default = [
 [dependencies]
 sentry = { version = "0.20.1", path = "../sentry", default-features = false }
```
Swatinem, line 25: I think we can change this to depend on `sentry-core` instead. (with a dev-dependency on `sentry` for doctests)

Swatinem (commented):
This is awesome!

I left a few comments where I think we can further simplify things.

.github/workflows/ci.yml
```diff
@@ -105,7 +105,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [1.40.0]
+        rust: [1.42.0]
```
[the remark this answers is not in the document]
aramperes: In this case the workspace doesn't build unless the Rust version is bumped. Is there another way to handle it?

Thanks for the code review! I'll do the changes asap.

sentry-actix/src/lib.rs
aramperes: Since `ActixWebHubExt` was removed, there isn't a need to store the Hub in the extensions anymore. However, to make it possible for users to access the current Hub through `Hub::current`, is there a function to 'bind the Hub to the future'?

sentry-actix/src/lib.rs
aramperes: That's my mistake, it should use the given Hub if specified and fall-back to the Main hub otherwise.

Is using `Hub::current` sufficient to get the hub or is it necessary to use `Hub::new_from_top`? See latest commit for context.

.github/workflows/ci.yml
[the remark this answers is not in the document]
Swatinem: Oh yes, that testcase runs on the workspace, I thought it would only run on the `sentry` crate…
I’m fine with just bumping MSRV, unless someone else from the team objects.

sentry-actix/src/lib.rs
Swatinem: yes, thats what I meant: https://github.com/getsentry/sentry-rust/blob/c8ef5fecea34215776863b0333a734c247a6f23b/sentry-core/src/futures.rs#L55

sentry-actix/src/lib.rs
Swatinem: You should still call `new_from_top`, otherwise concurrent futures will mess with each others scope, such as breadcrumbs.
Also, `add_event_processor` callbacks would multiply and leak for each request.

sentry-actix/src/lib.rs
aramperes: Just to make sure this is the change you're suggesting:

```rust
let fut = self.service.call(req).bind_hub(hub.clone());
```

sentry-actix/src/lib.rs
```diff
@@ -119,230 +106,200 @@ impl SentryMiddlewareBuilder {
...
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: ServiceRequest) -> Self::Future {
+        let inner = self.inner.clone();
+        let hub = inner
+            .hub
+            .clone()
+            .unwrap_or_else(|| Arc::new(Hub::new_from_top(Hub::main())));
```
Swatinem, line 192: This should rather be `Hub::new_from_top(inner.hub.unwrap_or_else(Hub::main))`

The same comment about multiple requests pushing and poping scopes from the *same* hub in arbitrary order applies here. Each request has to have its own Hub instance.

Also, I think we don’t need to use `push_scope` at all in that case.

Swatinem:
Some unit tests for this would be nice, also to make sure that concurrent requests don’t clash with each other.

aramperes:
@Swatinem I'll look into adding some unit tests combining `actix_rt::test` and Sentry.

aramperes:
I added 2 unit tests: one that tests explicit events via `sentry::capture_message`, and the other for events dispatched by the middleware through service errors.

Both tests also ensure the Hubs are isolated by checking whether `last_event_id` is set in different contexts, and finally for the main Hub to ensure it was never accessed.

sentry-actix/src/lib.rs
```diff
@@ -299,3 +299,120 @@ fn process_event(
...
+
+            // Add processor to ensure proper event fields
+            sentry::configure_scope(|scope| {
+                scope.add_event_processor(Box::new(move |event| {
+                    let request = event.request.expect("Request should be set.");
+                    assert_eq!(event.transaction, Some("/test".into()));
+                    assert_eq!(event.message, Some("Message".into()));
+                    assert_eq!(event.level, Level::Warning);
+                    assert_eq!(request.method, Some("GET".into()));
+                    None
```
Swatinem, line 348: I’m a bit confused why you would get a last_event_id if you discard the event in the processor.

sentry-actix/src/lib.rs
```diff
@@ -299,3 +299,120 @@ fn process_event(
...
+            App::new()
+                .wrap(Sentry::new())
+                .service(web::resource("/test").to(service)),
+        )
+        .await;
+
+        // Call the service twice to check the isolation between them
+        for _ in 0..2 {
+            let req = TestRequest::get().uri("/test").to_request();
+            let res = call_service(&mut app, req).await;
```
Swatinem, line 371: doesn’t the await here make these sequencial instead of concurrent?

Swatinem (commented):
Sorry for the churn here.

I wonder if we can use the `sentry::test` module to capture the events (not sure if it uses `main` or `current`), and then just fire some requests and `block_on` having them all complete.

sentry-actix/src/lib.rs
aramperes: Returning None discards the event, but it will still set the `last_event_id` to a `Default::default` Uuid in that case. It's a _bit_ of a hack, but it's the only way I could find to ensure the event was captured in the current Hub.

`sentry::test::with_captured_events` looks like a promising replacement, but the hurdle is that it creates a new Hub instance with an alternative transport to capture all events within it. It should still be feasible by passing the mock Hub to `SentryBuilder::with_hub`. I'll take a look!

sentry-actix/src/lib.rs
aramperes: Yes I was mainly testing the isolation from sequential requests (i.e. making sure the middleware isn't sticky).

Testing the concurrent isolation would be a different beast, I think. How would you determine that both requests used different Hub instances? Is there a way to identify individual Hubs?

Swatinem (approved):

sentry-actix/src/lib.rs
Swatinem: > Is there a way to identify individual Hubs?

you could use `configure_scope` to set different properties. But I think this is very good now

Swatinem:
This is amazing! Thanks so much 🚀

Daggy1234:
Thanks been waiting for this!

PR #284, opened 2020-10-20
feat: Add support for pre-aggregated Sessions

Swatinem:
This should work together with https://github.com/getsentry/relay/pull/815

### DIFF
--- a/sentry-core/src/clientoptions.rs
+++ b/sentry-core/src/clientoptions.rs
@@ -11,6 +11,32 @@ use crate::{Integration, IntoDsn, TransportFactory};
 /// Type alias for before event/breadcrumb handlers.
 pub type BeforeCallback<T> = Arc<dyn Fn(T) -> Option<T> + Send + Sync>;
 
+/// The Session Mode of the SDK.
+///
+/// Depending on the use-case, the SDK can be set to two different session modes:
+///
+/// * **Application Mode Sessions**:
+///   This mode should be used for user-attended programs, which typically have
+///   a single long running session that span the applications' lifetime.
+///
+/// * **Request Mode Sessions**:
+///   This mode is intended for servers that use one session per incoming
+///   request, and thus have a lot of very short lived sessions.
+///
+/// Setting the SDK to *request-mode* sessions means that session durations will
+/// not be tracked, and sessions will be pre-aggregated before being sent upstream.
+/// This applies both to automatic and manually triggered sessions.
+///
+/// See the [Documentation on Session Modes](https://develop.sentry.dev/sdk/sessions/#sdk-considerations)
+/// for more information.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum SessionMode {
+    /// Long running application session.
+    Application,
+    /// Lots of short per-request sessions.
+    Request,
+}
+
 /// Configuration settings for the client.
 ///
 /// These options are explained in more detail in the general
@@ -97,6 +123,8 @@ pub struct ClientOptions {
     /// is started at the time of `sentry::init`, and will persist for the
     /// application lifetime.
     pub auto_session_tracking: bool,
+    /// Determine how Sessions are being tracked.
+    pub session_mode: SessionMode,
     /// Border frames which indicate a border from a backtrace to
     /// useless internals. Some are automatically included.
     pub extra_border_frames: Vec<&'static str>,
@@ -164,6 +192,7 @@ impl fmt::Debug for ClientOptions {
             .field("https_proxy", &self.https_proxy)
             .field("shutdown_timeout", &self.shutdown_timeout)
             .field("auto_session_tracking", &self.auto_session_tracking)
+            .field("session_mode", &self.session_mode)
             .field("extra_border_frames", &self.extra_border_frames)
             .field("trim_backtraces", &self.trim_backtraces)
             .field("user_agent", &self.user_agent)
@@ -194,6 +223,7 @@ impl Default for ClientOptions {
             https_proxy: None,
             shutdown_timeout: Duration::from_secs(2),
             auto_session_tracking: false,
+            session_mode: SessionMode::Application,
             extra_border_frames: vec![],
             trim_backtraces: true,
             user_agent: Cow::Borrowed(&USER_AGENT),
--- a/sentry-core/src/clientoptions.rs
+++ b/sentry-core/src/clientoptions.rs
@@ -27,6 +27,8 @@ pub type BeforeCallback<T> = Arc<dyn Fn(T) -> Option<T> + Send + Sync>;
 /// not be tracked, and sessions will be pre-aggregated before being sent upstream.
 /// This applies both to automatic and manually triggered sessions.
 ///
+/// **NOTE**: Support for *request-mode* sessions was added in Sentry `21.2`.
+///
 /// See the [Documentation on Session Modes](https://develop.sentry.dev/sdk/sessions/#sdk-considerations)
 /// for more information.
 #[derive(Copy, Clone, Debug, PartialEq)]
--- a/sentry-core/src/session.rs
+++ b/sentry-core/src/session.rs
@@ -2,16 +2,21 @@
 //!
 //! <https://develop.sentry.dev/sdk/sessions/>
 
+use std::collections::HashMap;
 use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 use std::thread::JoinHandle;
 use std::time::{Duration, Instant};
 
+use chrono::{Duration as ChronoDuration, DurationRound};
+
 use crate::client::TransportArc;
+use crate::clientoptions::SessionMode;
 use crate::protocol::{
-    EnvelopeItem, Event, Level, SessionAttributes, SessionStatus, SessionUpdate,
+    EnvelopeItem, Event, Level, SessionAggregateItem, SessionAggregates, SessionAttributes,
+    SessionStatus, SessionUpdate,
 };
 use crate::scope::StackLayer;
-use crate::types::{Utc, Uuid};
+use crate::types::{DateTime, Utc, Uuid};
 use crate::{Client, Envelope};
 
 #[derive(Clone, Debug)]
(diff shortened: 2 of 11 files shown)


sentry-core/src/hub.rs
```diff
@@ -337,16 +337,8 @@ impl Hub {
...
-                    if let Some(item) = session.create_envelope_item() {
-                        let mut envelope = Envelope::new();
-                        envelope.add_item(item);
-                        if let Some(ref client) = top.client {
-                            client.capture_envelope(envelope);
-                        }
-                    }
-                }
+                // drop will close and enqueue the session
+                top.scope.session.lock().unwrap().take();
```
Swatinem, line 341: this was actually missing from #279, as manually calling `end_session` would not go through the flusher but eagerly send envelopes.

sentry-core/src/session.rs
```diff
@@ -184,23 +232,77 @@ impl SessionFlusher {
...
+            }
+            return;
+        }
+
+        let aggregate = queue.aggregated.get_or_insert_with(|| AggregatedSessions {
+            buckets: HashMap::with_capacity(1),
+            attributes: session_update.attributes.clone(),
+        });
+
+        use chrono::{Duration, DurationRound};
```
flub, line 252: why not at the top of the file?  You could rename to `ChronoDuration` or so

sentry-core/src/session.rs
```diff
@@ -293,11 +397,94 @@ mod tests {
...
+                // this error will be discarded because of the event processor,
+                // but the session will still be updated accordingly.
+                let err = "NaN".parse::<usize>().unwrap_err();
+                sentry::capture_error(&err);
+            },
+            crate::ClientOptions {
+                release: Some("some-release".into()),
+                session_mode: SessionMode::Request,
+                ..Default::default()
+            },
```
flub, line 439: Not really relevant, but having this as 2nd arg kind of reads the wrong way round.  I'd rather know this before starting to read the code that'll execute with this configuration.

sentry-core/src/session.rs
```diff
@@ -293,11 +397,94 @@ mod tests {
...
+        );
+        assert_eq!(envelopes.len(), 3);
+
+        let session_id;
```
flub, line 443: huh?  shouldn't this need to be `mut`?

sentry-core/src/session.rs
```diff
@@ -293,11 +397,94 @@ mod tests {
...
+        let mut items = envelopes[0].items();
+        assert!(matches!(items.next(), Some(EnvelopeItem::Event(_))));
+        if let Some(EnvelopeItem::SessionUpdate(session)) = items.next() {
+            session_id = session.session_id;
+            assert_eq!(session.status, SessionStatus::Ok);
+            assert_eq!(session.errors, 1);
+            assert_eq!(session.init, true);
+        } else {
+            panic!("expected session");
+        }
```
flub, line 454: `items.next().expect("expected session")` would be more concise?

sentry-types/src/protocol/session.rs
```diff
@@ -130,3 +130,44 @@ pub struct SessionUpdate<'a> {
...
 }
+
+#[allow(clippy::trivially_copy_pass_by_ref)]
+fn is_zero(val: &u32) -> bool {
+    *val == 0
+}
+
+/// An aggregation grouped by `started` and `distinct_id`.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct SessionAggregateItem {
```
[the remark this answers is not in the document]
jan-auer: No, this is still work in progress, so we haven't published documentation on this yet. To your point, though, at least the current state can be documented @Swatinem, with an "unstable" or "preview" note.

sentry-core/src/session.rs
Swatinem: you can split declaration from initialization. the compiler makes sure its definitely initialized (in a different block). you only need `mut` if you re-initialize (overwrite)

sentry-core/src/session.rs
Swatinem: well no, since the next item can also be an event.

sentry-core/src/session.rs
Swatinem: agreed, although that would be a breaking change now :-(

sentry-types/src/protocol/session.rs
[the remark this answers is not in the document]
Swatinem: I have the docs prepared since quite some time: https://github.com/getsentry/develop/pull/136
I can update them with a NOTE saying that while this is supported by now in relay, its still in preview and severely limited.

sentry-core/src/session.rs
flub: ooh, fair enough

sentry-core/src/session.rs
flub: that's a public api?  i thought this was just some test code.  oh well

sentry-core/src/clientoptions.rs
```diff
@@ -91,6 +116,8 @@ pub struct ClientOptions {
     /// is started at the time of `sentry::init`, and will persist for the
     /// application lifetime.
     pub auto_session_tracking: bool,
+    /// Determine how Sessions are being tracked.
+    pub session_mode: SessionMode,
```
jan-auer, line 120: Could we consider combining this with the session tracking flag into a `session_mode: Application | Request | Off`?

sentry-core/src/clientoptions.rs
```diff
@@ -11,6 +11,31 @@ use crate::{Integration, IntoDsn, TransportFactory};
 /// Type alias for before event/breadcrumb handlers.
 pub type BeforeCallback<T> = Arc<dyn Fn(T) -> Option<T> + Send + Sync>;
 
+/// The Session Mode of the SDK.
+///
+/// Depending on the use-case, the SDK can be set to two different session modes:
+///
+/// * **Application Mode Sessions**:
+///   This mode should be used for user-attended programs, which typically have
+///   a single long running session.
```
jan-auer, line 20: _nit_: You could mention that the session spans the "application's" lifetime, which should make it more apparent why this is called "Application Mode".

sentry-core/src/session.rs
```diff
@@ -122,27 +125,71 @@ impl Session {
 const MAX_SESSION_ITEMS: usize = 100;
 const FLUSH_INTERVAL: Duration = Duration::from_secs(60);
 
-type SessionQueue = Arc<Mutex<Vec<SessionUpdate<'static>>>>;
+#[derive(Default)]
```
jan-auer, line 128: _nit_: Implement `Debug` on those structs.

sentry-core/src/session.rs
```diff
@@ -184,23 +232,77 @@ impl SessionFlusher {
...
     ///
-    /// When the queue is full, it will be flushed immediately.
+    /// This will aggregate session counts in request mode, for all sessions
+    /// that were not yet partially sent.
     pub fn enqueue(&self, session_update: SessionUpdate<'static>) {
         let mut queue = self.queue.lock().unwrap();
-        queue.push(session_update);
-        if queue.len() >= MAX_SESSION_ITEMS {
-            SessionFlusher::flush(queue, &self.transport);
+        if self.mode == SessionMode::Application || !session_update.init {
```
jan-auer, line 239: Shouldn't you always enqueue an update if there is a session? Even in Request mode, one may want to push an individual session with full attributes.

sentry-core/src/session.rs
```diff
@@ -184,23 +232,77 @@ impl SessionFlusher {
...
+
+        let aggregate = queue.aggregated.get_or_insert_with(|| AggregatedSessions {
+            buckets: HashMap::with_capacity(1),
+            attributes: session_update.attributes.clone(),
+        });
+
+        use chrono::{Duration, DurationRound};
+        let started = session_update
+            .started
+            .duration_trunc(Duration::hours(1))
```
jan-auer, line 255: Where is this hour coming from?

sentry-core/src/session.rs
```diff
@@ -184,23 +232,77 @@ impl SessionFlusher {
...
+                }
+            }
+            SessionStatus::Crashed => {
+                bucket.crashed += 1;
+            }
+            SessionStatus::Abnormal => {
+                bucket.abnormal += 1;
+            }
+            SessionStatus::Ok => {
+                unreachable!("only closed sessions will be enqueued");
```
jan-auer, line 280: Please log an internal error to avoid crashing a user's application.

sentry-core/src/session.rs
```diff
@@ -184,23 +232,77 @@ impl SessionFlusher {
...
     /// Flushes the queue to the transport.
     ///
     /// This is a static method as it will be called from both the background
     /// thread and the main thread on drop.
-    fn flush(mut queue_lock: MutexGuard<Vec<SessionUpdate<'static>>>, transport: &TransportArc) {
-        let queue: Vec<_> = std::mem::take(queue_lock.as_mut());
+    fn flush(mut queue_lock: MutexGuard<SessionQueue>, transport: &TransportArc) {
+        let (queue, aggregate) = (
+            std::mem::take(&mut queue_lock.individual),
+            queue_lock.aggregated.take(),
```
jan-auer, line 292: Since you have a mutex on the queue, can you assign this in two regular statements?

sentry-core/src/clientoptions.rs
Swatinem: I thought about this a bit, and a tri-state would not work when you don’t want to have auto-tracking, but want to manually track request-mode sessions.

Also related to your other question: Request-Mode applies to all sessions, I updated the doc comment to accordingly. So choosing between application/request-mode basically means that you make a choice if you want aggregation at the cost of losing durations, or not.

sentry-types/src/protocol/session.rs
```diff
@@ -130,3 +130,44 @@ pub struct SessionUpdate<'a> {
...
+    /// The timestamp of when the session itself started.
+    pub started: DateTime<Utc>,
```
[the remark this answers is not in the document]
Swatinem: The started time of the bucket.

jan-auer (approved):

PR #287, opened 2020-10-21
fix: scope transaction name can be overriden in sentry-actix

aramperes:
Configure the transaction name in the actix middleware at the `scope` level, instead of the event processor.

This fixes the middleware overriding the transaction name when capturing events, and allows users to set a different transaction in scope and sub-scopes.

### DIFF
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -197,9 +197,8 @@ where
 
         let (tx, sentry_req) = sentry_request_from_http(&req, with_pii);
         hub.configure_scope(|scope| {
-            scope.add_event_processor(Box::new(move |event| {
-                process_event(event, tx.clone(), &sentry_req)
-            }))
+            scope.set_transaction(tx.as_deref());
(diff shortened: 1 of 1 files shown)


sentry-actix/src/lib.rs
```diff
@@ -411,4 +403,41 @@ mod tests {
...
+    /// Ensures transaction name can be overridden in handler scope.
+    #[actix_rt::test]
+    async fn test_override_transaction_name() {
+        let events = sentry::test::with_captured_events(|| {
+            block_on(async {
+                #[get("/test")]
+                async fn original_transaction(_req: HttpRequest) -> Result<String, Error> {
+                    // Override transaction name
+                    Hub::current()
+                        .configure_scope(|scope| scope.set_transaction(Some("new_transaction")));
```
Swatinem, line 416:
```suggestion
                    sentry::configure_scope(|scope| scope.set_transaction(Some("new_transaction")));
```
This will dispatch to the current hub

Swatinem (approved):
you will need to rebase on top of your other PR now

PR #288, opened 2020-10-22
tracing: client options and Hub/Scope API

aramperes:
Hi again! I'm opening this PR to further the discussion on integrating tracing/APM here. I'd like to scope this PR to implement the basics of the tracing API. This includes

- [x] Adding client options for tracing: `traces_sample_rate` and `traces_sampler`
- [x] Adding the required API functions in `Scope`: `set_span` and `get_transaction` (WIP)
- [x] Hub changes: add `trace_headers`, `start_transaction`
- [x] Static API: add `start_transaction`
- [ ] Dispatching transactions in Client & Transport

For now I'd appreciate a review on how I implemented the new functions in `Scope`. I'm still unsure of how the `Span <-> Transaction` relation will be implemented, so I assumed `Span.transaction` would be `Arc<Transaction<'static>>`. However, that makes the transaction immutable and thus breaks this [guideline](https://develop.sentry.dev/sdk/unified-api/tracing/#scope-changes):

> Setting the transaction property on the Scope (legacy) should overwrite the name of the Transaction stored in the Scope, if there is one. With that we give users the option to change the transaction name even if they don't have access to the instance of the Transaction directly.

Alternatively, the transaction could be owned by the Scope instead of propagating references in individual Spans. Not sure if that's preferred.

Also if there's another preferred medium for chatting about design stuff like this let me know ^.^


sentry-core/src/clientoptions.rs
```diff
@@ -98,6 +99,16 @@ pub struct ClientOptions {
     pub trim_backtraces: bool,
     /// The user agent that should be reported.
     pub user_agent: Cow<'static, str>,
+    /// The sample rate for traces. (if set, must be inside the 0.0 - 1.0 range)
+    ///
+    /// Tracing is enabled if either this or `traces_sampler` is defined. If both
+    /// are defined, `traces_sample_rate` is ignored.
+    pub traces_sample_rate: Option<f32>,
```
Swatinem, line 106: I’m thinking if we should rather default this to `0.0` instead of having it as `Option`.

sentry-core/src/scope/real.rs
```diff
@@ -171,9 +174,41 @@ impl Scope {
...
+    #[deprecated(note = "Use set_transaction_name instead")]
     pub fn set_transaction(&mut self, transaction: Option<&str>) {
-        self.transaction = transaction.map(|txn| Arc::new(txn.to_string()));
+        self.set_transaction_name(transaction)
+    }
+
+    /// Sets the transaction name.
+    pub fn set_transaction_name(&mut self, name: Option<&str>) {
+        self.transaction = name.map(|txn| Arc::new(txn.to_string()));
+        // TODO: Update transaction name (not feasible with Arc<Transaction<'static>>)
```
Swatinem, line 186: Or we would have to wrap it in a rwlock maybe?

sentry-core/src/tracing.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Data passed to the `traces_sampler` function,
+/// which forms the basis for whatever decisions it might make.
+///
+/// TODO: This is a placeholder.
+#[derive(Debug, Default, Clone, Eq, PartialEq)]
+#[non_exhaustive]
+pub struct SamplingContext {}
+
+/// Function to compute tracing sample rate dynamically and filter unwanted traces.
+pub type TraceSampler = Arc<dyn TraceSamplerInner>;
```
Swatinem, line 13: I would rather explicitly use `Arc<dyn …>` in places where it is used instead of having an `…Inner` trait and type alias.
Ideally in the future, we would have functions for this instead of having people directly write to the props of `ClientOptions`.

aramperes: My understanding is that `Fn` doesn't implement `Debug`, so trait aliasing is necessary. I removed the need for the `Inner` trait but the `TraceSampler` still needs to be there to implement `Debug`.

At least with these changes, the need for `Arc` will be more evident to end-users.

aramperes:
Apologies for the staleness, I've been busy contributing to other projects but will get back to this ASAP

minyov:
Hello.
Looking forward to this PR merged.
I also want to suggest adding getter to retrieve `trace_id` from Scope.
It would be very useful if you work with different microservices and need to pass `trace_id` further.

aspenluxxxy:
Has there been any progress on this? I would find this rather useful in my projects :)

aramperes:
I can see there is some interest in this, but I'm choosing to close it. The current SDK architecture for tracing is heavily dependent on OOP concepts, and IMO the [SDK Guidelines for Tracing](https://develop.sentry.dev/sdk/performance/) will need a lot of clarifying for them to apply to a wider set of languages.

The current implementations are in Python and JavaScript, which allow for a lot of liberties that Rust simply cannot take. There's also a lot of backward-compatibility legacy code in them which makes it even harder to understand what's going on.

Of course, actual Sentry developers may have more insight in implementing the required architecture changes, but I feel like I'm constantly fighting the language to get any results. Hopefully some of the code here will become useful down the line to another contributor or to the Sentry team.

aspenluxxxy:
Oh, alright, that's understandable.

PR #289, opened 2020-11-11
feat: End sessions with explicit status

Swatinem:
This adds the `end_session_with` global and Hub functions that allow
providing an explicit `SessionStatus`.

### DIFF
--- a/sentry-core/src/api.rs
+++ b/sentry-core/src/api.rs
@@ -285,7 +285,7 @@ pub fn start_session() {
 
 /// End the current Release Health Session.
 pub fn end_session() {
-    end_session_with(SessionStatus::Exited)
+    end_session_with_status(SessionStatus::Exited)
 }
 
 /// End the current Release Health Session with the given [`SessionStatus`].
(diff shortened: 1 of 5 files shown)


sentry-core/src/api.rs
```diff
@@ -283,5 +285,16 @@ pub fn start_session() {
...
+}
+
+/// End the current Release Health Session with the given [`SessionStatus`].
+///
+/// By default, the SDK will only consider the `Exited` and `Crashed` status
+/// based on the type of events that were captured during the session.
+///
+/// When an `Abnormal` session should be captured, it has to be done explicitly
+/// using this function.
+pub fn end_session_with(status: SessionStatus) {
```
flub, line 298: bikeshedding: an api ending in `something_with()` or a bare `with()` I expect to take an `FnOnce` argument.  If the name is `with_something()` is where I expect the arg to be something else than a function.

So maybe `end_session_with_status()` is better?  It is a mouthfull though.  And `.end_session_with_status(SessionStatus::Crashed)` is also rather redundant.  So perhaps not the best suggestion.

Swatinem: I was thinking the exact same thing, in the end, I went with the shorter version, even though as you said, plain `X_with` usually takes a closure.

But I think `_with_status` makes sense in case we want to add more combinators later on.

flub (approved):

PR #291, opened 2020-11-18
docs: Document integrations and the Hub better

flub:
This adds more documentation for how to use integrations and uses much
more intra-doc links.  The Hub documentation as well as the anyhow
integration documentation have also seen various improvements.

The docs no build again without any warnings.

#skip-changelog

### DIFF
--- a/sentry-anyhow/src/lib.rs
+++ b/sentry-anyhow/src/lib.rs
@@ -1,4 +1,17 @@
-//! Adds support for capturing Sentry errors from `anyhow::Error`.
+//! Adds support for capturing Sentry errors from [`anyhow::Error`].
+//!
+//! This integration adds a new event *source*, which allows you to create events directly
+//! from an [`anyhow::Error`] struct.  As it is only an event source it only needs to be
+//! enabled using the `anyhow` cargo feature, it does not need to be enabled in the call to
+//! [`sentry::init`](../../fn.init.html).
+//!
+//! This integration does not need to be installed, instead it provides an extra function to
+//! capture [`anyhow::Error`], optionally exposing it as a method on the
+//! [`sentry::Hub`](../../struct.Hub.html) using the [`AnyhowHubExt`] trait.
+//!
+//! Like a plain [`std::error::Error`] being captured, [`anyhow::Error`] is captured with a
+//! chain of all error sources, if present.  See
+//! [`sentry::capture_error`](../../fn.capture_error.html) for details of this.
 //!
 //! # Example
 //!
@@ -13,6 +26,8 @@
 //!     capture_anyhow(&err);
 //! }
 //! ```
+//!
+//! [`anyhow::Error`]: https://docs.rs/anyhow/*/anyhow/struct.Error.html
 
 #![doc(html_favicon_url = "https://sentry-brand.storage.googleapis.com/favicon.ico")]
 #![doc(html_logo_url = "https://sentry-brand.storage.googleapis.com/sentry-glyph-black.png")]
@@ -22,16 +37,27 @@
 use sentry_core::types::Uuid;
 use sentry_core::Hub;
 
-/// Captures an `anyhow::Error`.
+/// Captures an [`anyhow::Error`].
+///
+/// This will capture an anyhow error as a sentry event if a
+/// [`sentry::Client`](../../struct.Client.html) is initialised, otherwise it will be a
+/// no-op.  The event is dispatched to the thread-local hub, with semantics as described in
+/// [`Hub::current`].
 ///
 /// See [module level documentation](index.html) for more information.
+///
+/// [`anyhow::Error`]: https://docs.rs/anyhow/*/anyhow/struct.Error.html
 pub fn capture_anyhow(e: &anyhow::Error) -> Uuid {
     Hub::with_active(|hub| hub.capture_anyhow(e))
 }
 
-/// Hub extension methods for working with `anyhow`.
+/// Hub extension methods for working with [`anyhow`].
+///
+/// [`anyhow`]: https://docs.rs/anyhow
 pub trait AnyhowHubExt {
     /// Captures an [`anyhow::Error`] on a specific hub.
+    ///
+    /// [`anyhow::Error`]: https://docs.rs/anyhow/*/anyhow/struct.Error.html
     fn capture_anyhow(&self, e: &anyhow::Error) -> Uuid;
 }
 
--- a/sentry-anyhow/src/lib.rs
+++ b/sentry-anyhow/src/lib.rs
@@ -3,15 +3,17 @@
 //! This integration adds a new event *source*, which allows you to create events directly
 //! from an [`anyhow::Error`] struct.  As it is only an event source it only needs to be
 //! enabled using the `anyhow` cargo feature, it does not need to be enabled in the call to
-//! [`sentry::init`](../../fn.init.html).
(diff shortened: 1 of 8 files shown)


sentry-anyhow/src/lib.rs
```diff
@@ -1,4 +1,17 @@
-//! Adds support for capturing Sentry errors from `anyhow::Error`.
+//! Adds support for capturing Sentry errors from [`anyhow::Error`].
+//!
+//! This integration adds a new event *source*, which allows you to create events directly
+//! from an [`anyhow::Error`] struct.  As it is only an event source it only needs to be
+//! enabled using the `anyhow` cargo feature, it does not need to be enabled in the call to
+//! [`sentry::init`](../../fn.init.html).
```
Swatinem, line 6: I guess these links are made in a way that they work when you are building `sentry` docs, and you expose this via `integrations::anyhow`. However you should still be able to build the docs for this individually. Maybe intra-doc-links will just work when you have `sentry` as a (circular) dev-dependency?

sentry-anyhow/src/lib.rs
```diff
@@ -13,6 +26,8 @@
 //!     capture_anyhow(&err);
 //! }
 //! ```
+//!
+//! [`anyhow::Error`]: https://docs.rs/anyhow/*/anyhow/struct.Error.html
```
Swatinem, line 30: I think intra-doc-links should handle this?

Which reminds me, the relative links from above relative links will also fail when this is converted to the README via cargo-readme :-(

sentry-core/src/hub.rs
```diff
@@ -331,11 +344,11 @@ impl Hub {
 
     /// End the current Release Health Session.
     ///
-    /// See the global [`end_session`](crate::end_session_with)
-    /// for more documentation.
+    /// See the global [`sentry::end_session`](crate::end_session) for more documentation.
```
Swatinem, line 347: 👍 thanks for fixing this ;-)

sentry/src/lib.rs
```diff
@@ -90,8 +90,51 @@ pub use crate::init::{init, ClientInitGuard};
...
+///
+/// Integrations which *process* events in some way usually implement the
+/// [`Itegration`](crate::Integration) trait and need to be installed when sentry is
+/// initialised.
+///
+/// # Installing Integrations
+///
+/// Processing integrations which implement [`Integration`](crate::Integration) need to be
+/// installed when sentry is initialised.  This is done using the
+/// [`ClientOptions::integrations`](crate::ClientOptions::integrations) field, which you can
```
Swatinem, line 110:
```suggestion
/// [`ClientOptions::add_integration`](crate::ClientOptions::add_integration) function, which you can
```

I would rather promote the function than to mess with the raw integrations vec, which should ideally be a private implementation detail, but that would break struct literals.

sentry/src/lib.rs
```diff
@@ -90,8 +90,51 @@ pub use crate::init::{init, ClientInitGuard};
...
+/// ```
+///
+/// # Default Integrations
+///
+/// The [`ClientOptions::default_integrations`](crate::ClientOptions::default_integrations)
+/// option is a boolean field that when enabled will enable a number of default integrations
+/// **before** any integrations provided by
+/// [`ClientOptions::integrations`](crate::ClientOptions::integrations) are installed.  This
+/// is done using the [`apply_defaults`] function, which should be consulted for more
+/// details and the list of which integrations are by default enabled.
```
Swatinem, line 137:
```suggestion
/// details and the list of which integrations are enabled by default.
```

sentry/src/lib.rs
```diff
@@ -90,8 +90,51 @@ pub use crate::init::{init, ClientInitGuard};
...
+/// ```
+/// use sentry::ClientOptions;
+/// use sentry::integrations::debug_images::DebugImagesIntegration;
+///
+/// let options = ClientOptions {
+///     // Add a DSN.
+///     default_integrations: false,
+///     integrations: vec![DebugImagesIntegration::default()],
+///     ..Default::default()
+/// };
```
Swatinem, line 126:
```suggestion
///     default_integrations: false,
///     ..Default::default()
/// }.add_integration(DebugImagesIntegration::new());
```

sentry-anyhow/src/lib.rs
flub: ugh, this was the only solution I could get this to work reasonably.  I didn't think of it being standalone.  I guess that's how it ends up on docs.rs though?

I tried adding the circular dependency under the "doc" feature but cargo was not having any of that.  Do you think as a dev dependency would have more chance of working?

sentry-anyhow/src/lib.rs
flub: I don't understand what happened here.  When trying to use the plain intra-doc links  `cargo +nightly doc` did not give any warning, suggesting it was happy with the link, but they did not work, they just rendered literally.  I have no idea what I was doing wrong and failed to find.  This was the only way I got this to link correctly, it's not very nice but at least it's not wrong.

If you can get it to work with the intra-doc link then let me know what I did wrong!

Swatinem:
```
error[E0432]: unresolved import `sentry::integrations::debug_images`
 --> src/lib.rs:119:27
  |
5 | use sentry::integrations::debug_images::DebugImagesIntegration;
  |                           ^^^^^^^^^^^^ could not find `debug_images` in `integrations`
```

This fails when doing a `no-default-features` build. I think it should be possible to either add a self-reference with `all-features` to the dev-dependencies, or add the `sentry-debug-images` to the dev-dependencies and reference it directly. You decide. But if you want to go with the self-reference, can you also update all examples to use the `sentry::integrations` module?

sentry/src/lib.rs
flub: Oh, I totally missed the function!  This is what we need more of these docs for :D  I'll update the docs with this.

flub:
> ```
> error[E0432]: unresolved import `sentry::integrations::debug_images`
>  --> src/lib.rs:119:27
>   |
> 5 | use sentry::integrations::debug_images::DebugImagesIntegration;
>   |                           ^^^^^^^^^^^^ could not find `debug_images` in `integrations`
> ```
> 
> This fails when doing a `no-default-features` build. I think it should be possible to either add a self-reference with `all-features` to the dev-dependencies, or add the `sentry-debug-images` to the dev-dependencies and reference it directly. You decide. But if you want to go with the self-reference, can you also update all examples to use the `sentry::integrations` module?

Or we could change the feature gates to `#[cfg(any(doc, feature = "debug_images"))]`?  I think this is what they generally seems to recommend?
https://doc.rust-lang.org/rustdoc/advanced-features.html

Swatinem:
> Or we could change the feature gates to `#[cfg(any(doc, feature = "debug_images"))]`?

I don’t think that will work, as you can’t pull in the needed dependency.
Or you can just disable the doctest via:
```rust
/// # #[cfg(feature = "debug-images")] {
/// …
/// # }
```

sentry-anyhow/src/lib.rs
flub: oh, sentry was already a circular dev dependency.  so that doesn't work.  So it's basically between this option which break the doc.rs build for the standalone crate, or point it to `https://docs.rs/sentry/*/sentry/fn.init.html` I think.  Maybe the latter is the least evil :(

Swatinem (approved):

PR #292, opened 2020-11-19
fix: Propagate User updates to the active Session

Swatinem:
In case the session was not yet flushed, we can still set the User.

Question is: Should we flush sessions as soon as we capture anything else or not?
+ on flushing early is that at least we have *something*, - is that it creates more load on the backend, and becomes immutable.

### DIFF
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -178,6 +178,9 @@ impl Scope {
 
     /// Sets the user for the current scope.
     pub fn set_user(&mut self, user: Option<User>) {
+        if let Some(session) = self.session.lock().unwrap().as_mut() {
+            session.update_user(user.as_ref());
+        }
         self.user = user.map(Arc::new);
     }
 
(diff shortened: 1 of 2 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -178,6 +178,9 @@ impl Scope {
 
     /// Sets the user for the current scope.
```
flub, line 179: Can you update this doc comment.

What happens when there is a session?  When will the session not see this update?

sentry-core/src/scope/real.rs
```diff
@@ -178,6 +178,9 @@ impl Scope {
...
     pub fn set_user(&mut self, user: Option<User>) {
+        if let Some(session) = self.session.lock().unwrap().as_mut() {
```
flub, line 181: a method on a scope is supposed to affect the session?

szokeasaurusrex:
Closing as this is likely outdated

PR #293, opened 2020-11-19
fix: Correctly apply environment from env

Swatinem:
I basically broke this in https://github.com/getsentry/sentry-rust/pull/280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - The deprecated `error-chain` and `failure` integrations, features and crates were removed.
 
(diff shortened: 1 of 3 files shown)


flub (approved):
This probably warrants a changelog entry?

jan-auer (approved):

jan-auer:
To avoid this in future, we could add comments to the (empty) defaults mentioning where those values are then actually defaulted. I also didn't recall that we set a default environment.

bruno-garcia:
Just hit this bug. Looking forward to the next release :)

PR #294, opened 2020-11-20
fix: Make Rust 1.48 clippy happy

Swatinem:


flub (approved):

jan-auer (approved):

sentry-core/src/scope/real.rs
```diff
@@ -34,8 +34,8 @@ pub type EventProcessor = Box<dyn Fn(Event<'static>) -> Option<Event<'static>> +
 #[derive(Clone)]
 pub struct Scope {
     pub(crate) level: Option<Level>,
-    pub(crate) fingerprint: Option<Arc<Vec<Cow<'static, str>>>>,
-    pub(crate) transaction: Option<Arc<String>>,
+    pub(crate) fingerprint: Option<Arc<[Cow<'static, str>]>>,
```
flub, line 37: Why is this a `Cow`?

Swatinem: Because it’s defined that way in `Event`, and its a matter if you want to make it a `Cow` ahead of applying the scope to the Event, or whenever you do that, so without changing the `Event`, I think doing it this way is simpler.

PR #295, opened 2020-11-23
ci: Collect Code Coverage

Swatinem:
This uses the experimental nightly-only source-based code coverage based on llvm tooling.
Uses my own https://github.com/Swatinem/fucov github action to orchestrate all the steps. That repo has links to some tracking issues. Especially getting coverage for doctests is still broken in workspaces, which I am tracking.

I think this is good enough to land, and will get automatically better once rust/cargo upstream will get fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,3 +123,28 @@ jobs:
       - run: make checkfast
 
       - run: make testfast
+
+  codecov:
+    name: Code Coverage
+    runs-on: ubuntu-latest
+    continue-on-error: true # well, its nightly and highly experimental
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          components: llvm-tools-preview
+          override: true
+
+      - uses: Swatinem/rust-cache@v1
+
+      - uses: Swatinem/fucov@v1
+        with:
+          args: --workspace --all-features
+
+      - uses: codecov/codecov-action@e156083f13aff6830c92fc5faa23505779fbf649
+        with:
+          directory: coverage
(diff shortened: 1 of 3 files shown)


flub (commented):
where can i see the coverage report?

Swatinem:
https://codecov.io/gh/getsentry/sentry-rust/tree/e958d9004397b2990cf1a28e9a66f72e15a4dafe
Jan suggested to pass the cargo test command invocation to fucov via the action variable, but I can do that in the action without breaking.

.github/workflows/ci.yml
```diff
@@ -123,3 +123,26 @@ jobs:
...
+          profile: minimal
+          toolchain: nightly
+          components: llvm-tools-preview
+          override: true
+
+      - uses: Swatinem/rust-cache@v1
+
+      - uses: Swatinem/fucov@v1
+
+      - uses: codecov/codecov-action@v1
```
jan-auer, line 146: Please pin this to the current SHA of the action until we approve codecov as trusted party.

.github/workflows/ci.yml
```diff
@@ -123,3 +123,26 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          components: llvm-tools-preview
+          override: true
+
+      - uses: Swatinem/rust-cache@v1
+
+      - uses: Swatinem/fucov@v1
```
jan-auer, line 144: Before merging this, can you please add an input to `fucov` to specify the cargo arguments that it runs with?

jan-auer (changes requested):

jan-auer (approved):
To confirm, this now runs the test suite twice: Once with and once without coverage?

Swatinem:
> To confirm, this now runs the test suite twice: Once with and once without coverage?

Yes, though this only runs on nightly linux

PR #296, opened 2020-11-25
Add slog KVs to the Sentry Map

jrobsonchase:
Pretty straightforward - adds a wrapper for `Map` that lets it be used as a `slog::Serializer`, which `KV`s can be appended too.

This also enables the `nested-values` feature on `slog` so that it can accept any serde-serializable values. Considered adding a feature flag, but `sentry-core` already pulls in `serde` & co, so it wouldn't really save any dependencies.


Swatinem (commented):
This is awesome! Can you add a testcase for this?

jrobsonchase:
Alright, added a test that covers strings, numbers, booleans, and nested values.

Swatinem (approved):

PR #297, opened 2020-11-25
Add support for attachments

Jake-Shadle:
This PR adds support for adding Attachments to Envelope so that we can capture and send minidumps to sentry.

It also exposes 2 methods on Client so that users can fill out an event/session update with information on the scope, in our use case, when a crash occurs, as well as a way to push user created Envelopes to the transport, which again, is for our use case of loading and serializing minidumps into envelopes from a previous session.

I'd also be curious if there's any interest in bringing this [crash reporting](https://github.com/EmbarkStudios/sentry-contrib-rust/tree/impl) code into sentry itself, it's basically just a super small wrapper around Breakpad.

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -1,19 +1,22 @@
-use std::any::TypeId;
-use std::borrow::Cow;
-use std::fmt;
-use std::panic::RefUnwindSafe;
-use std::sync::Arc;
-use std::sync::RwLock;
-use std::time::Duration;
+use std::{
+    any::TypeId,
+    borrow::Cow,
+    fmt,
+    panic::RefUnwindSafe,
+    sync::{Arc, RwLock},
+    time::Duration,
+};
 
 use rand::random;
 use sentry_types::protocol::v7::SessionUpdate;
 
-use crate::constants::SDK_INFO;
-use crate::protocol::{ClientSdkInfo, Event};
-use crate::session::SessionFlusher;
-use crate::types::{Dsn, Uuid};
-use crate::{ClientOptions, Envelope, Hub, Integration, Scope, Transport};
+use crate::{
+    constants::SDK_INFO,
+    protocol::{ClientSdkInfo, Event},
+    session::SessionFlusher,
+    types::{Dsn, Uuid},
+    ClientOptions, Envelope, Hub, Integration, Scope, Transport,
+};
 
 impl<T: Into<ClientOptions>> From<T> for Client {
     fn from(o: T) -> Client {
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -9,13 +9,11 @@ use std::time::Duration;
 use rand::random;
 use sentry_types::protocol::v7::SessionUpdate;
 
-use crate::{
-    constants::SDK_INFO,
-    protocol::{ClientSdkInfo, Event},
-    session::SessionFlusher,
-    types::{Dsn, Uuid},
-    ClientOptions, Envelope, Hub, Integration, Scope, Transport,
-};
+use crate::constants::SDK_INFO;
+use crate::protocol::{ClientSdkInfo, Event};
+use crate::session::SessionFlusher;
+use crate::types::{Dsn, Uuid};
+use crate::{ClientOptions, Envelope, Hub, Integration, Scope, Transport};
 
 impl<T: Into<ClientOptions>> From<T> for Client {
     fn from(o: T) -> Client {
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -292,7 +292,7 @@ impl Client {
                 scope.session.lock().unwrap().as_mut().and_then(|session| {
                     // Sorry :p
                     match session.create_envelope_item() {
-                        Some(EnvelopeItem::SessionUpdate(update)) => Some(update),
+                        Some(crate::protocol::EnvelopeItem::SessionUpdate(update)) => Some(update),
                         _ => None,
                     }
                 })
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -151,12 +151,13 @@ impl Client {
         &self,
         mut event: Event<'static>,
         scope: Option<&Scope>,
+        is_sending: bool,
     ) -> Option<Event<'static>> {
         if let Some(scope) = scope {
             scope.update_session_from_event(&event);
         }
 
-        if !self.sample_should_send() {
+        if is_sending && !self.sample_should_send() {
             return None;
         }
 
@@ -172,6 +173,7 @@ impl Client {
         }
 
         if let Some(scope) = scope {
+            // Add eg Breadcrumbs to the event
             event = match scope.apply_to_event(event) {
                 Some(event) => event,
                 None => return None,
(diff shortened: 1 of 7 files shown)


sentry-types/src/protocol/envelope.rs
```diff
@@ -153,11 +155,21 @@ impl Envelope {
                 EnvelopeItem::Transaction(transaction) => {
                     serde_json::to_writer(&mut item_buf, transaction)?
                 }
+                EnvelopeItem::Attachment(attachment) => {
+                    attachment.to_writer(&mut item_buf)?;
```
Swatinem, line 159: maybe write this to `writer` directly instead of basically copying it into `item_buf`? I think down the road we want to stream this stuff out in the transport.

sentry-types/src/protocol/v7.rs
```diff
@@ -1648,3 +1648,82 @@ impl<'a> fmt::Display for Transaction<'a> {
...
+            Self::UnrealLogs => "unreal.logs",
+        }
+    }
+}
+
+#[derive(Clone, PartialEq)]
+/// Represents an attachment item
+pub struct Attachment {
+    /// The actual attachment data
+    pub buffer: std::sync::Arc<Vec<u8>>,
```
Swatinem, line 1693: I wonder if its worth doing the same `Cow` with `'static` bound as we have for the other envelope items.
Also, I will look this up real quick myself, but do we actually clone this somewhere internally that warrants using an `Arc` here?

sentry-types/src/protocol/v7.rs
```diff
@@ -1648,3 +1648,82 @@ impl<'a> fmt::Display for Transaction<'a> {
         )
     }
 }
+
```
Swatinem, line 1651: please move this into its own file, `v7.rs` is already big enough :-D

Swatinem (commented):
Thanks for doing this! I will look at the breakpad integration next

sentry-core/src/client.rs
```diff
@@ -275,6 +282,46 @@ impl Client {
         Default::default()
     }
 
+    /// Assembles an [`Event`](crate::protocol::Event), and possibly a
```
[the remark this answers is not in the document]
Jake-Shadle: Ahh ok, I've been waiting on intradoc links for years but figured I would just be explicit in this case, will do the simpler version and generate them locally to make sure it does what it should.

sentry-types/src/protocol/v7.rs
Jake-Shadle: Heh no problem I did think it was a pretty big boi. :wink:

sentry-types/src/protocol/v7.rs
Jake-Shadle: I think there is Clone derive on something somewhere, but I can try removing it and see what happens.

Jake-Shadle: So the reason I initially did that was because of the Clone requirement on Envelope/EnvelopeItem, but then I noticed that...those are never being used and just removed the Clone requirements altogether.

sentry-core/src/client.rs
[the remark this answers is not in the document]
Jake-Shadle: So I noticed what could be considered a bug in this code, I was doing `Hub::end_session_with_status(SessionStatus::Crashed)` before calling this method, but that actually causes `Hub` to remove the session from the `Scope` so when retrieving it getting null, which doesn't really work well for the crash case, but works as expected if you just...don't call `end_session_with_status`.

sentry-types/src/protocol/envelope.rs
```diff
@@ -73,7 +78,7 @@ impl<'s> Iterator for EnvelopeItemIter<'s> {
 ///
 /// See the [documentation on Envelopes](https://develop.sentry.dev/sdk/envelopes/)
 /// for more details.
-#[derive(Clone, Default, Debug, PartialEq)]
+#[derive(Default, Debug, PartialEq)]
```
Swatinem, line 81: Please keep this `Clone`

sentry-types/src/protocol/attachment.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+#[derive(PartialEq)]
+/// Represents an attachment item
+pub struct Attachment {
+    /// The actual attachment data
+    pub buffer: std::borrow::Cow<'static, [u8]>,
+    /// The filename of the attachment
+    pub filename: std::ffi::OsString,
```
Swatinem, line 46:
```suggestion
    pub filename: String,
```

Since this is about the envelope header which will be sent as JSON, this should be independent of FS encoding.

sentry-types/src/protocol/attachment.rs
```diff
@@ -0,0 +1,80 @@
...
+            Self::UnrealLogs => "unreal.logs",
+        }
+    }
+}
+
+#[derive(PartialEq)]
+/// Represents an attachment item
+pub struct Attachment {
+    /// The actual attachment data
+    pub buffer: std::borrow::Cow<'static, [u8]>,
```
Swatinem, line 44: I’m a bit unsure about usage of `Cow` here. All the other envelope items have `Cow`s inside, and expose that lifetime parameter.

Either way, I will look at the usage of Cow vs String in other places.

sentry-core/src/client.rs
```diff
@@ -275,6 +279,40 @@ impl Client {
         Default::default()
     }
 
+    /// Assembles an [`Event`] and [`SessionUpdate`].
+    ///
+    /// Adds the additional context provided in the specified [`Scope`]
+    /// just as when using [`capture_event`](crate::Client::capture_event), but **without** actually sending
+    /// the event to sentry.
+    pub fn assemble_event(
```
Swatinem, line 287: After checking with other SDK developers, the conclusion was that this proposed fn does not fit well into the unified API specs. Since this is independent from implementing attachment support, please remove this for now along with the changes to `prepare_event`.

sentry-core/src/client.rs
```diff
@@ -275,6 +279,40 @@ impl Client {
...
+            });
+
+            return (Some(event), session_item);
+        }
+
+        (None, None)
+    }
+
+    /// Sends the specified [`Envelope`] to sentry.
+    pub fn send_envelope(&self, envelope: Envelope) {
```
Swatinem, line 310: however, this is something thats useful across SDKs so we can keep this.

sentry-types/src/protocol/attachment.rs
```diff
@@ -0,0 +1,80 @@
+use std::fmt;
+
+/// The different types an attachment can have
```
Swatinem, line 3:
```suggestion
/// The different types an attachment can have.
```

Please make sure that all the doc comments have trailing periods ;-)

sentry-core/src/client.rs
Jake-Shadle: Do you have a suggestion on how I can create an event when a crash occurs so I can serialize it to disk with all of the relevant information on the current `Scope`?

Jake-Shadle: Specifically due to the stated requirement in the Envelopes documentation

> For minidump and apple crash report attachments, the corresponding "event" Item must be sent within the same Envelope.

Swatinem: I would have said to create a custom transport that can dump envelopes to disk, but that might be a bit complex to do.
In native we basically do the same, but we create the envelope in the minidump callback (= in the signal handler).

sentry-types/src/protocol/attachment.rs
Swatinem: I think a `Vec` would be the best idea here. I doubt you will have the case where you have a `'static` attachment buffer. Down the road, the user-facing attachment API will allow you to pass an owned buffer, or lazily load from disk at capture time, as is done in native today.

Jake-Shadle: Cool, done. :slightly_smiling_face:

Swatinem (approved):

PR #300, opened 2020-12-04
Avoid overflow panic

Jake-Shadle:
We encountered a really strange panic (reported in Sentry, yay!), unfortunately the only information was

`overflow when subtracting durations, file: library\core\src\time.rs:895:31, thread: sentry-session-flusher`

And, AFAICT, this is the only place in that thread that dues a Duration subtraction, so putting this as a tentative workaround for it.


Swatinem (approved):
oh, I think this is indeed the culprit here.

PR #301, opened 2020-12-04
fix(log): Create a plain message for exception logs

jan-auer:
Valid exception records in Sentry require a type that reflects the exception name, and frames with function names. Otherwise, issue grouping cannot work properly.


Swatinem (commented):
can you also do this for the `slog` integration then?

PR #302, opened 2020-12-07
ci: Run weekly clippy beta and fix some of the problems

Swatinem:


jan-auer (commented):
Can you align the workflow with the one in Relay and in Symbolicator?

Swatinem:
> Can you align the workflow with the one in Relay and in Symbolicator?

by which you mean: run twice a week, and keep using the matrix?

flub (approved):

flub:
> Can you align the workflow with the one in Relay and in Symbolicator?

I appreciate what you're trying to do, but this is hard to do.  E.g. I experimented with adding the name to the clippy run to see if it would improve reporting, but only did that in symbolicator since... well why do something everywhere when you don't know it'll work.

So ideally there would be some sort of tooling for this, but that's also meta-hard (not a typo!).  In any case, I feel like this isn't worth so much at the moment.  Our CI configs aren't super complicated yet, even if github yaml is somewhat verbose.

jan-auer (approved):
I see your points, makes sense. Action looks good, and I should've approved with my previous comment already.

The general intention to keep the configurations as simple as possible is mostly due to the lack of tooling you mentioned. My hopes are that similar configurations make it easier to apply new learnings and improvements in the workflows.

PR #305, opened 2021-01-07
fix: Remove abandoned `im` crate in favor of Arc::make_mut

### LINKED ISSUE #258: Soundness issues in dependency `sized-chunks` (via `im` crate)
See https://github.com/bodil/sized-chunks/issues/11.

Swatinem:
This incurs a lot more cloning when using lots of scopes and breadcrumbs for example.

fixes #258

### DIFF
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -20,7 +20,7 @@ harness = false
 
 [features]
 default = []
-client = ["im", "rand"]
+client = ["rand"]
 # I would love to just have a `log` feature, but this is used inside a macro,
 # and macros actually expand features (and extern crate) where they are used!
 debug-logs = ["log_"]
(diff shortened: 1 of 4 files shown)


flub (approved):
`Arc::make_mut` is a strange magical api

Swatinem:
NOTE: Please don’t merge yet. There were concerns about the performance, and I would like to measure the impact first. There are some benchmarks in #306

Swatinem:
Okay, I re-ran the benchmarks and get these results:

Tags:
* inactive: 70ns +2%
* active: 21us -18%
* capture: 39us -26%

Breadcrumbs:
* inactive: 2us +1%
* active: 55us +5%
* capture: 92us +6%

So at least with the testcase (relatively few [10-30] tags in the hashmap), we even get a nice speedup.
The breadcrumbs test (overflowing breadcrumbs even), we have ~-5%, so a small hat.

Interesting that the hashmap testcase is sped up, but it kind of makes sense, since immutable hashmaps use buckets (I think 32 elements maybe?) so with a small number of tags they don’t win us anything. I think the number of k/v pairs should be fairly small usually, but I can also create another test with a higher number.

For the breadcrumbs, the testcase runs into the default `max_breadcrumbs` limit and the slowdown is acceptable IMO. I guess the slowdown will get worse if people explicitly increase the number of `max_breadcrumbs`.

PR #306, opened 2021-01-07
meta: Add a small benchmark suite to sentry-core

Swatinem:
The benchmarks so far test the scope with tags and breadcrumbs in a few
situations:
* inactive client,
* active client, only scope manipulation
* active client, scope and capturing messages

The Benchmarks also allow measuring the "minimal API" mode, which
basically removes the complete implementation at compile time.

Measurements on my system are:

Minimal API (compile-time noop): ~2.6ns

Tags:
* inactive: 66ns
* active: 26us
* capture: 53us

Breadcrumbs:
* inactive: 2us
* active: 52us
* capture: 88us

### DIFF
--- a/sentry-core/benches/scope_benchmark.rs
+++ b/sentry-core/benches/scope_benchmark.rs
@@ -1,0 +1,159 @@
+//! Sentry Scope Benchmarks
+//!
+//! Run the benchmarks with:
+//!
+//! ```text
+//! $ cargo bench -p sentry-core
+//! ```
+//!
+//! We have the following tests:
+//! * [`scope_with_breadcrumbs`]
+//! * [`scope_with_tags`]
+//!
+//! We do test the following permutations:
+//! * No active Hub is bound, meaning most API functions will just noop
+//! * Doing scope manipulation with an active Hub, but *not* capturing any messages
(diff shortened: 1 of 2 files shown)


sentry-core/benches/scope_benchmark.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    impl sentry::Transport for NoopTransport {
+        fn send_envelope(&self, envelope: sentry::Envelope) {
+            drop(envelope)
+        }
+    }
+
+    let client = Arc::new(sentry::Client::from(sentry::ClientOptions {
+        dsn: Some("https://public@sentry.invalid/1".parse().unwrap()),
+        // lol, this double arcing -_-
```
flub, line 113: lol, how did that happen?

flub (approved):

sentry-core/benches/scope_benchmark.rs
Swatinem: The first is `impl TransportFactory for Arc<Transport>`, the second one is because clientoptions takes an `Arc<dyn TransportFactory>` explicitly :-(

PR #314, opened 2021-01-29
ref: Rewrite Transports to be able to explicitly flush

Swatinem:
This implements a new `flush` method for Transports which can flush the current queue without closing the transport.

Also, the default transports were rewritten in a more generic way using an async-capable background thread.

Additionally, the internal transports now support more fine grained rate limiting via `X-Sentry-Rate-Limits`. The `RateLimiter` is currently private, but we can decide how to best expose it later on.

TODO
* [ ] Write some integration tests based on a local server which can test proper http requests, rate limiting, etc…
* [ ] Figure out borrow errors related to passing/returning `RateLimiter` by value instead of `&mut`.


sentry-core/src/transport.rs
```diff
@@ -14,15 +14,19 @@ pub trait Transport: Send + Sync + 'static {
...
-    /// Drains the queue if there is one.
+    /// Flushes the transport queue if there is one.
     ///
-    /// The default implementation does nothing.  If the queue was successfully
-    /// shutdowned the return value should be `true` or `false` if events were
-    /// left in it.
-    fn shutdown(&self, timeout: Duration) -> bool {
+    /// If the queue was successfully drained, the return value should be
+    /// `true` or `false` if events were left in it.
+    fn flush(&self, timeout: Duration) -> bool {
```
flub, line 21: As we're re-writing this I can't help but wonder if Duration is really that great here.  Maybe something that can be dropped instead with a wrapper that can drop on a Duration.  Admittedly that's probably a bit heavy weight and overthinking it.  I have no usecase :)

sentry/src/transports/ratelimit.rs
```diff
@@ -0,0 +1,141 @@
...
+impl RateLimiter {
+    /// Create a new RateLimiter.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Updates the RateLimiter with information from a `Retry-After` header.
+    pub fn update_from_retry_after(&mut self, header: &str) {
+        let new_time = if let Ok(value) = header.parse::<f64>() {
+            Some(SystemTime::now() + Duration::from_secs(value.ceil() as u64))
```
flub, line 24: Why is rate limiting done with SystemTime rather than Instant?

Swatinem: Not quite sure actually. I just copied the old code, without thinking too much about it.

sentry-core/src/transport.rs
Swatinem: I don’t quite understand what you mean?

flub: Never mind, I'm not sure this random musing is very useful.  I'm struggling to come up with a reasonable use-case for something else that a Duration.

flub (approved):

PR #316, opened 2021-02-12
build(actix): Update actix-web to v4.

fourbytes:
Updated sentry-actix to support actix-web v4 (currently still in beta).

### DIFF
--- a/sentry-actix/Cargo.toml
+++ b/sentry-actix/Cargo.toml
@@ -12,8 +12,8 @@ Sentry client extension for actix-web 3.
 edition = "2018"
 
 [dependencies]
-sentry-core = { version = "0.23.0", path = "../sentry-core", default-features = false }
-actix-web = { version = "3", default-features = false }
+sentry-core = { version = "0.23.0", path = "../sentry-core", default-features = false, features = ["client"] }
+actix-web = { version = "4.0.0-beta", default-features = false }
 futures-util = { version = "0.3.5", default-features = false }
 
 [dev-dependencies]
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -150,12 +150,11 @@ impl Default for Sentry {
     }
 }
 
-impl<S, B> Transform<S> for Sentry
+impl<S, B> Transform<S, ServiceRequest> for Sentry
 where
-    S: Service<Request = ServiceRequest, Response = ServiceResponse<B>, Error = Error>,
+    S: Service<ServiceRequest, Response = ServiceResponse<B>, Error = Error>,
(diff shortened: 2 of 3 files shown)


aramperes (commented):
Note: this raises the minimum supported Rust version from 1.45 to 1.46: https://github.com/actix/actix-web/pull/1858

The `Test MSRV` job will need to be updated:

https://github.com/getsentry/sentry-rust/blob/e8d0e1b5723dab23f74c2e9cd734859ae36f546f/.github/workflows/ci.yml#L102

Swatinem (commented):
lgtm! lints are still failing, and I will wait with merging this until we have a stable release.

sentry-actix/Cargo.toml
```diff
@@ -13,7 +13,7 @@ edition = "2018"
 
 [dependencies]
 sentry-core = { version = "0.22.0", path = "../sentry-core", default-features = false }
```
JakubKoralewski, line 15: I need to add this for my project using actix-web v4 and this sentry-actix branch to compile for some reason:

```suggested
sentry-core = { version = "0.22.0", path = "../sentry-core", default-features = false, features=["client"] }
```

mlodato517: I also had to do this in my local fork

mlodato517 (commented):
I think this needs to update the tests to remove the `&mut`s in [`call_service`](https://github.com/getsentry/sentry-rust/blob/d798b3c594bc05a4ac66beb724ed702dcc6ca7a8/sentry-actix/src/lib.rs#L360) which now takes a [shared reference](https://docs.rs/actix-web/4.0.0-beta.8/actix_web/test/fn.call_service.html).

sentry-actix/Cargo.toml
fourbytes: Hmm, not sure about this as it seems to be working in my project and wasn't needed in previous versions.

fourbytes:
Latest update should fix those lints, but happy to wait for the full release of v4 for this to be merged.

sentry-actix/Cargo.toml
fourbytes: Ah, just managed to reproduce this. Honestly not sure why it wasn't happening before. It does appear it is now necessary though so I've updated the PR.

ppamorim:
@fourbytes Any update on this, I can't use it because I am getting this error:
```
the trait `Transform<actix_web::app_service::AppRouting, ServiceRequest>` is not implemented for `Sentry`
```
This error is similar to the one I found here: https://github.com/actix/actix-web/issues/2633

Swatinem:
superceded by #437

PR #323, opened 2021-04-08
feat(scope): Add Scope::set_context_other() method

flub:
This introduces a new method which allows to more easily set custom
values in a Context.  It also introduces a missing API to allow
manipulating such contexts.

### DIFF
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -1,5 +1,5 @@
 use std::borrow::Cow;
-use std::collections::{HashMap, VecDeque};
+use std::collections::{BTreeMap, HashMap, VecDeque};
 use std::fmt;
 use std::sync::{Arc, Mutex, PoisonError, RwLock};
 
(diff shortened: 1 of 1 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -204,7 +204,54 @@ impl Scope {
...
     }
 
+    /// Returns a mutable reference to a context key if it exists.
+    ///
+    /// This allows in-place modification of a context value, especially useful when dealing
+    /// with [`Context::Other`] and need to do something more complicated then the
+    /// [`Scope::set_context_other`] shortcut allows.
+    ///
+    /// If the `key` does not exist `None` is returned.
+    pub fn context_mut(&mut self, key: &str) -> Option<&mut BTreeMap<String, Value>> {
```
Swatinem, line 214: We don’t really want to expose mutable refs into the scope itself, as that would in this specific case expose the inner `BTreeMap`, which we just recently changed from (it used to be a map from the unmaintained `im` crate).

Also, provided you could always return `None`, the `noop.rs` build should actually noop all these methods.

flub: err, the `BTreeMap` is entirely part of the public API currently: https://docs.rs/sentry/0.22.0/sentry/protocol/enum.Context.html#variant.Other

Swatinem: Oh, my bad, I was confusing which map you were actually returning. Though still not convinced we want to have this API 🤔

flub: I'm not convinced of the naming, but currently you don't have enough api to add multiple entries to the same context key in different `configure_scope()` calls.  I think @jan-auer also complain about something unified API with this addition.  But I still think you can't do something very reasonable with the current API.

sentry-core/src/scope/real.rs
```diff
@@ -204,7 +204,54 @@ impl Scope {
...
+    ///
+    /// The context has a number of well-defined sections but also allows creating custom
+    /// sections with arbitrary key-value mappings using the [`Context::Other`] map.  This
+    /// is useful to create application-specific sections.
+    ///
+    /// This method is a shortcut which allows you to directly set such a key-value pair as
+    /// this is much more convenient than using the [`Scope::set_context`] method.  If the
+    /// section does not yet exist it is created, if the key already exists in this section
+    /// it is overwritten.
+    pub fn set_context_other(
```
jan-auer, line 231: This method does not adhere to the unified specification and looks like it is just another form of `set_extra`, except that it writes into a non-standard `"other"` context, that isn't useful. The entire point of the context API is to encourage semantic names for contexts.

Let's bring this up in the next SDK evolution comittee, I'll add you to this meeting @flub.

flub: now i'm being told at the same time to prefer `Context::Other` over `set_extra` as well as the other way round 😕   Looking forward to the meeting 👍

PR #324, opened 2021-04-09
fix(deps): correct minium chrono version and MSRV

flub:
The sentry-core crate uses the chrono::DurationRound trait which was
only introduced in chrono 0.4.13.  The sentry-types crate is updated
just to keep the versions in sync.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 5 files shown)


flub:
Do not be confused by the unfinished CI checks, everything works if we're ok with bumping the MSRV to 1.46 (I couldn't find this in any documentation so didn't update it).  The branch protection needs to be updated if we approve this.

So this is as ready for review as it gets.

Swatinem (approved):

Swatinem:
#316 would require a MSRV bump as well since actix bumped it as well.

PR #326, opened 2021-04-26
fix: Explicitly drop SessionFlusher

### LINKED ISSUE #322: Test execution hangs if guards are used in multiple tests
### Environment

How do you use Sentry?
sentry.io

Which SDK and version?
sentry-rust (0.22.0)

Toolchain:
stable-x86_64-pc-windows-msvc (default)
rustc 1.50.0 (cb75ad5db 2021-02-10)

### Steps to Reproduce
1. Define two tests that creates guards: 
```rust
#[test]
fn test_a() {
    let _guard = sentry::init(sentry::ClientOptions {
        ..Default::default()
    });
}

#[test]
fn test_b() {
    let _guard = sentry::init(sentry::ClientOptions {
        ..Default::default()
    });
}
```
2. Run `cargo test`
3. Get output:
```text
running 2 tests
test tests::test_a ... ok
test tests::test_b ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s 
```
4. Test execution freezes and doesn't return
5. Here is an example application to reproduce the problem: https://github.com/rasviitanen/sentry-bug

Running the tests individually works fine

### Expected Result

The test execution should finish with exit 0.

### Actual Result

```
running 2 tests
test observability::tests::test_a ... ok
test observability::tests::test_b ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 63 filtered out; finished in 0.03s 
<here it just freezes and you have to press ctrl+c to exit>
```

We didn't experience any issues with this in `0.19.0`

Swatinem:
It seems that doing a thread::join from within the Drop impl of a thread
local THREAD_HUB will deadlock for some reason.
Explicitly joining the background flusher as part of the ClientGuard
Drop does work, and is also more correct, as otherwise the PROCESS_HUB
might keep the worker alive forever.

Fixes #322

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -43,7 +43,7 @@ pub(crate) type TransportArc = Arc<RwLock<Option<Arc<dyn Transport>>>>;
 pub struct Client {
     options: ClientOptions,
     transport: TransportArc,
-    session_flusher: SessionFlusher,
+    session_flusher: RwLock<Option<SessionFlusher>>,
     integrations: Vec<(TypeId, Arc<dyn Integration>)>,
     sdk_info: ClientSdkInfo,
 }
@@ -60,7 +60,10 @@ impl fmt::Debug for Client {
 impl Clone for Client {
     fn clone(&self) -> Client {
         let transport = Arc::new(RwLock::new(self.transport.read().unwrap().clone()));
-        let session_flusher = SessionFlusher::new(transport.clone(), self.options.session_mode);
+        let session_flusher = RwLock::new(Some(SessionFlusher::new(
+            transport.clone(),
+            self.options.session_mode,
+        )));
         Client {
             options: self.options.clone(),
             transport,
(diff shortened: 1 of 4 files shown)


sentry-core/src/client.rs
```diff
@@ -43,7 +43,7 @@ pub(crate) type TransportArc = Arc<RwLock<Option<Arc<dyn Transport>>>>;
 pub struct Client {
     options: ClientOptions,
     transport: TransportArc,
-    session_flusher: SessionFlusher,
+    session_flusher: RwLock<Option<SessionFlusher>>,
```
flub, line 46: Would it be possible to somehow leave an implementation note that explains why this is an option?  If I read this correctly it only is an option because it allows you to explicitly drop the SessionFlusher on close.  So this rather different from the thing being actually optional as you'd expect from an option.

flub (approved):

sentry-core/src/client.rs
Swatinem: agreed. Although maybe it would be better to just make `client` a newtype around `Option<ClientInner>` or something.

flub: that sounds even nicer, also more work though ;)

PR #329, opened 2021-05-07
feat(tracing): add tracing-subscriber integration

Tuetuopay:
This adds a new integration to `tracing-subscriber`'s registry by providing a `Layer` capable of capturing `tracing` events (and `log` logs) as Sentry Exceptions/Events/Breadcrumbs.

Also, this restores the functionality provided by `sentry-log` when using `tracing-subscriber`, as it installs its own logger to capture `log`'s logs as `tracing` events.

The API is heavily inspired by `sentry-log`, and provides the same methods for matching log/event levels to a sentry type (exceptions/events/breadcrumbs).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,5 +10,6 @@ members = [
     "sentry-log",
     "sentry-panic",
     "sentry-slog",
+    "sentry-tracing",
     "sentry-types",
 ]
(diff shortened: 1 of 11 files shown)


Tuetuopay:
@Swatinem I just force-pushed with the following changes:
- run `cargo fmt`
- add asserts on event messages

Swatinem (approved):

Issue #343, opened 2021-06-16
Sentry client doesn't send events in v0.22.0


onelson:
I see there are already several tickets open regarding "sentry doesn't send events" but they cite different versions of the sdk and seem to be related to different causes. This may very well be a duplicate -- I'm unsure.

I've got an actix-web (3.x.x) app and I'm trying to follow along with the examples in the docs to add sentry support.

### Environment

In my `Cargo.toml`

```toml
[dependencies.sentry]
version = "0.22.0"
features = [
    "anyhow",
    "log",
    "debug-logs",
]

[dependencies.sentry-actix]
version = "0.22.0"
```

In my `main.rs`:

```rust
#[actix_web::main]
async fn main() -> Result<()> {
    #[cfg(feature = "dotenv")]
    dotenv::dotenv().ok();

    env_logger::init();

    let _guard = sentry::init((
        dbg!(std::env::var("SENTRY_DSN").unwrap_or_default()),
        sentry::ClientOptions {
            release: sentry::release_name!(),
            ..Default::default()
        },
    ));
    std::env::set_var("RUST_BACKTRACE", "1");

    log::error!("smoke test");

// ...

    log::info!("Server starting on `{}`", bind_addr);

    Ok(HttpServer::new(move || {
        App::new()
            .wrap(middleware::Logger::default())
            .wrap(sentry_actix::Sentry::new())
            .configure(handlers::configure_routes)
    })
    .bind(bind_addr)?
    .run()
    .await?)
}
```

### Steps to Reproduce

Starting the service, and making a request to a handler like:

```rust
#[get("/whoops")]
pub async fn whoops(_req: HttpRequest) -> crate::Result<&'static str> {
    Err(std::io::Error::new(std::io::ErrorKind::Other, "smoke test").into())
}
```

### Expected Result

2 new issues arriving in my sentry project

- the `log::error!("smoke test");` in main, and 
- the Err returned by the handler.

### Actual Result

Neither issue made it through to sentry.

```
web    | [2021-06-16T18:01:27Z DEBUG sentry::transport] spawning reqwest transport
web    | [2021-06-16T18:01:27Z DEBUG sentry::init] enabled sentry client for DSN https://xxxx.sentry.io/api/0000
web    | [2021-06-16T18:01:27Z ERROR chibi_cerebro] smoke test
web    | [2021-06-16T18:01:27Z INFO  chibi_cerebro] Server starting on `0.0.0.0:7878`
[...]

PR #345, opened 2021-06-19
docs: Provide more upfront info on features and integrations, nitpicky fixes

relaxolotl:
Adds some extra info in crate docs that should smooth out the initial setup process without bloating the quickstart section. This includes some information about integrations that do not work out of the box and require additional setup to function, as suggested in #343. 

I've taken the liberty to try to lay everything out in a table in hopes that this makes things easier to parse at a glance. The table notes which features are enabled by default, as well as ones that are deprecated (?) which will most likely need to be double-checked. Most of these changes reside in `/sentry/README.md`, and everything else is mostly composed of nitpicky fixes that I noticed while I double-checked to make sure all of the docs were consistent.

### DIFF
--- a/sentry/README.md
+++ b/sentry/README.md
@@ -7,18 +7,20 @@
 # Sentry Rust SDK: sentry
 
 This crate provides support for logging events and errors / panics to the
-[Sentry](https://sentry.io/) error logging service.  It integrates with the standard panic
+[Sentry] error logging service. It integrates with the standard panic
 system in Rust as well as a few popular error handling setups.
+
+[Sentry]: https://sentry.io/
 
 ## Quickstart
 
-The most convenient way to use this library is the [`sentry::init`] function,
+The most convenient way to use this library is via the [`sentry::init`] function,
 which starts a sentry client with a default set of integrations, and binds
 it to the current [`Hub`].
 
 The [`sentry::init`] function returns a guard that when dropped will flush Events that were not
-yet sent to the sentry service.  It has a two second deadline for this so shutdown of
-applications might slightly delay as a result of this.  Keep the guard around or sending events
+yet sent to the sentry service. It has a two second deadline for this so shutdown of
+applications might slightly delay as a result of this. Keep the guard around or sending events
 will not work.
 
 ```rust
@@ -33,14 +35,15 @@ sentry::capture_message("Hello World!", sentry::Level::Info);
 
 ## Integrations
 
-What makes this crate useful are the various integrations that exist.  Some of them are enabled
-by default, some uncommon ones or for deprecated parts of the ecosystem a feature flag needs to
-be enabled.  For the available integrations and how to use them see
-[integrations](https://docs.rs/sentry/0.21.0/sentry/integrations/index.html) and [apply_defaults](https://docs.rs/sentry/0.21.0/sentry/fn.apply_defaults.html).
+What makes this crate useful are its various integrations. Some of them are enabled by 
+default; See [Features]. Uncommon integrations or integrations for deprecated parts of 
+the ecosystem require a feature flag. For available integrations and how to use them, see
+[integrations](https://docs.rs/sentry/0.21.0/sentry/integrations/index.html) and
+[apply_defaults](https://docs.rs/sentry/0.21.0/sentry/fn.apply_defaults.html). 
 
 ## Minimal API
 
-This crate comes fully featured. If the goal is to instrument libraries for usage
+This crate comes fully-featured. If the goal is to instrument libraries for usage
 with sentry, or to extend sentry with a custom [`Integration`] or a [`Transport`],
 one should use the [`sentry-core`] crate instead.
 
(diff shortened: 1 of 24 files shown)


sentry/README.md
```diff
@@ -7,18 +7,20 @@
...
+
 ## Quickstart
 
-The most convenient way to use this library is the [`sentry::init`] function,
+The most convenient way to use this library is via the [`sentry::init`] function,
 which starts a sentry client with a default set of integrations, and binds
 it to the current [`Hub`].
 
 The [`sentry::init`] function returns a guard that when dropped will flush Events that were not
-yet sent to the sentry service.  It has a two second deadline for this so shutdown of
```
flub, line 20: 😱 😂

sentry/README.md
```diff
@@ -50,35 +58,57 @@ one should use the [`sentry-core`] crate instead.
...
+| `test` | | | | |
+| `debug-images` | | 🔌 | | |
+| `log` | | 🔌 |  |Requires additional setup; See [`sentry-log`]'s documentation. |
+| `debug-logs` | | | ✂️ | | Requires additional setup; See [`sentry-log`]'s documentation. |
+| `slog` | | 🔌 | | | Requires additional setup; See [`sentry-slog`]'s documentation. |
+| `reqwest` | ✅ | | | |
+| `native-tls` | ✅ | | | `reqwest` must be enabled. | |
+| `rustls` | | | | `reqwest` must be enabled. `native-tls` must be disabled via `default-features = false`. | |
+| `curl` | | | | |
+| `surf` | | | | |
```
flub, line 80: You might want to consider a little more visual alignment in this table to make editing this easier.  I guess it's fine if the lines end up really long with the comments.  And also fully visually aligning is probably not helpful, I'd only try to align the first 4 columns.

sentry/README.md
```diff
@@ -50,35 +58,57 @@ one should use the [`sentry-core`] crate instead.
...
+| `transport` | ✅ | | | |
+| `anyhow` | | 🔌 | | |
+| `test` | | | | |
+| `debug-images` | | 🔌 | | |
+| `log` | | 🔌 |  |Requires additional setup; See [`sentry-log`]'s documentation. |
+| `debug-logs` | | | ✂️ | | Requires additional setup; See [`sentry-log`]'s documentation. |
+| `slog` | | 🔌 | | | Requires additional setup; See [`sentry-slog`]'s documentation. |
+| `reqwest` | ✅ | | | |
+| `native-tls` | ✅ | | | `reqwest` must be enabled. | |
+| `rustls` | | | | `reqwest` must be enabled. `native-tls` must be disabled via `default-features = false`. | |
```
flub, line 78: I think there's an extra trailing column here

flub (approved):
❤️

sentry/README.md
flub: (because seeing this comment anywhere else then in a side-by-side diff makes little sense: yes this is a comment on single-vs-double space sentence ending 😄 )

flub:
I wonder why weekly/periodic CI has not been picking up on that lint... we should debug that sometime.  Maybe compare with symbolicator's periodic CI.

sentry/README.md
```diff
@@ -50,35 +58,57 @@ one should use the [`sentry-core`] crate instead.
...
+| --- | --- | --- | --- | --- |
+| `backtrace` | ✅ | 🔌 | | |
+| `contexts` | ✅ | 🔌 | | |
+| `panic` | ✅ | 🔌 | | |
+| `transport` | ✅ | | | |
+| `anyhow` | | 🔌 | | |
+| `test` | | | | |
+| `debug-images` | | 🔌 | | |
+| `log` | | 🔌 |  |Requires additional setup; See [`sentry-log`]'s documentation. |
+| `debug-logs` | | | ✂️ | | Requires additional setup; See [`sentry-log`]'s documentation. |
```
Swatinem, line 74: This is not really an integration per-se, it is rather a compile-time switch to have sentry output some internal logs to the `log` crate.

sentry/README.md
Swatinem: a markdown formatter will take care of all of this ideally.

sentry/README.md
```diff
@@ -50,35 +58,57 @@ one should use the [`sentry-core`] crate instead.
...
+
+### Debugging/Testing
+- `anyhow`: Enables support for the `anyhow` crate.
+- `test`: Enables testing support.
+- `debug-images`: Attaches a list of loaded libraries to events (currently only supported on Unix).
+
+### Logging
+- `log`: Enables support for the `log` crate.
+- `slog`: Enables support for the `slog` crate.
+- `debug-logs`: **Deprecated**. Uses the `log` crate for internal logging.
```
Swatinem, line 102: not really deprecated

Swatinem (changes requested):
This is awesome, BUT:
All but the top-level readme are actually generated via `cargo-readme` (+ some custom shell script logic) from the crates top-level rustdoc.

So just make sure to copy over all your changes to the rustdocs instead, and please have the markdown table auto-formatted ;-)

sentry/README.md
flub: that would be nice.  but we should run it in CI in that case

sentry/README.md
relaxolotl: 😢 I'm glad you pointed this out actually, this is actually a bit of a dilemma for me: Some places in the docs use double spaces, which some auto-formatter that I need to hunt down has been aggressively converting to single spaces. Should I be using 2 spaces, or 1 space?

sentry/README.md
relaxolotl: just so i have this sorted out: sentry-core's `debug-logs` feature is deprecated according to its docs: https://github.com/getsentry/sentry-rust/blob/master/sentry-core/README.md#features

`debug-logs` on sentry specifies sentry-core's `debug-logs` as a dependency. is `debug-logs` on sentry-core deprecated, but not on sentry itself?

relaxolotl:
Thank you for the feedback, especially the part where the READMEs are actually automatically generated; I'd completely missed that. All of the changes have been migrated over to the rustdocs.

The table has also been run through a markdown formatter, but I reckon it would be preferable if we used an extension to enforce this. I wasn't able to find any extensions that target this specific use case with a quick search, so any guidance or advice on what to do here would be great.

While resolving the lints I'd noticed that some of our links to integrations pointed to nonexistent entries due to feature flags. I've opened up a separate PR to address that linter warning which should also expose previously-hidden bits of the API which were gated behind feature flags: #346

> I wonder why weekly/periodic CI has not been picking up on that lint... we should debug that sometime. Maybe compare with symbolicator's periodic CI.

If I understand this correctly, it looks like the weekly CI has been complaining for a while about this, actually: https://github.com/getsentry/sentry-rust/actions/runs/826285550 is the first time it started throwing a fit and the weekly CI runs have been failing since that run. 1.53.0 was just stabilized last week which included this check, so perhaps this is more of a visibility/notification issue than it is a CI problem?

cc @Swatinem @flub

Swatinem (approved):

sentry/README.md
flub: honestly, i think it doesn't matter and personally I'm fine with mixed.  and if we want consistency we should have an auto-formatter running in CI.  I just found it funny that you changed them, but if it was auto-formatted that explains :)

PR #349, opened 2021-07-01
Update the Transaction struct to match the transaction payload spec

leops:
There are two main changes here to fix errors displayed on the Sentry dashboard when submitting envelopes using the current `Transaction` struct:
- First the `name` field is renamed to `transaction`, this is technically a semver breaking change although given there is not high-level API to work with transactions yet in the SDK this struct probably doesn't have too many consumers yet. If that proves to be a problem it's always possible to just rename the field at the serde level
- Second, the `span_id` fields are now using a custom serializer to truncate the Uuid string at 16 bytes. I went for the simplest change here again, although it means using the Uuid type for fields that are not in fact holding a full UUID in the wire format. Maybe the span IDs should be either a String (like in `parent_span_id`) or a specialized `SpanId` struct of only 16 bytes

### DIFF
--- a/sentry-types/src/protocol/v7.rs
+++ b/sentry-types/src/protocol/v7.rs
@@ -1259,7 +1259,10 @@ pub struct BrowserContext {
 #[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq)]
 pub struct TraceContext {
     /// The ID of the trace event
-    #[serde(default = "event::default_id", serialize_with = "event::serialize_id")]
+    #[serde(
+        default = "event::default_id",
+        serialize_with = "event::serialize_span_id"
+    )]
     pub span_id: Uuid,
     /// Determines which trace the transaction belongs to.
     #[serde(default = "event::default_id", serialize_with = "event::serialize_id")]
@@ -1304,6 +1307,11 @@ mod event {
 
     pub fn serialize_id<S: Serializer>(uuid: &Uuid, serializer: S) -> Result<S::Ok, S::Error> {
         serializer.serialize_some(&uuid.to_simple_ref().to_string())
+    }
+
+    pub fn serialize_span_id<S: Serializer>(uuid: &Uuid, serializer: S) -> Result<S::Ok, S::Error> {
+        let uuid = uuid.to_simple_ref().to_string();
+        serializer.serialize_some(&uuid[..16])
     }
 
     pub fn default_level() -> Level {
@@ -1520,7 +1528,10 @@ impl<'a> fmt::Display for Event<'a> {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct Span {
     /// The ID of the span
-    #[serde(default = "event::default_id", serialize_with = "event::serialize_id")]
+    #[serde(
+        default = "event::default_id",
+        serialize_with = "event::serialize_span_id"
+    )]
     pub span_id: Uuid,
     /// Determines which trace the span belongs to.
     #[serde(default = "event::default_id", serialize_with = "event::serialize_id")]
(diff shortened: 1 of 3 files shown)


Swatinem (approved):
lgtm, I will check back with the performance team for these changes.

sentry-types/src/protocol/v7.rs
```diff
@@ -1603,7 +1614,7 @@ pub struct Transaction<'a> {
     pub event_id: Uuid,
     /// The transaction name.
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub name: Option<String>,
+    pub transaction: Option<String>,
```
rhcarvalho, line 1617: The transaction itself has a `name`, not to be confused with how the transaction gets transformed into an event before sending to Sentry. `Transaction.name` is mapped into `Event.transaction`.

rhcarvalho (commented):
Without diving into the code base, it seems from these changes that there are some gaps in the current implementation.

The change in length of `span_id` is correct, but `span_id` (nor `trace_id`) should NOT be UUIDs.

Renaming `Transaction.name` -> `Transaction.transaction` is incorrect.

sentry-types/src/protocol/v7.rs
```diff
@@ -1306,6 +1309,11 @@ mod event {
         serializer.serialize_some(&uuid.to_simple_ref().to_string())
     }
 
+    pub fn serialize_span_id<S: Serializer>(uuid: &Uuid, serializer: S) -> Result<S::Ok, S::Error> {
+        let uuid = uuid.to_simple_ref().to_string();
+        serializer.serialize_some(&uuid[..16])
```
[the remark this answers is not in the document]
leops: Alright I'll introduce new `SpanId` and `TraceId` types defined as fully random `[u8; 16]` and `[u8; 32]` respectively for these fields (and `parent_span_id` as well I guess ?)

sentry-types/src/protocol/v7.rs
leops: I might have been confused here since the module is called `protocol`, but these definitions don't actually map one-to-one to the payloads that are sent to the store endpoint, is that correct ? If I understand correctly there should be a piece of code (currently unimplemented) that turns a `Transaction` struct into an `Event` instead of it being serialized and submitted directly ?

rhcarvalho: Correct, the in-memory representation is different than the over-the-wire representation.

sentry-types/src/protocol/v7.rs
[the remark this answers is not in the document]
rhcarvalho: If we talk bytes, `u8`, `SpanID` (and `ParentSpanID`) is 8 bytes and `TraceID` is 16 bytes, so we might use smaller arrays and hex-encode them when serializing.

leops:
I pushed a proper fix for the `span_id` and `trace_id` fields to use new `SpanId` and `TraceId` types defined respectively as fully random `[u8; 8]` and `[u8; 16]` that serialize to hex strings. This required introducing 2 new dependencies, one on the `getrandom` library (this one doesn't actually change anything since it's the same dependency used to generate v4 UUID anyway) and one on the `hex` library as I preferred not to get into the business of implementing hex string parsing.

For the `name` field the story is a little more complicated since as I understand it using the high-level `startTransaction` API calling `finish` on a `Transaction` would turn it into an `Event` and send it through `capture_event` on the current `Hub`. However this part of the API is currently unimplemented (I guess due to the difficulty of conciliating the Span tree builder pattern with the rigid ownership rules in Rust ?), and what the SDK currently has instead is a direct translation of the `Transaction` into an `Envelope` object, without going through the intermediate `Event`. Down the line it means the `Transaction` struct is directly serialized into a `transaction` envelope item, and more or less has to masquerade as an `Event` object at least as far as the serialization code is concerned. For now I simply restored the `name` field but renamed it to `transaction` on serialization, but according to the SDK development documentation it might be better to remove the implementation of the `From<Transaction> for Envelope` translation and implement a `From<Transaction> for Event` transform instead ?

Swatinem:
> it might be better to remove the implementation of the `From<Transaction> for Envelope` translation and implement a `From<Transaction> for Event` transform instead ?

That would be a breaking change either way. I think we can go ahead with this as-is, and gather some feedback on how it feels.

Swatinem (approved):

PR #351, opened 2021-07-12
Add SpanStatus to the protocol.

dmitrymur:
This pull request adds a `SpanStatus` enumeration to the protocol according to the documentation [Span Interface](https://develop.sentry.dev/sdk/event-payloads/span/)
It also changes `status` field type from `Option<String>` to `Option<SpanStatus>`  for `Span` and `Transaction`.

### DIFF
--- a/sentry-types/src/protocol/v7.rs
+++ b/sentry-types/src/protocol/v7.rs
@@ -1597,6 +1597,7 @@ impl fmt::Display for Span {
 
 /// The status of a Span.
 #[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[non_exhaustive]
 pub enum SpanStatus {
     /// The operation completed successfully.
     #[serde(rename = "ok")]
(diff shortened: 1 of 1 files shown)


sentry-types/src/protocol/v7.rs
```diff
@@ -1595,6 +1595,113 @@ impl fmt::Display for Span {
     }
 }
 
+/// The status of a Span.
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum SpanStatus {
```
Swatinem, line 1600: please make this `#[non_exhaustive]`

sentry-types/src/protocol/v7.rs
```diff
@@ -1595,6 +1595,113 @@ impl fmt::Display for Span {
...
+    /// Operation was attempted past the valid range.
+    #[serde(rename = "out_of_range")]
+    OutOfRange,
+    /// Unrecoverable data loss or corruption
+    #[serde(rename = "data_loss")]
+    DataLoss,
+}
+
+impl str::FromStr for SpanStatus {
+    type Err = ParseLevelError;
```
Swatinem, line 1655: Maybe a more descriptive error would be a good idea. I have to look at more examples, I’m not sure what kind of errors we use for all the other enums here.

dmitrymur: I am sorry, going to add `ParseStatusError` struct and use it here.

Swatinem (approved):

Issue #180, opened 2020-02-06
`tracing` integration


khionu:
Yet Another Integration Issue. As with #160, support for `tracing` would be fantastic.

nCrazed:
Rocket (a popular web framework) is currently in the process of migrating to tracing, having this integration for when that happens would be very useful.

https://github.com/SergioBenitez/Rocket/pull/1410

kellpossible:
It would also be good to have support for https://github.com/yaahc/eyre too along with this.

khionu:
@kellpossible should open a separate issue for a separate crate to support.

kellpossible:
I've started working on this https://github.com/kellpossible/sentry-tracing I'll make it into a pull request here when I've tested it out more and sure about the feature set. Happy for any comments, or if someone else wants to rip it out and make a PR of it now.

Swatinem:
Some preliminary typedefs related to tracing just landed in #276

dashed:
@kellpossible were you able to generate any useful spans/traces?

kellpossible:
I've got events working as breadcrumbs, and issues. Span support is limited at the moment I think, haven't tested it much.

gakonst:
@kellpossible thank you for the work so far! are you still working on the feature?

kellpossible:
@gakonst not really, it was working well enough for my purposes at the time, and I'm currently reassigned on a different project which doesn't involve working this, but perhaps will come back to it.

leops:
I've forked @kellpossible's repository to https://github.com/levels3d/sentry-tracing and I'm currently working on adding transaction collection. I've implemented the conversion of tracing spans to sentry span data, but I'm not too sure on how to proceed to submit the transaction to Sentry as wrapping it in an enveloppe and sending that doesn't seem to work properly (there might be an issue in the Sentry SDK as the span IDs are getting serialized as full 32 bytes UUID strings when it looks like they should only be 16 characters long)

jplatte:
@leops Idk what transaction collection is, but am I right in assuming that I have to use your fork to get span support? It doesn't seem like the official `sentry-tracing` from this repo currently reports spans to sentry, which makes the reported errors much less useful than they should be.

leops:
[...]

PR #359, opened 2021-09-10
feat: create Exception from tracing event

seanpianka:
This PR relates to my message in #180 

This change will build an exception differently than event. It will look
for two hardcoded keys in the "extras" of an event:
1. "error_kind" - this is optional and defaults to "Unknown"
2. "error" - this is optional and defaults to the event's message

I am open to ideas for improving this! This PR simply gets exception 
handling off the ground, but makes arbitrary assumptions about how the
tracing event will communicate the error information.

### DIFF
--- a/sentry-tracing/Cargo.toml
+++ b/sentry-tracing/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2018"
 
 [dependencies]
 sentry-core = { version = "0.23.0", path = "../sentry-core", features = ["client"] }
+sentry-backtrace = { version = "0.23.0", path = "../sentry-backtrace" }
 tracing-core = "0.1"
 tracing-subscriber = { version = "0.3.1", default-features = false, features = ["std"] }
 
--- a/sentry-tracing/Cargo.toml
+++ b/sentry-tracing/Cargo.toml
@@ -12,7 +12,7 @@ Sentry integration for tracing and tracing-subscriber crates.
 edition = "2018"
 
 [dependencies]
-sentry-core = { version = "0.23.0", path = "../sentry-core" }
+sentry-core = { version = "0.23.0", path = "../sentry-core", features = ["client"] }
 tracing-core = "0.1"
 tracing-subscriber = { version = "0.3.1", default-features = false, features = ["std"] }
 
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -1,13 +1,14 @@
 use std::collections::BTreeMap;
 
-use sentry_core::protocol::{self, Event, TraceContext, Value};
-use sentry_core::{Breadcrumb, Level};
 use tracing_core::{
     field::{Field, Visit},
     span, Subscriber,
 };
 use tracing_subscriber::layer::Context;
 use tracing_subscriber::registry::LookupSpan;
+
+use sentry_core::protocol::{self, Event, Exception, TraceContext, Value};
+use sentry_core::{Breadcrumb, Level};
 
 use crate::Trace;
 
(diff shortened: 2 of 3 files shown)


sentry-tracing/src/converters.rs
```diff
@@ -141,9 +142,66 @@ pub fn exception_from_event<S>(event: &tracing_core::Event, ctx: Context<S>) ->
...
+    let exception = Exception {
+        ty,
+        value,
+        module,
+        ..Default::default()
+    };
+    let culprit = match (event.metadata().file(), event.metadata().line()) {
+        (Some(f), Some(l)) => Some(format!("{}L{}", f, l)),
+        _ => None,
+    };
```
flub, line 174: I've been trying to find out what this field is for and so far the SDK folks seem to be telling me that this is an internal field that's set during processing and also that it is deprecated.  So we shouldn't set it here.
(see also https://github.com/getsentry/relay/pull/1082)

flub (commented):
I like the general approach to this, thanks for starting to work on it.  Some high level questions I have though:

* I would kind of expect to be able to pass an `std::error::Error` type in the `error` field of an trancing span/event and for the sentry integration to extract the type and value and backtrace just like it would do for directly capturing such error types.  Can this be done?
* I think the crate's doc comment should clearly document that the integration treats those tracing fields specially and how to use them to create sentry errors.
* I'm not entirely sure if I'm reading this code correctly, but does this try to create an exception for any tracing event?  Should it only do this for error-level events maybe?

sentry-tracing/src/converters.rs
seanpianka: Removed!

I tried to shoehorn the filename and line number of the span into the exception somehow, but perhaps this information will be included elsewhere? I am unsure about this, since there's no stacktrace available.

seanpianka:
> * I would kind of expect to be able to pass an `std::error::Error` type in the `error` field of an trancing span/event and for the sentry integration to extract the type and value and backtrace just like it would do for directly capturing such error types.  Can this be done?

Yes, this should be possible. The [values](https://docs.rs/tracing/0.1.26/tracing/trait.Value.html) for a field in a tracing span/event are able to record a `dyn Error + 'static`, so retrieving that from the event _should_ be possible. However, when looking over the documentation for [`#[tracing::instrument(err)]`](https://tracing.rs/tracing/attr.instrument.html), it seems to me that it is expecting the error to be `Display` so that it can be printed in the "error" field of any raised `ERROR` events.

> * I think the crate's doc comment should clearly document that the integration treats those tracing fields specially and how to use them to create sentry errors.

Agreed. My choice of field names is arbitrary, and I'd like to align with whichever fields `#[tracing::instrument(err)]` uses for errors.

> * I'm not entirely sure if I'm reading this code correctly, but does this try to create an exception for any tracing event?  Should it only do this for error-level events maybe?

Perhaps I'm misunderstanding the [`on_event`](https://github.com/getsentry/sentry-rust/blob/master/sentry-tracing/src/layer.rs#L278) method in the layer, but I don't think it's creating an exception for any tracing event, only those that are ERROR level or above.

seanpianka:
So, I've re-worked the `record_error` impl using @leops' fork as [an example](https://github.com/levels3d/sentry-tracing/blob/master/src/converters.rs#L135-L142). Now, issues reported in Sentry have their own exceptions with relevant type, title/message, and a (semi-relevant) backtrace.

The "ty" of the exception is still not great. For an error event, the type is `Event path/to/file.rs:LineNumber`. Better than "Unknown", but still not great.

Additionally, it seems like some "extras" are being lost in the issue being sent to Sentry. In my testing with a function annotated with `#[tracing::instrument(err, fields(data = 1))]`, I am unable to find `data = 1` anywhere in the reported Sentry issue... so some more debugging is required.

sentry-tracing/src/converters.rs
```diff
@@ -22,93 +24,99 @@ pub fn convert_tracing_level(level: &tracing_core::Level) -> Level {
...
+
     fn record_bool(&mut self, field: &Field, value: bool) {
         self.record(field, value);
     }
+
     fn record_str(&mut self, field: &Field, value: &str) {
         self.record(field, value);
     }
+
+    fn record_error(&mut self, _field: &Field, value: &(dyn Error + 'static)) {
```
flub, line 87: ooh, very nice that tracing provides this!

sentry-tracing/src/converters.rs
```diff
@@ -22,93 +24,99 @@ pub fn convert_tracing_level(level: &tracing_core::Level) -> Level {
...
+        let event = event_from_error(value);
+        for exception in event.exception {
```
flub, line 89: We should probably provide a better API for this, but this will do for now.

sentry-tracing/src/converters.rs
```diff
@@ -22,93 +24,99 @@ pub fn convert_tracing_level(level: &tracing_core::Level) -> Level {
...
+            self.exceptions.push(exception);
```
flub, line 90: I think you can use `self.exceptions.append(event.exception)` instead of using a loop.  that's implemented a little more efficiently

sentry-tracing/src/converters.rs
```diff
@@ -145,5 +173,37 @@ where
...
+    let (message, mut visitor) = extract_event_data(event);
+
+    let exception = if visitor.exceptions.is_empty() {
+        let ty = event.metadata().name().into();
+        let value = visitor
+            .json_values
+            .remove(DEFAULT_ERROR_VALUE_KEY)
+            .map(|v| v.as_str().map(|s| s.to_owned()))
+            .flatten()
+            .or_else(|| message.clone());
```
flub, line 187: Is this whole thing still needed now that you discovered `record_error`?

flub (commented):
Apologies for the delay in getting back!

sentry-tracing/src/converters.rs
flub: Yes, it would be nice to have this somewhere.  But indeed without a stacktrace that seems tricky.  I'm not very familiar with the event data format, https://develop.sentry.dev/sdk/event-payloads/#other-interfaces might have info on where else could be suitable.

sentry-tracing/src/converters.rs
seanpianka: Fixed, thanks!

sentry-tracing/src/converters.rs
seanpianka: I left this branch as a fallback in the case that the event does not have any exceptions. This would at least create a single exception using what data is available from either the `error` or `message` field (in that order).

sentry-tracing/src/converters.rs
seanpianka: You could create a separate issue for it, if you have a description in mind. "Sentry event exceptions should not be exposed through a public field?"

sentry-tracing/Cargo.toml
```diff
@@ -12,7 +12,8 @@ Sentry integration for tracing and tracing-subscriber crates.
 edition = "2018"
 
 [dependencies]
-sentry-core = { version = "0.23.0", path = "../sentry-core" }
+sentry-core = { version = "0.23.0", path = "../sentry-core", features = ["client"] }
+sentry-backtrace = { version = "0.23.0", path = "../sentry-backtrace" }
```
Swatinem, line 16: why was this change necessary? I don’t see these imports being used? Please revert this in that case.

Swatinem (commented):
lgtm, but it would be nice to have a test for this.

Swatinem:
I think a simple test where you supply some kind of `Error` type, and have that be captured correctly by sentry would be nice.

Swatinem:
@seanpianka 👋🏻 what is still needed to push this over the finish line?

I might be a bit tempted to just merge this as is, as we are currently adding manual performance monitoring support and will then probably overhaul the tracing integration completely.

Swatinem:
I went ahead and merged the code with the latest upstream changes in #412

PR #362, opened 2021-09-20
deps: enable the backtrace feature by default

flub:
It is probably rather counter-intuitive to have this disabled by
default.

### DIFF
--- a/sentry-anyhow/src/lib.rs
+++ b/sentry-anyhow/src/lib.rs
@@ -29,6 +29,16 @@
 //! }
 //! ```
 //!
+//! # Features
+//!
+//! The `backtrace` feature will enable the corresponding feature in anyhow and allow you to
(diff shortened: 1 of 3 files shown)


sentry-anyhow/src/lib.rs
```diff
@@ -29,6 +29,16 @@
...
+//! # Features
+//!
+//! The `backtrace` feature will enable the corresponding feature in anyhow and allow you to
+//! capture backtraces with your events.  It is enabled by default.
+//!
+//! # Resources
+//!
+//! - [Discord](https://discord.gg/ez5KZN7) server for project discussions.
+//! - Follow [@getsentry](https://twitter.com/getsentry) on Twitter for updates.
+//!
```
Swatinem, line 41: This is added by cargo-readme via the template. Not sure this should be part of the rustdoc.

PR #363, opened 2021-09-29
Make  an optional dependency for the `transport` feature

Jasper-Bekkers:
The `tokio` crate is quite a heavy dependency that we deny in our project through `cargo deny`, and it seems to be completely optional for `sentry` as well if we don't compile with the `transport` feature (which we don't - we provide our own transport layer).

This change make `tokio` optional for the non-dev dependencies saving on build times.


Swatinem:
It might be better to couple it to the `reqwest`/`curl`/`surf` features, in line with the mod declaration that pulls this in:

https://github.com/getsentry/sentry-rust/blob/420444d59206e86f5d010f7d1f6b4e1088e44aae/sentry/src/transports/mod.rs#L11-L12

PR #364, opened 2021-09-29
feat: Add support of tracing::Span::record operation for sentry-tracing

zryambus:
Before this PR this code had no effect:

```rust
tracing::Span::current().record("field", &"data");
```

### DIFF
--- a/sentry/tests/test_tracing.rs
+++ b/sentry/tests/test_tracing.rs
@@ -7,10 +7,9 @@ use tracing_subscriber::prelude::*;
 #[test]
 fn test_tracing() {
     // Don't configure the fmt layer to avoid logging to test output
-    tracing_subscriber::registry()
+    let _dispatcher = tracing_subscriber::registry()
         .with(sentry_tracing::layer())
-        .try_init()
-        .unwrap();
+        .set_default();
 
     let events = sentry::test::with_captured_events(|| {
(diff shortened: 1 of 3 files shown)


Swatinem (approved):

Swatinem:
Can you run this through rustfmt real quick? That seems to be the only failure.

zryambus:
> Can you run this through rustfmt real quick? That seems to be the only failure.

Done

sentry/tests/test_tracing.rs
```diff
@@ -65,23 +65,32 @@ fn test_span_record() {
...
+    );
 
     assert_eq!(envelopes.len(), 1);
 
     let envelope_item = envelopes[0].items().next().unwrap();
     let ref transaction = match envelope_item {
         sentry::protocol::EnvelopeItem::Transaction(t) => t,
-        _ => { assert!(false); return; }
+        _ => {
+            assert!(false);
```
Swatinem, line 87: Since this is test code, you can as well `panic!("expected only a transaction item");`
As the panic diverges, you also don’t need the return/block afterwards.

Swatinem (approved):

sentry/tests/test_tracing.rs
zryambus: Thanks. I'll know

PR #366, opened 2021-10-04
feat: Add Windows support to debug-images

Swatinem:


vthib:
Thanks for this feature! I'd love to be able to use it, would it be possible to release a new version of the sentry crates soon?

bobrik:
This also adds Linux support with build-id, which enables debuginfod support in Sentry for Rust projects.

I'd love a release with this as well.

Issue #369, opened 2021-10-14
Sentry-debug-images is not supported on Android and iOS


fzyzcjy:
Hi thanks for the lib! However sentry-debug-images does not support android and ios.

I have made an experiment and it shows nothing. Moreover, looking at code, it uses findshlibs, and the latter has 

```
pub const TARGET_SUPPORTED: bool = cfg!(any(
    target_os = "macos",
    target_os = "linux",
    target_os = "windows"
));
```

which reveals that android&ios are not supported.

Thus, how can we symbolize images when the shipped .so file in android is stripped and symbols are removed? (originally I think I could upload unstripped .so file to sentry and let sentry parse it; but now since no loaded .so file information, sentry cannot symbolize it?)

fzyzcjy:
PR: https://github.com/getsentry/sentry-rust/pull/370

PR #370, opened 2021-10-16
Upgrade `findshlibs` dependency

fzyzcjy:


fzyzcjy:
findshlibs has upgraded. #369 @Swatinem

fzyzcjy:
Done.

Swatinem:
I just went ahead and merged this, no need to wait for a magical CI fix if master is broken because of that anyway.

PR #378, opened 2021-10-27
fix: Remove versions from local dev-dependencies

### LINKED ISSUE #374: .cargo_vcs_info.json not included in packaged crates
When publishing crates via cargo, it will attempt to embed [`.cargo_vcs_info.json`](https://doc.rust-lang.org/cargo/commands/cargo-package.html#description) in to the package so that one can retrieve info about the exact revision a release was made from in the repository the package was built from, however, this file is not present in any of the packages in this repository. I'm guessing this is due to the use of the `--allow-dirty` flag in the release process due to copying the README in to each package based on a cursory glance of the release workflow.

This is related to #373, as my fallback when a license is not part of the packaged crate is to attempt to look up the LICENSE file from the repo, but to ensure it is valid I use the commit that's part of the package to ensure the LICENSE is correct at the exact point in time the release was made, but this method doesn't work on any of the sentry crates.

Swatinem:
This way, cargo will not verify the versions when publishing, removing
the need to use `cargo-hack` for publishing.

fixes #374

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -4,7 +4,6 @@ artifactProvider:
   name: none
 targets:
   - name: crates
-    noDevDeps: true
   - name: github
   - name: registry
     sdks:
--- a/sentry-actix/Cargo.toml
+++ b/sentry-actix/Cargo.toml
@@ -17,6 +17,6 @@ actix-web = { version = "3", default-features = false }
 futures-util = { version = "0.3.5", default-features = false }
 
 [dev-dependencies]
-sentry = { version = "0.23.0", path = "../sentry", features = ["test"] }
+sentry = { path = "../sentry", features = ["test"] }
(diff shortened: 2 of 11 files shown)


relaxolotl (approved):
the changes look good, but it looks like CI isn't happy with this right now 😞 

i don't see anything in the diff that deliberately tries to fix the problem with circular imports that you mentioned in the linked issue - will that be a problem once this is merged in?

Swatinem:
The code coverage CI job is broken right now, its optional anyway.

The actual fix here is that cargo > 1.40 will ignore path-only dev-dependencies when publishing, thus we don’t need `cargo-hack` anymore. (Before, cargo would reject the release, because it checks that dev-dependencies actually exist on crates.io before publishing, which they don’t in case of circular deps)

PR #395, opened 2021-12-22
feat: API for manual performance monitoring (NATIVE-309)

Swatinem:
This implements a manual performance monitoring API, which tries to follow the unified API spec.

One can start a transaction on its own, or based on a distributed tracing header (or another Span), as well as start new child Spans.

Hooking this up to other external services downstream should be possible via an `iter_headers` that allows attaching the headers to outgoing Http requests.

Spans are recorded as part of the transaction they ultimately belong to on `finish`.
Transactions are being sent to sentry properly when they are being `finish`ed.

A Span can be attached to the scope as the *current* span, and queried via `Scope::set_span/get_span`.
This makes it possible to for example query for distributed tracing headers anywhere in the application, or create a child span anywhere.
The *current* span is also used to attach the trace context to any event that is being captured as long as a *current* span exists.

What is currently missing is the ability to add more attrs/metadata to a span after it was created, and the API as a whole would need some polishing down the road, however this implementation is enough to consider this feature "done" and to ship it to customers (and possibly dogfood it ourselves in symbolicator).

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -8,6 +8,12 @@ const MAX_SPANS: usize = 1_000;
 
 // global API:
 
+/// Start a new Performance Monitoring Transaction.
+///
+/// The transaction needs to be explicitly finished via [`Transaction::finish`],
+/// otherwise it will be discarded.
+/// The transaction itself also represents the root span in the span hierarchy.
+/// Child spans can be started with the [`Transaction::start_child`] method.
 pub fn start_transaction(ctx: TransactionContext) -> Transaction {
     let client = Hub::with_active(|hub| hub.client());
     Transaction::new(client, ctx)
@@ -16,6 +22,9 @@ pub fn start_transaction(ctx: TransactionContext) -> Transaction {
 // Hub API:
 
 impl Hub {
+    /// Start a new Performance Monitoring Transaction.
+    ///
+    /// See the global [`start_transaction`] for more documentation.
     pub fn start_transaction(&self, ctx: TransactionContext) -> Transaction {
         Transaction::new(self.client(), ctx)
     }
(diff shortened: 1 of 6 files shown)


sentry-core/src/performance.rs
```diff
@@ -0,0 +1,452 @@
...
+/// The Transaction Context defines the metadata for a Performance Monitoring
+/// Transaction, and also the connection point for distributed tracing.
+#[derive(Debug)]
+pub struct TransactionContext {
+    name: String,
+    op: String,
+    trace_id: protocol::TraceId,
+    parent_span_id: Option<protocol::SpanId>,
+    sampled: Option<bool>,
+}
```
relaxolotl, line 46: slightly nitpicky question; is there any reason why this couldn't be abstracted away into `Transaction` itself? what's the benefit of explicitly exposing `TransactionContext` as opposed to just having just `Transaction` by itself (presumably with all of `TransactionContext`'s methods on `Transaction` instead)? you could also move `start_transaction` so it becomes `Transaction::start()`, and then creating a transaction is a matter of just:

```
Transaction::new("honk", "beep");
# Transaction::set_sampled(...);
Transaction::start();
```

or 
```
Transaction::continue_from_headers(...);
Transaction::start();
```

relaxolotl: it also avoids interesting discrepancies between the two structs, such as `TransactionContext`'s `op` being `String` but `Transaction`'s `op` being `Option<String>`.

relaxolotl (approved):
overall this looks like a solid start 🎉 this is missing a few things on the `Transaction`/`Span`/`TransactionOrSpan` structs such as the ability to change or add a name after their creation, but they're not necessary for basic performance monitoring.

some unit tests ona few methods like `continue_from_headers` or `iter_headers` would be neat as well.

sentry-core/src/performance.rs
Swatinem: Interesting. I don’t really want to go down the route of bikeshedding this.

Swatinem:
> this is missing a few things on the `Transaction`/`Span`/`TransactionOrSpan` structs such as the ability to change or add a name after their creation

lets talk about that later today. This is needed for the followup #400 as its currently not possible to add data/tags to the span/transaction.

PR #396, opened 2022-01-03
feat: Implement category-based rate limiting (NATIVE-412)

Swatinem:
Parsing of headers and rate limits was implemented, but filtered the
envelope was not so far.

### DIFF
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -140,14 +140,14 @@ impl Envelope {
             .next()
     }
 
-    /// Filters the Envelope according to the predicate.
+    /// Filters the Envelope's [`EnvelopeItem`]s based on a predicate,
+    /// and returns a new Envelope containing only the filtered items.
     ///
-    /// Only keeps the [`EnvelopeItem`]s for which the predicate return `true`.
-    /// Additionally, [`Attachment`]s are only kept if the envelope contains
-    /// an [`Event`] or [`Transaction`].
+    /// Retains the [`EnvelopeItem`]s for which the predicate returns `true`.
+    /// Additionally, [`EnvelopeItem::Attachment`]s are only kept if the Envelope
+    /// contains an [`EnvelopeItem::Event`] or [`EnvelopeItem::Transaction`].
     ///
-    /// Returns the filtered envelope, or [`None`] if the envelope would remain
-    /// empty.
+    /// [`None`] is returned if no items remain in the Envelope after filtering.
     pub fn filter<P>(self, mut predicate: P) -> Option<Self>
     where
         P: FnMut(&EnvelopeItem) -> bool,
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -140,6 +140,40 @@ impl Envelope {
             .next()
     }
 
+    /// Filters the Envelope according to the predicate.
+    ///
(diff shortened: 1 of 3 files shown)


sentry-types/src/protocol/envelope.rs
```diff
@@ -140,6 +140,40 @@ impl Envelope {
...
+    ///
+    /// Only keeps the [`EnvelopeItem`]s for which the predicate return `true`.
+    /// Additionally, [`Attachment`]s are only kept if the envelope contains
+    /// an [`Event`] or [`Transaction`].
+    ///
+    /// Returns the filtered envelope, or [`None`] if the envelope would remain
+    /// empty.
+    pub fn filter<P>(self, mut predicate: P) -> Option<Self>
+    where
+        P: FnMut(&EnvelopeItem) -> bool,
```
relaxolotl, line 153: why `FnMut` over `Fn`? it doesn't look like `self` nor the `EnvelopeItem` are being mutated. `filter` itself appears to re-construct the envelope. does the predicate also need `mut`? it looks like this might be leftovers of an attempt at an implementation that mutates and filters the envelope items out of the original envelope.

sentry-types/src/protocol/envelope.rs
```diff
@@ -140,6 +140,40 @@ impl Envelope {
...
     }
 
+    /// Filters the Envelope according to the predicate.
+    ///
+    /// Only keeps the [`EnvelopeItem`]s for which the predicate return `true`.
+    /// Additionally, [`Attachment`]s are only kept if the envelope contains
+    /// an [`Event`] or [`Transaction`].
+    ///
+    /// Returns the filtered envelope, or [`None`] if the envelope would remain
+    /// empty.
```
relaxolotl, line 150: nit:
```suggestion
    /// Filters the Envelope's `EnvelopeItem`s based on a predicate, 
    /// and returns a new Envelope containing only the filtered items. 
    ///
    /// Retains the [`EnvelopeItem`]s for which the predicate returns `true`.
    /// Additionally, [`EnvelopeItem::Attachment`]s are only kept if the Envelope 
    /// contains an [`EnvelopeItem::Event`] or [`EnvelopeItem::Transaction`].
    ///
    /// `None` is returned if no items remain in the Envelope after filtering.
```

relaxolotl (approved):
left a few comments, mostly nitpicky but this otherwise looks good 👍

sentry/src/transports/ratelimit.rs
```diff
@@ -82,11 +83,29 @@ impl RateLimiter {
...
+                _ => RateLimitingCategory::Any,
+            })
+            .is_none()
+        })
+    }
 }
 
 /// The Category of payload that a Rate Limit refers to.
 #[non_exhaustive]
-#[allow(dead_code)]
```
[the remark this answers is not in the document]
Swatinem: We wouldn’t want that. Right now this is completely private, but once we make it public, we would want to have the freedom to add more categories later on.
Also the compiler is smart enough to exhaustively match *inside* the crate that defines the `#[non_exhaustive]` enum.

sentry-types/src/protocol/envelope.rs
Swatinem: This is not about *us* being mutable, rather that it makes the callback as flexible as possible, so *it* can possibly mutate its environment if it so choses.
I think the idea is to prefer `FnOnce` > `FnMut` > `Fn` *if possible*. For example if we save the callback on the struct, and need access to it inside a `&self` method, we can only go as far as `Fn`.

PR #397, opened 2022-01-03
feat: Add a Http Service integration with trace propagation (NATIVE-314) 

Swatinem:
Based on #395, this creates a new sentry-tower feature, which enables a new tower `Layer`, which will start a new transaction for each Http request, using the incoming headers for distributed trace propagation.

### DIFF
--- a/sentry-tower/src/http.rs
+++ b/sentry-tower/src/http.rs
@@ -99,26 +99,27 @@ where
 
     fn call(&mut self, request: Request<Body>) -> Self::Future {
         let transaction = sentry_core::configure_scope(|scope| {
-            // https://develop.sentry.dev/sdk/event-payloads/request/
-            let mut ctx = SentryMap::new();
-
-            ctx.insert(
-                "method".into(),
-                Value::String(format!("{}", request.method())),
-            );
-            ctx.insert("url".into(), Value::String(format!("{}", request.uri())));
-
-            // headers
-            let headers =
-                ValueMap::from_iter(request.headers().into_iter().map(|(header, value)| {
-                    (
-                        header.to_string(),
-                        value.to_str().unwrap_or("<Opaque header value>").into(),
-                    )
-                }));
-            ctx.insert("headers".into(), headers.into());
-
-            scope.set_context("request", sentry_core::protocol::Context::Other(ctx));
+            let sentry_req = sentry_core::protocol::Request {
+                method: Some(request.method().to_string()),
+                url: request.uri().to_string().parse().ok(),
+                headers: request
+                    .headers()
+                    .into_iter()
+                    .map(|(header, value)| {
+                        (
+                            header.to_string(),
+                            value.to_str().unwrap_or_default().into(),
+                        )
+                    })
+                    .collect(),
+                ..Default::default()
+            };
+            scope.add_event_processor(move |mut event| {
+                if event.request.is_none() {
+                    event.request = Some(sentry_req.clone());
+                }
+                Some(event)
+            });
 
             if self.start_transaction {
                 let headers = request.headers().into_iter().flat_map(|(header, value)| {
--- a/sentry-tower/src/http.rs
+++ b/sentry-tower/src/http.rs
@@ -1,0 +1,104 @@
+use std::collections::BTreeMap;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use tower_layer::Layer;
+use tower_service::Service;
+
+#[derive(Clone)]
+pub struct SentryHttpLayer;
+
+#[derive(Clone)]
+pub struct SentryHttpService<S> {
+    service: S,
+}
+
+impl<S> Layer<S> for SentryHttpLayer {
+    type Service = SentryHttpService<S>;
+
+    fn layer(&self, service: S) -> Self::Service {
+        Self::Service { service }
+    }
+}
+
+pub struct SentryHttpFuture<F> {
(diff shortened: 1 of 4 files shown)


sentry-tower/src/http.rs
```diff
@@ -0,0 +1,144 @@
...
+                Some(event)
+            });
+
+            if self.start_transaction {
+                let headers = request.headers().into_iter().flat_map(|(header, value)| {
+                    value.to_str().ok().map(|value| (header.as_str(), value))
+                });
+                let tx_ctx = sentry_core::TransactionContext::continue_from_headers(
+                    // TODO: whats the name here?
+                    "", "http", headers,
```
Swatinem, line 127: op: "http.server"
name: "{METHOD} {url.path}"

sentry-tower/src/http.rs
```diff
@@ -0,0 +1,146 @@
...
+    start_transaction: bool,
+}
+
+impl SentryHttpLayer {
+    /// Creates a new Layer that only logs Request Headers.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Creates a new Layer that also starts a new performance monitoring transaction.
```
flub, line 24: this sounds like it always starts a new transaction rather than that it continues a transaction from any headers found (which is what the struct comment seems to suggest)

sentry-tower/src/http.rs
```diff
@@ -0,0 +1,146 @@
...
+        Self {
+            start_transaction: true,
+        }
+    }
+}
+
+/// Tower Service that logs Http Request Headers.
+///
+/// The Service can also optionally start a new performance monitoring transaction
+/// based on incoming distributed tracing headers.
```
flub, line 35: i find this wording a bit ambiguous (same for the layer): is it a new transaction or continues the transaction set in the tracing headers?

Maybe: "The service can optionally set an active transaction to continue a transaction present in the request's distributed tracing headers"

sentry-tower/src/http.rs
```diff
@@ -0,0 +1,146 @@
...
+#[derive(Clone)]
+pub struct SentryHttpService<S> {
+    service: S,
+    start_transaction: bool,
```
flub, line 39: rather minor, but if you agree with the above re-wording (or a variation) maybe calling this `continue_transaction` or so would be better as well

sentry-tower/src/http.rs
```diff
@@ -0,0 +1,146 @@
...
+                headers: request
+                    .headers()
+                    .into_iter()
+                    .map(|(header, value)| {
+                        (
+                            header.to_string(),
+                            value.to_str().unwrap_or_default().into(),
+                        )
+                    })
+                    .collect(),
```
flub, line 111: this seems expensive.  do we really need to copy all the headers?  can we be selective, i.e. only copy a few headers we need?  can we avoid copying?

sentry-tower/src/http.rs
```diff
@@ -0,0 +1,146 @@
...
+                if event.request.is_none() {
+                    event.request = Some(sentry_req.clone());
+                }
+                Some(event)
+            });
+
+            if self.start_transaction {
+                let headers = request.headers().into_iter().flat_map(|(header, value)| {
+                    value.to_str().ok().map(|value| (header.as_str(), value))
+                });
```
flub, line 124: similar here, i'm fairly certain you don't need to have all headers and can filter on only the sentry-tracing ones

sentry-tower/src/http.rs
Swatinem: maybe the concepts here are a bit confusing. It is a new transaction, but it belongs to an existing trace. :-D

sentry-tower/src/http.rs
Swatinem: thats what `TransactionContext::continue_from_headers` does internally, the API is part of the unified spec:
https://develop.sentry.dev/sdk/performance/#new-span-and-transaction-classes
> TransactionContext should have a static/ctor method called continueFromHeaders(headerMap) which is really just a thin wrapper around fromSentryTrace(headerMap.get("sentry-trace")) right now. This should be preferred by integration/framework-sdk authors over fromSentryTrace as it hides the exact header names used deeper in the core sdk, and leaves opportunity for using additional headers (from the W3C) in the future without changing all integrations.

sentry-tower/src/http.rs
Swatinem: The `protocol::Request` owns all of its data (and so does `protocol::Event` [the `Cow` in it are a lie basically, or rather only `'static` ones are allowed -_-])
So we can’t avoid the copies, and I’m not sure we should be opinionated what kind of headers we log. Also, the `actix` integration does the same already.

sentry-tower/src/http.rs
Swatinem: Well it still does a `start_transaction`, and as commented above, the transaction itself is independent, it just belongs to a trace based on the incoming headers. Also the incoming headers are completely optional so this starts a completely new transaction if the headers are not present.

sentry-tower/src/http.rs
flub: this is true, but we know which headers `continue_from_headers` actually uses, so we can filter them up-front can we not?  though maybe if this stays as an iterator it doesn't have that much of an impact?

flub (approved):

sentry-tower/src/http.rs
flub: the new wording is better 👍

PR #398, opened 2022-01-03
ci: Pin known-good nightly for coverage tests

Swatinem:


Swatinem:
I consider this pretty much blocked on https://github.com/taiki-e/cargo-llvm-cov/pull/122 right now, as any kind of coverage coming from doctests is not being considered. Unfortunately, we rely a lot on doctests for coverage.

flub (approved):

PR #403, opened 2022-01-12
feat: Add ability to set a SpanStatus

Swatinem:
This allows the ability to set a SpanStatus, and does so in the tower http integration.

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -1,8 +1,7 @@
 use std::sync::Arc;
 use std::sync::Mutex;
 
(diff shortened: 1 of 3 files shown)


loewenheim (approved):

sentry-core/src/performance.rs
```diff
@@ -168,6 +167,22 @@ impl TransactionOrSpan {
         }
     }
 
+    /// Set the status of the Transaction/Span.
```
flub, line 170: I'd really like it if there where pointers to the docs on what this status is and what setting it does.

PR #407, opened 2022-01-14
docs: Improve DebugImage docs

Swatinem:
This brings the docs in line with how they are documented in relay,
except leaving out the platform specific details.


kamilogorek (approved):
Although specifying that something is "required" is IMO optional, as those values are not `Option`s anyway

PR #410, opened 2022-01-18
fix: Apply correct rate limiting behavior/precedence

Swatinem:
Rate limiting should prefer the `sentry-rate-limits` header before the
`retry-after` one, and apply a default when no header is present but a
429 response code.


kamilogorek (approved):
We could extract `Duration::from_secs(60)` to a cost, as it's used in 2 places, but it's just a nitpick.

PR #415, opened 2022-01-21
Disable performance when client feature disabled

Jake-Shadle:
The new performance features rely on the `client` feature being enabled, and when it's not enabled sentry-core fails to compile, this just gates the entire `performance` module behind the `client` feature for now so that it at least compiles, as it seemed non-trivial to untangle types like `TransactionOrSpan` from `client`.

### DIFF
--- a/sentry-core/src/lib.rs
+++ b/sentry-core/src/lib.rs
@@ -63,7 +63,6 @@ mod futures;
 mod hub;
 mod integration;
 mod intodsn;
-mod performance;
 mod scope;
 mod transport;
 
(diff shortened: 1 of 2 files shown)


Swatinem:
Oh well, that was a major oversight. This also means our test matrix is incomplete, as this should have been caught ideally.

And I think we have to jump through the hoops and make sure the API at least exists and compiles even without the client, in case someone really wants to instrument their libraries with sentries perf API (though unlikely).

Swatinem:
I opened #416 instead which also correctly tests the feature flag in CI, and keeps the API around if the feature is not specified.

PR #419, opened 2022-01-26
feat: Add `ureq` transport support

MarijnS95:
Ureq is easy to use, smaller than the other http crates and doesn't pull in as many crate dependencies either, which is why [we] are actively switching to it and in need of an `ureq` transport in Sentry too.

[we]: https://github.com/Traverse-Research

### DIFF
--- a/sentry/Cargo.toml
+++ b/sentry/Cargo.toml
@@ -44,6 +44,8 @@ surf-h1 = ["surf_/h1-client", "httpdate"]
 surf = ["surf_/curl-client", "httpdate", "tokio"]
 native-tls = ["reqwest_/default-tls"]
 rustls = ["reqwest_/rustls-tls"]
+ureq = ["ureq_/tls", "httpdate", "tokio"]
+ureq-native-tls = ["ureq_/native-tls", "httpdate", "tokio"]
 
 [dependencies]
 sentry-core = { version = "0.24.2", path = "../sentry-core", features = ["client"] }
@@ -63,6 +65,7 @@ httpdate = { version = "1.0.0", optional = true }
 surf_ = { package = "surf", version = "2.0.0", optional = true, default-features = false }
 serde_json = { version = "1.0.48", optional = true }
 tokio = { version = "1.0", features = ["rt"], optional = true }
+ureq_ = { package = "ureq", version = "2.3.0", optional = true, default-features = false }
 
 [dev-dependencies]
 sentry-anyhow = { path = "../sentry-anyhow" }
--- a/sentry/src/lib.rs
+++ b/sentry/src/lib.rs
@@ -56,24 +56,26 @@
 //! Additional functionality and integrations are enabled via feature flags. Some features require
 //! extra setup to function properly.
 //!
-//! | Feature        | Default | Is Integration | Deprecated | Additional notes                                                                         |
-//! | -------------- | ------- | -------------- | ---------- | ---------------------------------------------------------------------------------------- |
-//! | `backtrace`    | ✅      | 🔌             |            |                                                                                          |
-//! | `contexts`     | ✅      | 🔌             |            |                                                                                          |
-//! | `panic`        | ✅      | 🔌             |            |                                                                                          |
-//! | `transport`    | ✅      |                |            |                                                                                          |
-//! | `anyhow`       |         | 🔌             |            |                                                                                          |
-//! | `test`         |         |                |            |                                                                                          |
-//! | `debug-images` |         | 🔌             |            |                                                                                          |
-//! | `log`          |         | 🔌             |            | Requires extra setup; See [`sentry-log`]'s documentation.                               |
-//! | `debug-logs`   |         |                | ❗         | Requires extra setup; See [`sentry-log`]'s documentation.                               |
-//! | `slog`         |         | 🔌             |            | Requires extra setup; See [`sentry-slog`]'s documentation.                              |
-//! | `reqwest`      | ✅      |                |            |                                                                                          |
-//! | `native-tls`   | ✅      |                |            | `reqwest` must be enabled.                                                               |
-//! | `rustls`       |         |                |            | `reqwest` must be enabled. `native-tls` must be disabled via `default-features = false`. |
-//! | `curl`         |         |                |            |                                                                                          |
-//! | `surf`         |         |                |            |                                                                                          |
-//! | `tower`        |         | 🔌             |            | Requires extra setup; See [`sentry-tower`]'s documentation.                              |
+//! | Feature           | Default | Is Integration | Deprecated | Additional notes                                                                         |
+//! | --------------    | ------- | -------------- | ---------- | ---------------------------------------------------------------------------------------- |
+//! | `backtrace`       | ✅      | 🔌             |            |                                                                                          |
+//! | `contexts`        | ✅      | 🔌             |            |                                                                                          |
+//! | `panic`           | ✅      | 🔌             |            |                                                                                          |
+//! | `transport`       | ✅      |                |            |                                                                                          |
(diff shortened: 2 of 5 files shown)


sentry/Cargo.toml
```diff
@@ -39,11 +39,14 @@ debug-logs = ["log_", "sentry-core/debug-logs"]
...
 reqwest = ["reqwest_", "httpdate", "tokio"]
-curl = ["curl_", "httpdate", "serde_json", "tokio"]
+curl = ["curl_", "httpdate", "tokio"]
 surf-h1 = ["surf_/h1-client", "httpdate"]
 surf = ["surf_/curl-client", "httpdate", "tokio"]
 native-tls = ["reqwest_/default-tls"]
 rustls = ["reqwest_/rustls-tls"]
+ureq = ["ureq_", "httpdate", "tokio"]
+ureq-tls = ["ureq", "ureq_/tls"]
+ureq-native-tls = ["ureq", "ureq_/native-tls"]
```
MarijnS95, line 49: Unfortunately these can't be added to `native-tls`/`rustls` as the `reqwest_/<feat>` dependency automatically turns on `reqwest_` too. Should we rename the `reqwest`-related features?

sentry/src/transports/ureq.rs
```diff
@@ -0,0 +1,105 @@
+use std::time::Duration;
+
+#[cfg(fdoc)]
+use ureq_ as ureq;
+// TODO: Rename to UreqAgent?
+use ureq_::{Agent as UreqClient, AgentBuilder, Proxy};
```
flub, line 6: why rename at all?

sentry/src/transports/ureq.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Creates a new Transport.
+    pub fn new(options: &ClientOptions) -> Self {
+        Self::new_internal(options, None)
+    }
+
+    /// Creates a new Transport that uses the specified [`ureq::Agent`].
+    // TODO: Rename to with_agent
+    pub fn with_client(options: &ClientOptions, client: UreqClient) -> Self {
+        Self::new_internal(options, Some(client))
+    }
```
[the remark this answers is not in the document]
flub: i agree that staying with ureq's naming is probably the better choice

sentry/src/transports/ureq.rs
```diff
@@ -0,0 +1,105 @@
...
+        let url = dsn.envelope_api_url().to_string();
+
+        let thread = TransportThread::new(move |envelope, mut rl| {
+            let mut body = Vec::new();
+            envelope.to_writer(&mut body).unwrap();
+            let request = client
+                .post(&url)
+                .set("X-Sentry-Auth", &auth)
+                // TODO: Can also send from a Reader, but envelope can only dispatch to a Writer
+                .send_bytes(&body);
```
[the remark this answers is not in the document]
flub: maybe, i think generally events are meant to be not too large.  though i wouldn't leave a todo in a transport for this, seems more general

sentry/Cargo.toml
flub: hmm, while technically we don't guarantee api stability i don't think renaming features is that big an improvement that i'd change this now, it'll just create churn for people.

flub (commented):
probably should update the docs wrt to this as well

sentry/src/transports/ureq.rs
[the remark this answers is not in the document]
MarijnS95: Todo wasn't supposed to remain here for the merge, just a mental note to create a comment here. In hindsight implementing a read cursor for an envelope is probably super trivial and error-prone.

sentry/src/transports/ureq.rs
MarijnS95: For the same reason as the rename in other modules and the comment about `Client` vs `Agent`. It's all gone now.

sentry/src/transports/curl.rs
```diff
@@ -63,7 +63,7 @@ impl CurlHttpTransport {
             let mut headers = curl::easy::List::new();
             headers.append(&format!("X-Sentry-Auth: {}", auth)).unwrap();
             headers.append("Expect:").unwrap();
-            headers.append("Content-Type: application/json").unwrap();
+            // headers.append("Content-Type: application/json").unwrap();
```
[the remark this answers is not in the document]
MarijnS95: https://github.com/getsentry/sentry-rust/pull/420

sentry/src/lib.rs
```diff
@@ -103,6 +106,10 @@
 //!   dependencies.
 //! - `curl`: Enables the curl transport.
 //! - `surf`: Enables the surf transport.
+//! - `ureq`: Enables the ureq transport.
+//! - `ureq-tls`: Enables `rustls` support for `ureq`. Please note that this has to be enabled for
+//!    https support, and cannot be enabled by default.
+//! - `ureq-native-tls`: Enables `native-tls` support for `ureq`.
```
flub, line 112: you could recommend here to prefer setting one of `ureq-tls` or `ureq-native-tls` and mention they automatically pull in `ureq`.

Or maybe even go further and rename `ureq` to `ureq_base_` or something and make it internal only.

sentry/Cargo.toml
```diff
@@ -44,6 +44,9 @@ surf-h1 = ["surf_/h1-client", "httpdate"]
 surf = ["surf_/curl-client", "httpdate", "tokio"]
 native-tls = ["reqwest_/default-tls"]
 rustls = ["reqwest_/rustls-tls"]
+ureq = ["ureq_", "httpdate", "tokio"]
```
flub, line 47: do you happen to know why this needs the `tokio` dependency?  I thought ureq was blocking?  I have the same question for the `surf` dependency as well to be fair, but that's clearly out of scope here :)

flub (commented):
i think this looks good, what do you think of the feature names/usage? I'm not objecting to the current layout if you prefer it this way.

sentry/src/lib.rs
MarijnS95: That sounds like a better plan, I don't think a Sentry transport without https makes a lot of sense :sweat_smile:

sentry/Cargo.toml
MarijnS95: Because of the `TransportThread` every backend seems to use. `surf` has a clear `requestr.await`, I think that's async driven by `tokio` too? Perhaps you meant `curl` which - just like `ureq` - only has a single `async move { rl }` and is blocking otherwise.

Getting rid of `tokio` is our next stop unless it's simpler/cleaner to implement this directly. Is a `Transport` - or at least the `send_envelope` function - supposed to return quickly or is it fine if we perform a blocking request here?

Swatinem: It would be fair to split the `TransportThread` into two sync/async variants for transports that are based on one or the other in the future.

Swatinem (approved):

sentry/Cargo.toml
flub: oh, the `TrnasportThread` builds an internal Tokio runtime...  that is indeed a bit needless for sync transports and weird for the surf one.  would be nice to fix this sometime indeed.

MarijnS95: I'll pick this up in a followup PR!

sentry/src/lib.rs
MarijnS95: @flub I opted to go for `ureq` and `ureq-native-tls`. No `ureq_base_` though, as there's only `httpdate` support here after we get rid of `tokio`.

flub (approved):

sentry/Cargo.toml
MarijnS95: https://github.com/getsentry/sentry-rust/pull/422

PR #422, opened 2022-01-28
refactor: Remove tokio dependency for blocking ureq and curl transports

MarijnS95:
These transport backends don't need the overhead of a tokio runtime and all the dependency crates that come with it, when they are already blocking on their own.  A simple Rust thread and mpsc channel is enough.  The transport thread implementation is duplicated between a tokio and non-tokio module that are used by reqwest/surf, and curl/ureq respectively.

See also https://github.com/getsentry/sentry-rust/pull/419#discussion_r794287640 where this was briefly discussed.

### TODO
- Refactor this a bit to not duplicate the entire thing.  Perhaps leave both implementations in the same file with some cfgs so that common changes are more easily shared.  In the end the only difference is in the `fn new()` signature and the contents of the thread closure...
- Figure out what to do with `surf`, which [has been mentioned](https://github.com/getsentry/sentry-rust/pull/419#discussion_r794287640) multiple times in not needing `tokio`?

### DIFF
--- a/sentry/Cargo.toml
+++ b/sentry/Cargo.toml
@@ -39,13 +39,13 @@ debug-logs = ["log_", "sentry-core/debug-logs"]
 # transports
 transport = ["reqwest", "native-tls"]
 reqwest = ["reqwest_", "httpdate", "tokio"]
-curl = ["curl_", "httpdate", "tokio"]
+curl = ["curl_", "httpdate"]
 surf-h1 = ["surf_/h1-client", "httpdate"]
 surf = ["surf_/curl-client", "httpdate", "tokio"]
 native-tls = ["reqwest_/default-tls"]
 rustls = ["reqwest_/rustls-tls"]
-ureq = ["ureq_/tls", "httpdate", "tokio"]
-ureq-native-tls = ["ureq_/native-tls", "httpdate", "tokio"]
+ureq = ["ureq_/tls", "httpdate"]
+ureq-native-tls = ["ureq_/native-tls", "httpdate"]
 
 [dependencies]
 sentry-core = { version = "0.24.2", path = "../sentry-core", features = ["client"] }
--- a/sentry/src/transports/curl.rs
+++ b/sentry/src/transports/curl.rs
@@ -39,7 +39,7 @@ impl CurlHttpTransport {
         let scheme = dsn.scheme();
 
         let mut handle = client;
-        let thread = TransportThread::new(move |envelope, mut rl| {
+        let thread = TransportThread::new(move |envelope, rl| {
             handle.reset();
             handle.url(&url).unwrap();
             handle.custom_request("POST").unwrap();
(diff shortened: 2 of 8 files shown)


Swatinem (approved):
IMO, the level of code duplication is acceptable, considering that the generics do not make it really possible to do conditional compilation.

flub:
> * Figure out what to do with `surf`, which [has been mentioned](https://github.com/getsentry/sentry-rust/pull/419#discussion_r794287640) multiple times in not needing `tokio`?

Don't worry too much about surf, certainly not in this PR.  It's fine to leave this just about async/sync transports.  The issue with surf is that is supports both async-std as well as tokio, so really the async version of the thread should have an implementations for both these runtimes.

flub (approved):

flub:
thanks for this!  I agree with @Swatinem that this is fine as is as well without further refactoring.  so if you are fine with that as well we can merge as-is.

MarijnS95:
@flub yes please, all set for merging! It's probably already considered, but I'd love a new release with these changes :grimacing:

MarijnS95:
An unrelated comment to this PR: When cloning all branches from this repo are adopted with it, though most appear to be stale remnants from closed PRs: https://github.com/getsentry/sentry-rust/branches/stale - would you mind deleting them for cleanliness?

flub:
> An unrelated comment to this PR: When cloning all branches from this repo are adopted with it, though most appear to be stale remnants from closed PRs: https://github.com/getsentry/sentry-rust/branches/stale - would you mind deleting them for cleanliness?

what happens to closed PRs when you delete their branch?  do they keep giving useful info?  seems a bit of lost history if they'd become useless.

I'm kind of used to having many useless branches in git, it's a side-effect of how git works and it doesn't cost much i think, they're mostly just a ref afaik.

MarijnS95:
> > An unrelated comment to this PR: When cloning all branches from this repo are adopted with it, though most appear to be stale remnants from closed PRs: https://github.com/getsentry/sentry-rust/branches/stale - would you mind deleting them for cleanliness?
> 
> what happens to closed PRs when you delete their branch? do they keep giving useful info? seems a bit of lost history if they'd become useless.

History should still be accessible though the GitHub UI and fetchable from `refs/pull/<id>/{head,merge}`. If the branch with the same name is never pushed GitHub also offers a "restore" button on the PR, just in case. Note that some branches are 2-4 years old, I wonder how useful that still is :)

> I'm kind of used to having many useless branches in git, it's a side-effect of how git works and it doesn't cost much i think, they're mostly just a ref afaik.

Branches are cheap but do cause the commits they reference to be downloaded too, in a non-shallow clone. It's not computing/storage cost as the changes are generally tiny, but the mental overhead in seeing a bunch of unused, "meaningless" branches lingering around. I'm more used to clean upstream repos and PRs from forks (every branch has meaning, such as "stable" branches for older tags), though `gh pr create` lately sure mixes that up more often than I'd like.

Swatinem:
Most of those branches are mine ;-) And they contain a few ideas that I was kicking around very early, I would rather keep those if I ever come around to revisit that. But sure, some are also useless by now, but I don’t really want to look at each and every one in detail. I guess thats more of an argument to just get rid of them, but meh.

PR #426, opened 2022-01-31
meta: Add #519 (`ureq` transport support) to the changelog

MarijnS95:
The previous `meta` PR(s) for the changelog seemed autogenerated, but given recent commits I thought I should at least add my own items to the list to save you the hassle ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Added `ureq` transport support ([#419](https://github.com/getsentry/sentry-rust/pull/419))
+
 **Fixes**:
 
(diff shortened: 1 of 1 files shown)


flub:
thanks!

MarijnS95:
@flub Thank _you_ and this project in general for the overall quick responses and merges!

Then it only remains me to ask if we can have a new (patch) release with the new features, if it wasn't scheduled already :grimacing:

flub:
We don't really have an established way to decide on when we release, it's
a bit ad hoc.  Maybe we should figure out a way to improve this so it is
more predictable for people.

MarijnS95:
@flub In my (shared maintenance) crates, releases are ad-hoc as well: they usually happen when a new feature/fix lands and there's nothing directly in the pipeline for merging soon. Especially when a non-breaking patch release is possible (and I try to merge PRs in such order to facilitate that) and/or a contributor is craving for a release with his/her change :)

PR #428, opened 2022-02-02
Add GpuContext

Jake-Shadle:
This adds a GPU Context that matches the API description from [Sentry](https://develop.sentry.dev/sdk/event-payloads/contexts/#gpu-context), with one exception, I've added a `driver_version` field as that particular piece of information is much more important/interesting than some of the other fields that are already present in the format.


flub (approved):
thanks for also adding all the doc comments!

PR #430, opened 2022-02-03
feat(tracing): Add line information from tracing

flub:
This adds the line information that tracing provides back into the
event.  It also moves around the tags into a custom context instead of
the "extra" grab-bag as this is less likely to step on the user's
ad-hoc toes.

Related to #427

### DIFF
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -103,6 +103,36 @@ pub fn breadcrumb_from_event(event: &tracing_core::Event) -> Breadcrumb {
     }
 }
 
+fn contexts_from_event(
+    event: &tracing_core::Event,
+    event_tags: BTreeMap<String, Value>,
+) -> BTreeMap<String, sentry_core::protocol::Context> {
+    let event_meta = event.metadata();
+    let mut location_map = BTreeMap::new();
+    if let Some(module_path) = event_meta.module_path() {
+        location_map.insert("module_path".to_string(), module_path.into());
+    }
+    if let Some(file) = event_meta.file() {
+        location_map.insert("file".to_string(), file.into());
+    }
+    if let Some(line) = event_meta.line() {
+        location_map.insert("line".to_string(), line.into());
+    }
+
+    let mut context = BTreeMap::new();
+    context.insert(
+        "Rust Tracing Tags".to_string(),
+        sentry_core::protocol::Context::Other(event_tags),
+    );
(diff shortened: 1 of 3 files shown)


loewenheim (approved):

sentry-tracing/src/converters.rs
```diff
@@ -103,6 +103,36 @@ pub fn breadcrumb_from_event(event: &tracing_core::Event) -> Breadcrumb {
...
+        location_map.insert("file".to_string(), file.into());
+    }
+    if let Some(line) = event_meta.line() {
+        location_map.insert("line".to_string(), line.into());
+    }
+
+    let mut context = BTreeMap::new();
+    context.insert(
+        "Rust Tracing Tags".to_string(),
+        sentry_core::protocol::Context::Other(event_tags),
```
Swatinem, line 125: can we also wrap all of this in a `event_tags.is_empty` check?

sentry-tracing/src/converters.rs
```diff
@@ -103,6 +103,36 @@ pub fn breadcrumb_from_event(event: &tracing_core::Event) -> Breadcrumb {
...
+    );
+    if !location_map.is_empty() {
+        context.insert(
+            "Rust Tracing Location".to_string(),
```
Swatinem, line 129: its quite unfortunate we don’t have any other event attrs for source location :-(

sentry/tests/test_tracing.rs
```diff
@@ -64,6 +80,21 @@ fn test_tracing() {
...
+            let value = Value::String("tagvalue".to_string());
+            assert_eq!(*tags.get("tagname").unwrap(), value);
+        }
+        _ => panic!("Wrong context type"),
+    }
+    match event.contexts.get("Rust Tracing Location").unwrap() {
+        Context::Other(tags) => {
+            assert!(matches!(tags.get("module_path").unwrap(), Value::String(_)));
+            assert!(matches!(tags.get("file").unwrap(), Value::String(_)));
+            assert!(matches!(tags.get("line").unwrap(), Value::Number(_)));
```
Swatinem, line 94: might as well assert the file/line ;-)

Swatinem (approved):

flub (commented):
There are like lots more weird edge cases that i just don't understand what tracing is doing - which is why there's no test for the empty tags.  I'm giving up though as this seems hopeless, I think the current implementation gives some useful information so I'm going to close my eyes and click merge.

PR #433, opened 2022-02-11
ref: Forward set_transaction to a running Transaction

Swatinem:
A call to `Scope::set_transaction` will now also forward the given name to the currently running performance monitoring transaction to override the `name` it was started with.

### DIFF
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -160,6 +160,17 @@ impl Scope {
     /// Sets the transaction.
     pub fn set_transaction(&mut self, transaction: Option<&str>) {
         self.transaction = transaction.map(Arc::from);
+        if let Some(name) = transaction {
+            let trx = match self.span.as_ref() {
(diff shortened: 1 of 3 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -160,6 +160,17 @@ impl Scope {
     /// Sets the transaction.
     pub fn set_transaction(&mut self, transaction: Option<&str>) {
         self.transaction = transaction.map(Arc::from);
+        if let Some(name) = transaction {
```
flub, line 163: nit: if you're switching to `if let` i think it makes more sense to move line 162 into this block as well to keep the logic the same which make reading easier.

flub (approved):
lgtm but please commit with a description of why and what.  the title made no sense until after i understood the code, and but really the why is important to describe.

sentry-core/src/scope/real.rs
Swatinem: The line above should explicitly set `None` as well though

PR #444, opened 2022-03-09
build(backtrace): Update regex to 1.5.5

hannes-vernooij:
This fixes a security vulnerability. https://github.com/rust-lang/regex/commit/ae70b41d4f46641dbc45c7a4f87954aea356283e


jan-auer:
Thank you for the update!

For the record, `sentry-rust` is a library with a _minimum_ regex requirement, so the exact version of the regex crate is controlled by the target application. By bumping the version requirement here, we will force upgrades in all consumers, which is a good side effect.

There are no untrusted regexes used in this repository, and as such sentry-rust was not affected by the vulnerability.

jan-auer (approved):

PR #445, opened 2022-03-14
add initial values for functions to the data

### LINKED ISSUE #440: tracing: added fields don't show up in Sentry
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) 

Which SDK and version?
Rust, master

### Steps to Reproduce

Adding custom fields in the `tracing::instrument` don't show up anywhere in sentry.

example:
```
#[tracing::instrument(
        fields(
            db.name=%self.info.current_database,
            db.system="postgresql",
            db.version=%self.info.version,
            db.user=%self.info.user,
            db.statement=%query,
            otel.kind="client",
        ),
        skip(self),
    )]
```

### Expected Result

I kind thought those would be tags or data, judging by the code it looks like data but either way works.

### Actual Result

The data does not show up in sentry at all

jessfraz:
this fixes #440 

spans later in the functions were working and showing up in the UI but not the original set of values passed to the function.

### DIFF
--- a/sentry-tracing/src/layer.rs
+++ b/sentry-tracing/src/layer.rs
@@ -162,7 +162,7 @@ where
             return;
         }
 
-        let (description, _data) = extract_span_data(attrs);
+        let (description, data) = extract_span_data(attrs);
         let op = span.name();
 
         // Spans don't always have a description, this ensures our data is not empty,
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
oh, that sure was a big oversight, thanks for fixing this up!

A test for this might be nice. I think we do have tests for manual `record`, etc.

jessfraz:
will add some tests!

DmitrySamoylov:
But those are logged to console via `tracing_subscriber::fmt::layer()`, it doesn't have to do with Sentry, right? :thinking:

jessfraz:
yes for sure i was thinking its more a sentry UI issue

PR #451, opened 2022-03-25
feat(contexts): Report macOS version instead of Darwin kernel version

### LINKED ISSUE #449: Report macOS versions instead of Darwin kernel versions
In other Sentry SDKs, the version of macOS is either found directly by using Objective-C APIs (ex: [sentry-cocoa](https://github.com/getsentry/sentry-cocoa/blob/cd561c7b8f5e0396f3510600167c17b2c076deed/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_System.m#L510)) or using  the `sw_vers` command and falling back to a heuristic (ex: [sentry-electron](https://github.com/getsentry/sentry-electron/blob/11a6d8c64e5f26a5b95bc3144459f85ea8f38336/src/main/context.ts#L105)) to figure out the macOS version from the Darwin kernel version. Currently, errors coming from the Sentry Rust SDK [report the Darwin kernel version](https://github.com/getsentry/sentry-rust/blob/9dcbeebdf08e4eaa3f6007a3830a07834478d3cd/sentry-contexts/src/utils.rs#L90) instead of the macOS version. It would be great if the macOS version is reported, in line with the other SDKs. 

I think the best approach would be to do `sw_vers` and fall back to the heuristic, like the Electron SDK does, but this wouldn't work on iOS (maybe there's a heuristic for that too?). I'm happy to work on this change but I wanted to get some feedback before beginning.

rajivshah3:
In line with the other Sentry SDKs, the Rust SDK should report the macOS name ("macOS" instead of "Darwin"), version, and build. This PR uses the [approach](https://github.com/getsentry/sentry-native/blob/9eecb1bdf2c7762ba7babe9ea437545a1f759bdc/src/sentry_os.c#L130-L151) from `sentry-native` by calling `sysctlbyname` with`kern.osproductversion` and `kern.osversion`.

Closes #449

### DIFF
--- a/sentry-contexts/src/utils.rs
+++ b/sentry-contexts/src/utils.rs
@@ -119,15 +119,31 @@ pub fn os_context() -> Option<Context> {
     {
         use uname::uname;
         if let Ok(info) = uname() {
-            Some(
-                OsContext {
-                    name: Some(info.sysname),
-                    kernel_version: Some(info.version),
-                    version: Some(info.release),
-                    ..Default::default()
-                }
-                .into(),
-            )
+            #[cfg(target_os = "macos")]
+            {
+                Some(
+                    OsContext {
+                        name: Some("macOS".into()),
+                        kernel_version: Some(info.version),
+                        version: model_support::get_macos_version(),
+                        build: model_support::get_macos_build(),
+                        ..Default::default()
+                    }
+                    .into(),
+                )
+            }
+            #[cfg(not(target_os = "macos"))]
+            {
+                Some(
+                    OsContext {
+                        name: Some(info.sysname),
+                        kernel_version: Some(info.version),
+                        version: Some(info.release),
+                        ..Default::default()
+                    }
+                    .into(),
+                )
+            }
         } else {
             None
         }
--- a/sentry-contexts/src/utils.rs
+++ b/sentry-contexts/src/utils.rs
@@ -52,6 +52,14 @@ mod model_support {
         sysctlbyname_call("hw.model")
     }
 
+    pub fn get_macos_version() -> Option<String> {
+        sysctlbyname_call("kern.osproductversion")
+    }
+
+    pub fn get_macos_build() -> Option<String> {
+        sysctlbyname_call("kern.osversion")
+    }
+
     pub fn get_family() -> Option<String> {
         get_model().map(|mut s| {
             let len = s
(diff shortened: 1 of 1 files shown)


sentry-contexts/src/utils.rs
```diff
@@ -41,6 +48,18 @@ mod model_support {
         }
     }
 
+    pub fn get_model() -> Option<String> {
+        sysctlbyname_call("hw.model")
+    }
+
+    pub fn get_macos_version() -> Option<String> {
+        sysctlbyname_call("kern.osproductversion")
```
rajivshah3, line 56: `kern.osproductversion` seems to have been introduced in [xnu-4570.51.1](https://github.com/apple/darwin-xnu/commit/5bbb823c13f3ab1ab58878f96b35433a29882676#diff-1daf70413d68262120e719485aac47dd7da773d2333383c9fe5fd34dc93a5ffc), so it seems to be supported in macOS 10.13.4 and later (per [opensource.apple.com](https://opensource.apple.com/releases/)). What should the SDK do if `kern.osproductversion` is not available? Should it return `None` or use a heuristic to determine the version?

sentry-contexts/src/utils.rs
```diff
@@ -71,6 +105,14 @@ mod model_support {
...
         None
     }
+
+    pub fn get_macos_version() -> Option<String> {
+        None
+    }
+
+    pub fn get_macos_build() -> Option<String> {
+        None
+    }
```
Swatinem, line 115: I don’t think these are needed, I’m surprised this is not flagged as unused code even.

Swatinem (approved):

sentry-contexts/src/utils.rs
Swatinem: I think returning `None` is a good idea. Luckily macOS has very quick deprecation/EOL cycles, so you don’t need to worry about older versions for too long.

sentry-contexts/src/utils.rs
Swatinem: Ah, it is indeed flagged as unused in CI ;-)

Swatinem:
While you are at it, feel free to add this change to the [CHANGELOG.md](https://github.com/getsentry/sentry-rust/blob/master/CHANGELOG.md)

Swatinem (approved):

PR #452, opened 2022-03-29
ref: Change the internal representation of project IDs from u64s to strings

relaxolotl:
Does what it says on the tin. AFAIK this is because project IDs aren't guaranteed to be numbers forever, and there's a shared goal between multiple SDKs to switch their types over from some integer-y representation to a string representation, in preparation for a possible switch to a more sophisticated ID system.

Effort has been put in to ensure that most of the API remains the same, however there have been some changes to signatures whose original type would be expensive to retain given the unclonability of strings. 

jira breadcrumb: NATIVE-503

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Features**:
 
 - Request data can now be attached to Transactions and Spans via `set_transaction`. ([#439](https://github.com/getsentry/sentry-rust/pull/439))
+- macOS versions are now reported instead of the Darwin kernel version. ([#451](https://github.com/getsentry/sentry-rust/pull/451))
 
 **Thank you**:
 
--- a/sentry-contexts/src/utils.rs
+++ b/sentry-contexts/src/utils.rs
@@ -5,13 +5,18 @@ include!(concat!(env!("OUT_DIR"), "/constants.gen.rs"));
 #[cfg(target_os = "macos")]
 mod model_support {
     use libc::c_void;
+    use std::ffi::CString;
     use std::ptr;
 
-    pub fn get_model() -> Option<String> {
+    fn sysctlbyname_call(name: &str) -> Option<String> {
         unsafe {
+            let c_name = match CString::new(name) {
+                Ok(name) => name.into_bytes_with_nul(),
+                Err(_e) => return None,
+            };
             let mut size = 0;
             let res = libc::sysctlbyname(
-                "hw.model\x00".as_ptr() as _,
+                c_name.as_ptr() as _,
                 ptr::null_mut(),
                 &mut size,
                 ptr::null_mut(),
@@ -20,9 +25,10 @@ mod model_support {
             if res != 0 {
                 return None;
             }
+
             let mut buf = vec![0u8; size as usize];
             let res = libc::sysctlbyname(
-                "hw.model\x00".as_ptr() as _,
+                c_name.as_ptr() as _,
                 buf.as_mut_ptr() as *mut c_void,
                 &mut size,
                 ptr::null_mut(),
@@ -31,6 +37,7 @@ mod model_support {
             if res != 0 {
                 return None;
             }
+
             Some(
                 buf.into_iter()
                     .take(size)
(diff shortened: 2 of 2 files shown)


sentry-types/src/dsn.rs
```diff
@@ -140,8 +140,8 @@ impl Dsn {
     }
 
     /// Returns the project_id
-    pub fn project_id(&self) -> ProjectId {
-        self.project_id
+    pub fn project_id(&self) -> &ProjectId {
```
relaxolotl, line 143: As noted in the description, this signature change may impact some users. I'm actually not too sure if too many folks actually make active use of this value seeing as it isn't required as a parameter or anything like that in the rest of the SDK's API. If I had to guess, this change is probably relatively low-impact but I could be wrong about this.

sentry-types/src/project_id.rs
```diff
@@ -17,26 +17,28 @@ pub enum ParseProjectIdError {
 }
 
 /// Represents a project ID.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Hash, Deserialize, Serialize)]
-pub struct ProjectId(u64);
+#[derive(Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Hash, Deserialize, Serialize)]
+#[serde(into = "u64", from = "u64")]
```
Swatinem, line 21: IMO, we should not serialize this to `u64`, but use String there as well.
Same down below for `new` and `value`, those should take/return `&str` as well.

relaxolotl:
Updated to reflect the original expectations that motivated this change (zero validation of the contents of the project ID). Reasonable checks have been kept (i.e. emptiness) but all other validation has been stripped away. All of the conversions from and to various integer values have been removed as the project ID should no longer be considered a newtype that thinly wraps around a string. This should be ready for another review.

sentry-types/src/project_id.rs
```diff
@@ -8,88 +7,45 @@ use thiserror::Error;
...
-        impl From<$ty> for ProjectId {
-            #[inline]
-            fn from(val: $ty) -> Self {
-                Self::new(val as u64)
-            }
-        }
-    };
-}
-
-impl_from!(u8);
```
Swatinem, line 54: IMO, we could keep the Try/From impls, as they would just stringify the given int. But its also fine to just remove it.

sentry-types/src/dsn.rs
Swatinem: I’m wondering if it might be simpler to just return `&str` here, and completely remove the `ProjectId` type.

Swatinem (approved):
lgtm. You can land this as-is, or consider removing the type completely. TBH, the Rust SDK went overboard with strict typing in some places, leading to a huge and unwieldy API surface, so it might as well be a good idea to cut back on that.

sentry-types/src/dsn.rs
flub: leaving it as an explicit type makes sure it's opaque and user's can't be tempted to try and parse things out of it, so I'd keep it.

relaxolotl: i was thinking about having this return `&str` initially and as suggested, to get rid of the type completely. 

if we were to switch to something like a UUID or otherwise well-structured unique identifier, we could add back in validation to this type that we'd removed. in an ideal situation this would mean not having to rewrite and convert all of the project ID code back to using a newtype. 

even if we're unable to reuse the code at all, i think it would make it easier for us to easily find and replace instances of project IDs with their replacements in the future. as @flub has mentioned, having a separate type makes it harder for users to do too much with the ID, but i suppose it's not like we're trying particularly hard to prevent them from doing so either by exposing `ProjectId::value`. 

still, i'm tempted to keep the type because it helps explicitly express an opinion about what `ProjectId`s are and how to work with them. we'll see if this burns us sometime in the future 🙃

sentry-types/src/project_id.rs
relaxolotl: i'll follow up on this in a different place, but i'm curious to hear the argument for keeping the Try/From impls. i've got an inkling of an idea, but it could just be completely wrong. 

i'm open to adding these back in a separate PR though. right now it feels like keeping impls just for integer values looks a little weird: why are there `Try/From` impls for integers but `parse` for `&str`s? why aren't there `try_from` impls for other primitives, why specifically all of the ints?

PR #453, opened 2022-04-01
fix: Handle #[instrument(err)] by extracting the error message

kvnvelasco:
When using `#[tracing::instrument(err)]` to emit error events when the wrapped function returns an error. Sentry logs show the event as an "unlabled event" 

image

This PR extracts the error message as a string when the message cannot be found.


Swatinem (approved):
lgtm. Would be nice to have a test for this in `sentry/tests/test_tracing.rs`

kvnvelasco:
> lgtm. Would be nice to have a test for this in `sentry/tests/test_tracing.rs`

Added a test case in `test_tracing`

PR #454, opened 2022-04-07
Take advantage of weak features in Rust 1.60 for TLS enablement

MarijnS95:
Previously it was impossible to have a unified feature named `rustls` or `native-tls` that would turn on the respective TLS backend in the chosen transport (`reqwest` or `ureq`) as a feature of `<crate>/tls` would implicitly turn on `<crate>`.  Since Rust 1.60 it is now possible to specify this crate-feature enablement through the use of the question mark in `<crate>?/tls`, which will only enable the `tls` feature if `<crate>` was enabled through other means (another feature).

Secondly we can now also let optional crates have the same name as a feature, and use `dep:<crate>` to refer to the crate instead of the dependency, making for a more pleasant experience without renames to underscore suffixes.

---

Filing this as draft just to get the word out and discuss this for a bit, as it will only be mergeable in 6 months in accordance with the MSRV policy.  It goes without saying that this is a breaking change.

Also note that I haven't renamed other optional crates like `anyhow_`, `log_` etc yet, until this is accepted as a (future) solution.

### DIFF
--- a/sentry/Cargo.toml
+++ b/sentry/Cargo.toml
@@ -39,17 +39,17 @@ profiling = ["sentry-core/profiling"]
 frame-pointer = ["sentry-core/frame-pointer"]
 # other features
 test = ["sentry-core/test"]
-debug-logs = ["log_", "sentry-core/debug-logs"]
+debug-logs = ["dep:log", "sentry-core/debug-logs"]
 # transports
 transport = ["reqwest", "native-tls"]
-reqwest = ["reqwest_", "httpdate", "tokio"]
-curl = ["curl_", "httpdate"]
-surf-h1 = ["surf_/h1-client", "httpdate"]
-surf = ["surf_/curl-client", "http-client", "httpdate", "isahc", "tokio"]
-native-tls = ["reqwest_/default-tls", "native-tls_", "ureq-native-tls"]
-rustls = ["reqwest_/rustls-tls", "rustls_", "webpki-roots"]
-ureq = ["ureq_/tls", "httpdate", "rustls_", "webpki-roots"]
-ureq-native-tls = ["ureq_/native-tls", "httpdate", "native-tls_"]
+reqwest = ["dep:reqwest", "httpdate", "tokio"]
+curl = ["dep:curl", "httpdate"]
+surf-h1 = ["surf/h1-client", "httpdate"]
+surf = ["surf/curl-client", "http-client", "httpdate", "isahc", "tokio"]
+ureq = ["dep:ureq", "httpdate"]
+# transport settings
+native-tls = ["dep:native-tls", "reqwest?/default-tls", "ureq?/native-tls"]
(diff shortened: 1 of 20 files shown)


sentry/src/transports/ureq.rs
```diff
@@ -1,8 +1,7 @@
 use std::time::Duration;
 
-#[cfg(fdoc)]
```
MarijnS95, line 3: Blame says that I misspelled this :grimacing:

MarijnS95: In turn `` [`ureq`]: https://crates.io/crates/ureq `` below wouldn't have been needed...

sentry/Cargo.toml
```diff
@@ -11,6 +11,7 @@ Sentry (getsentry.com) client for rust ;)
 """
 edition = "2018"
 autoexamples = true
+rust-version = "1.54.0"
```
flub, line 14: Wouldn't this have to be 1.60 with this PR?

sentry/src/transports/curl.rs
```diff
@@ -1,7 +1,7 @@
 use std::io::{Cursor, Read};
 use std::time::Duration;
 
-use curl_::{self as curl, easy::Easy as CurlClient};
+use curl::{self as curl, easy::Easy as CurlClient};
```
flub, line 4: `self as curl` can be just `self` now

flub (commented):
Agree that this would be a nice improvement.  Though you correctly note that we can't just change this right now and need to wait a while.

sentry/Cargo.toml
MarijnS95: Good catch - was already wondering why the CI didn't throw a human-readable error in the MSRV tests...

Swatinem:
@MarijnS95 Its been a while ;-)
We have now bumped MSRV to 1.60 in master, so the PR is unblocked now.
Are you up to update it? Otherwise I might take over the PR in the coming days.

MarijnS95:
@Swatinem cool, sure! That was a bit more work than expected, with both conflicts and more crates in other `Cargo.toml` that were imported with `_` suffixes (and I either missed them when initially creating this PR, or they were added recently).

Let's see how the CI likes this :)

sentry/src/transports/ureq.rs
MarijnS95: Looks like I fixed this in #455 already.

Swatinem (approved):
nice!

there is a few places that still has `pkg = { package = "pkg", …`

It might be worth removing that, but its also fine to merge like this.

MarijnS95:
@Swatinem Good catch, looks like I missed those while resolving conflicts!

There was also a stray `anyhow_` `dev-dependencies` import remaining, but it's not optional so shouldn't conflict with the `anyhow` feature.

Swatinem (approved):

PR #455, opened 2022-04-07
Build-test documentation in the CI and fix/improve broken links

MarijnS95:
Some intradoc links were broken because their crates are only available as `dev-dependencies`, while others unnecessarily had a forced link to `docs.rs` which could also be an intradoc link (in turn pointing to locally built documentation if not building on docs.rs).

To make sure these links don't break again over time, build-test the documentation in the CI while disallowing any warnings which includes broken links.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,3 +159,29 @@ jobs:
         with:
           directory: coverage
           #files: lcov.info
+
+  doc:
+    name: Build-test documentation
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          components: rust-docs
+
+      - uses: Swatinem/rust-cache@v1
+
+      - name: Run cargo doc
+        uses: actions-rs/cargo@v1
+        env:
+          RUSTDOCFLAGS: -Dwarnings
+        with:
+          command: doc
+          args: --workspace --all-features --no-deps --document-private-items
--- a/sentry-anyhow/src/lib.rs
+++ b/sentry-anyhow/src/lib.rs
@@ -59,12 +59,8 @@ pub fn capture_anyhow(e: &anyhow::Error) -> Uuid {
 }
 
 /// Hub extension methods for working with [`anyhow`].
-///
-/// [`anyhow`]: https://docs.rs/anyhow
 pub trait AnyhowHubExt {
     /// Captures an [`anyhow::Error`] on a specific hub.
-    ///
-    /// [`anyhow::Error`]: https://docs.rs/anyhow/*/anyhow/struct.Error.html
     fn capture_anyhow(&self, e: &anyhow::Error) -> Uuid;
 }
 
--- a/sentry-tower/src/lib.rs
+++ b/sentry-tower/src/lib.rs
@@ -127,6 +127,8 @@
 //!     .layer(sentry_tower::SentryHttpLayer::with_transaction());
 //! # }
 //! ```
+//!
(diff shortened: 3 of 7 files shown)


loewenheim (approved):

.github/workflows/ci.yml
```diff
@@ -159,3 +159,34 @@ jobs:
         with:
           directory: coverage
           #files: lcov.info
+
+  doc:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        rust: [nightly]
```
flub, line 168: Does this make sense for docs?  E.g. in symbolicator we check the doc comments simply on `runs-on: ubuntu-latest` and with stable rust

.github/workflows/ci.yml
```diff
@@ -159,3 +159,34 @@ jobs:
...
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
```
flub, line 182: I think you can add `components: rust-docs` here to make this a little faster.

.github/workflows/ci.yml
```diff
@@ -159,3 +159,34 @@ jobs:
...
+
+      - uses: Swatinem/rust-cache@v1
+
+      - name: Run cargo doc
+        uses: actions-rs/cargo@v1
+        env:
+          RUSTDOCFLAGS: -Dwarnings --cfg doc_cfg
+        with:
+          command: doc
+          args: --workspace --all-features --no-deps
```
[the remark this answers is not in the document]
flub: Currently there seems to be no errors when adding `--document-private-items` so I would be fine with adding it for now.  If it becomes a pain we can always remove it again.

flub (commented):
thanks for adding this!

.github/workflows/ci.yml
MarijnS95: For these specific tests we can do it without, as long as `--cfg doc_cfg` isn't replicated from `docs.rs` setup.

.github/workflows/ci.yml
MarijnS95: Isn't that additional, on top of what is already installed by `profile: minimal`?

(I'm surprised that this works, since `rust-docs` is only included in https://rust-lang.github.io/rustup/concepts/profiles.html - perhaps it's pulling it from GH's runner/)

MarijnS95: EDIT: Perhaps it doesn't work. The CI in this PR seems to run the workflow from `master`, so _excluding_ my job (something to do with possibly exposing secrets without maintainer approval, I suppose), and I wouldn't have caught this as the `on:` branch filter doesn't allow my branchname.

.github/workflows/ci.yml
flub: I'm not sure what you're saying here, you are adding `--cfg doc_cfg` below?  Is it preferable to check the docs on all platforms and with nightly?

I'm just worried we're blowing up the amount of CI resources for no real benefit with this matrix.

MarijnS95: 1. `nightly`: This is needed to enable `--cfg doc_cfg` for `#![cfg_attr(doc_cfg, feature(doc_cfg))]`. It doesn't seem worth it, since `rustdoc` doesn't validate `doc(cfg())` features to actually exist :(;
2. Having 3 `os`'es in a matrix is not needed, there are only very few os `cfg`s that have no effect on documentation. I'll remove them.

MarijnS95: Sorry for all the noise (repeated pushes) in all PRs, great lack of focus for the simplest of things :weary:

flub (approved):

.github/workflows/ci.yml
flub: Oh, I forgot that `--cfg doc_cfg` needs nightly.  thanks for reminding me :)

Issue #381, opened 2021-10-29
Support sending a serialized envelope via Transport


kornelski:
I need to report errors from a process that runs in an off-line sandbox, without its own access to a network. The process can serialize and pass on messages to be sent from another process (like Chrome Renderer helper and Chrome root process). Related to #281

Sentry supports pluggable `Transport` which can be used to capture an `Envelope`, and `Envelope` supports serialization via `to_writer`, so I can move it between processes easily. However, the `Transport` trait can't send a serialized `&[u8]` representation of the envelope, so I can't reuse Sentry's implementation of networked transports. I need to reinvent the whole transport logic myself, including retry-after and rate limiting logic, which I'd rather not duplicate. 

* Could you extend the `Transport` trait to support sending of an already-serialized envelope from `&[u8]` or `Vec<u8>` representation?

* Or, could you make it possible to parse `Envelope` object back from its serialized representation, so that I can feed it back to a `Transport`?

* Alternatively, could you make [`TransportThread`](https://github.com/getsentry/sentry-rust/blob/a77bdcefd7c3e976391587549e1a5818a192821b/sentry/src/transports/thread.rs) public, so I would only need to copy-paste only the easy parts of an http client implementation?

Swatinem:
I think one way to solve this would be to create a "raw" `Envelope`, similar to what we have in `sentry-native`.

Also, I wanted to expose the retry-after and rate-limiting logic so that it can be reused by third parties. I think we will be doing that soon-ish.

kamilogorek:
@Swatinem please ping me once it's decided and implemented, so I'll pull this down to `sentry-cli` as well :)

smeubank:
@Swatinem any update on this? We added this to our sentry-cli roadmap #1168

Swatinem:
We haven’t planned for this yet. However I don’t expect this to be a huge change.

PR #456, opened 2022-04-14
feat: Implement Envelope::from_path and Envelope::from_slice

kamilogorek:
The implementation is highly inspired by https://github.com/getsentry/relay/blob/master/relay-server/src/envelope.rs and it includes most its deserialization tests.

We do not handle unknown item types for now, as it would require significant changes in how serializing works and how we store the envelopes. Thus we decided to skip it, for now, to support only item types that `sentry-rust` implements and enhance this feature in the future.

This will be enough to unblock https://github.com/getsentry/sentry-cli/issues/703 anywya.
Closes https://github.com/getsentry/sentry-rust/issues/381

### DIFF
--- a/sentry-types/src/protocol/attachment.rs
+++ b/sentry-types/src/protocol/attachment.rs
@@ -1,20 +1,27 @@
 use std::fmt;
 
+use serde::Deserialize;
+
 /// The different types an attachment can have.
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Deserialize)]
 pub enum AttachmentType {
+    #[serde(rename = "event.attachment")]
     /// (default) A standard attachment without special meaning.
     Attachment,
     /// A minidump file that creates an error event and is symbolicated. The
     /// file should start with the `MDMP` magic bytes.
+    #[serde(rename = "event.minidump")]
     Minidump,
     /// An Apple crash report file that creates an error event and is symbolicated.
+    #[serde(rename = "event.applecrashreport")]
     AppleCrashReport,
     /// An XML file containing UE4 crash meta data. During event ingestion,
     /// event contexts and extra fields are extracted from this file.
+    #[serde(rename = "unreal.context")]
     UnrealContext,
     /// A plain-text log file obtained from UE4 crashes. During event ingestion,
     /// the last logs are extracted into event breadcrumbs.
+    #[serde(rename = "unreal.logs")]
     UnrealLogs,
 }
 
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -1,8 +1,75 @@
-use std::io::Write;
+use std::{io::Write, path::Path};
 
+use serde::Deserialize;
+use thiserror::Error;
 use uuid::Uuid;
 
-use super::v7::{Attachment, Event, SessionAggregates, SessionUpdate, Transaction};
+use super::{
+    attachment::AttachmentType,
+    v7::{Attachment, Event, SessionAggregates, SessionUpdate, Transaction},
+};
+
+/// Raised if a envelope cannot be parsed from a given input.
+#[derive(Debug, Error)]
+pub enum EnvelopeError {
+    /// Unexpected end of file
+    #[error("unexpected end of file")]
+    UnexpectedEof,
+    /// Missing envelope header
+    #[error("missing envelope header")]
+    MissingHeader,
+    /// Missing item header
+    #[error("missing item header")]
+    MissingItemHeader,
+    /// Missing newline after header or payload
+    #[error("missing newline after header or payload")]
+    MissingNewline,
+    /// Invalid envelope header
+    #[error("invalid envelope header")]
+    InvalidHeader(#[source] serde_json::Error),
+    /// Invalid item header
+    #[error("invalid item header")]
+    InvalidItemHeader(#[source] serde_json::Error),
+    /// Invalid item payload
+    #[error("invalid item payload")]
+    InvalidItemPayload(#[source] serde_json::Error),
+}
+
+#[derive(Deserialize)]
+struct EnvelopeHeader {
+    event_id: Option<Uuid>,
+}
+
+/// An Envelope Item Type.
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize)]
+enum EnvelopeItemType {
+    /// An Event Item type.
+    #[serde(rename = "event")]
+    Event,
+    /// A Session Item type.
+    #[serde(rename = "session")]
+    SessionUpdate,
+    /// A Session Aggregates Item type.
+    #[serde(rename = "sessions")]
+    SessionAggregates,
+    /// A Transaction Item type.
+    #[serde(rename = "transaction")]
+    Transaction,
+    /// An Attachment Item type.
+    #[serde(rename = "attachment")]
+    Attachment,
+}
+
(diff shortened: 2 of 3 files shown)


Swatinem (changes requested):
I’m a bit surprised any of this works, lol. Probably because it just skips all the item headers as something that is not parsable, and then parses the item payload itself. That in turn means you will have problems parsing an `Attachment` item, which I see is also missing a test, probably for that reason ;-)

Instead of using serde magic, it might be a better idea to manually parse/match the item headers. Might as well create a `derive(Deserialize)` `ItemHeader` type, which has at least a `type` field, and possibly a `length` as well. In order to support attachments, you would have to make your parser more sophisticated as well, as going by `lines()` will break for attachments.

sentry-types/src/protocol/attachment.rs
```diff
@@ -37,7 +39,7 @@ impl AttachmentType {
     }
 }
 
-#[derive(Clone, PartialEq, Default)]
+#[derive(Clone, PartialEq, Deserialize, Default)]
```
Swatinem, line 42: I don’t think we use the `Deserialize` directly, since attachments have their own deserialize logic.

sentry-types/src/protocol/envelope.rs
```diff
@@ -1,16 +1,82 @@
...
+    /// Invalid item header
+    #[error("invalid item header")]
+    InvalidItemHeader(#[source] serde_json::Error),
+    /// Invalid item payload
+    #[error("invalid item payload")]
+    InvalidItemPayload(#[source] serde_json::Error),
+}
+
+#[derive(Deserialize)]
+struct EnvelopeHeader {
```
Swatinem, line 36: Should we care about more optional attributes like `sent_at`, `dsn`, etc? Probably not now, since envelope "round-trips" are not really a goal for now.

sentry-types/src/protocol/envelope.rs
```diff
@@ -1,16 +1,82 @@
...
+    /// A Transaction Item type.
+    #[serde(rename = "transaction")]
+    Transaction,
+    /// An Attachment Item type.
+    #[serde(rename = "attachment")]
+    Attachment,
+}
+
+/// An Envelope Item Header.
+#[allow(dead_code)]
```
Swatinem, line 61:
```suggestion
```
We *are* using this, maybe we are not exposing the transitive users publicly?

sentry-types/src/protocol/envelope.rs
```diff
@@ -1,16 +1,82 @@
...
+
 /// An Envelope Item.
 ///
 /// See the [documentation on Items](https://develop.sentry.dev/sdk/envelopes/#items)
 /// for more details.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Deserialize, PartialEq)]
 #[non_exhaustive]
 #[allow(clippy::large_enum_variant)]
+#[serde(untagged)]
```
Swatinem, line 79: We don’t `Deserialize` this enum on itself, but rather the inner types directly, so we don’t need the annotations here.

sentry-types/src/protocol/envelope.rs
```diff
@@ -236,6 +302,125 @@ impl Envelope {
...
+        // Each header is terminated by a UNIX newline.
+        Self::require_termination(slice, stream.byte_offset())?;
+
+        Ok((header, stream.byte_offset() + 1))
+    }
+
+    fn parse_items(slice: &[u8], mut offset: usize) -> Result<Vec<EnvelopeItem>, EnvelopeError> {
+        let mut items = Vec::new();
+
+        while offset < slice.len() {
```
Swatinem, line 347: Instead of having a `mut offset` and sub-slicing from the original `slice` all the time, we can just have `mut slice` and do a `slice = slice.get(offset..)?` after we read each item, and above in `from_slice`.
But it doesn’t really matter that much anyway.

sentry-types/src/protocol/envelope.rs
```diff
@@ -236,6 +302,125 @@ impl Envelope {
...
+                let payload_end = payload_start + len as usize;
+                if slice.len() < payload_end {
+                    return Err(EnvelopeError::UnexpectedEof);
+                }
+
+                // Each payload is terminated by a UNIX newline.
+                Self::require_termination(slice, payload_end)?;
+                payload_end
+            }
+            None => match slice[payload_start..].iter().position(|b| *b == b'\n') {
```
Swatinem, line 386: the comment above says:

>`payload_start` may point past the end of the buffer.

so the sub-slice here can definitely panic, better to use `.get()`.

sentry-types/src/protocol/envelope.rs
```diff
@@ -236,6 +302,125 @@ impl Envelope {
...
+                Some(relative_end) => payload_start + relative_end,
+                None => slice.len(),
+            },
+        };
+
+        let payload = slice.get(payload_start..payload_end).unwrap();
+
+        let item = match header.r#type {
+            EnvelopeItemType::Event => EnvelopeItem::Event(
+                serde_json::from_slice(payload).map_err(EnvelopeError::InvalidItemPayload)?,
```
Swatinem, line 396: This duplicated `map_err` is a bit annoying.

We could potentially make this `serde_json::from_slice(payload).map(EnvelopeItem::Event)`, and an explicit `Ok(EnvelopeItem::Attachment(…))` below.
That way we can move the `.map_err(…)?` outside the match. That can potentially improve readability.

sentry-types/src/protocol/envelope.rs
```diff
@@ -256,13 +441,19 @@ impl From<Transaction<'static>> for Envelope {
...
+    };
 
     use time::format_description::well_known::Rfc3339;
     use time::OffsetDateTime;
 
     use super::*;
-    use crate::protocol::v7::{SessionAttributes, SessionStatus, Span};
+    use crate::protocol::{
+        latest::Level,
+        v7::{SessionAttributes, SessionStatus, Span},
```
Swatinem, line 455: `latest == v7`, so I guess this was an auto-import. Might be worth deduplicating.

sentry-types/src/protocol/envelope.rs
```diff
@@ -256,13 +441,19 @@ impl From<Transaction<'static>> for Envelope {
 
 #[cfg(test)]
 mod test {
-    use std::time::{Duration, SystemTime};
+    use std::{
+        str::FromStr,
+        time::{Duration, SystemTime},
+    };
```
Swatinem, line 447: This is just cosmetic really (and probably something that rustfmt/rust-analyzer should just autofix), throughout the rest of codebase, we prefer separate lines for `std` imports, instead of nesting them. Maybe the rust-analyzer default is set differently.
Doesn’t matter really, but we should at least be consistent.

sentry-types/src/protocol/envelope.rs
kamilogorek: We would need to change `to_writer` as well to handle this, as we currently ignore anything else other than `event_id`.

sentry-types/src/protocol/envelope.rs
kamilogorek: It was due to unused `attachment_type`.  Fixed it by using the value from item header and implementing `FromStr` for an `AttachmentType`

sentry-types/src/protocol/envelope.rs
kamilogorek: So much better now :)

Swatinem (approved):

PR #466, opened 2022-05-24
feat: Basic attachment support

timfish:
- Tested working against sentry.io
- Follows the other sdks by adding attachments to the scope

You can capture an event with a minidump attachment like:
```rust
    sentry::with_scope(
        |scope| {
            scope.add_attachment(Attachment {
                filename: "minidump.dmp".to_owned(),
                buffer: Vec::new(), // <-- minidump bytes go here
                ty: Some(AttachmentType::Minidump),
                content_type: None,
            })
        },
        || capture_event(Default::default()),
    );
```

### DIFF
--- a/sentry/tests/test_basic.rs
+++ b/sentry/tests/test_basic.rs
@@ -177,52 +177,27 @@ fn test_attached_stacktrace() {
 
 #[test]
 fn test_attachment_sent_from_scope() {
-    struct TestTransport(Arc<AtomicUsize>);
-
-    impl sentry::Transport for TestTransport {
-        fn send_envelope(&self, envelope: sentry::Envelope) {
-            for item in envelope.items() {
-                if let EnvelopeItem::Attachment(attachment) = item {
-                    assert_eq!(attachment.filename, "test-file.txt");
-                    assert_eq!(attachment.buffer, vec![1, 2, 3, 4, 5, 6, 7, 8, 9]);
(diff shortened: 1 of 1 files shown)


sentry/tests/test_basic.rs
```diff
@@ -173,3 +174,55 @@ fn test_attached_stacktrace() {
         .flat_map(|ev| ev.threads.into_iter().filter_map(|thrd| thrd.stacktrace));
     assert_eq!(stacktraces.count(), 3);
 }
+
+#[test]
+fn test_attachment_sent_from_scope() {
+    struct TestTransport(Arc<AtomicUsize>);
+
+    impl sentry::Transport for TestTransport {
+        fn send_envelope(&self, envelope: sentry::Envelope) {
```
Swatinem, line 183: This whole machinery is a bit over-engineered. You have [`sentry::test::with_captured_envelopes`](https://docs.rs/sentry/latest/sentry/test/fn.with_captured_envelopes.html) that does all this for you ;-) No need to define a transport or construct a client/hub.

Swatinem (approved):

Swatinem (approved):

sentry/tests/test_basic.rs
Swatinem: Ah, I think that test was really about testing the transport factory. All the other tests in that file use the appropriate helpers otherwise.

PR #469, opened 2022-05-26
feat: Add scope listener API

timfish:
I've been [messing around](https://github.com/timfish/sentry-rust-minidump) with the `minidumper` and `crash-handler` crates and now I seem to have minidumps being captured in pure Rust and sent via Sentry Rust.

Minidumps are sent from a separate process for reliability but this means for now they are missing all the scope context from the main app process. Adding a "scope updater listener" API would allow subscribing to scope updates and passing them to the crash reporter process to be included with minidump submissions.

I haven't checked all the SDKs, but in the JavaScript SDK there is an `scope.addScopeListener()` method that lets you subscribe to scope updates. We've made use of this in the Electron SDK to intercept, send and synchronise scope between the multiple processes.

This is in no way a finished PR, more a "is this likely to be accepted" and if so, "am I going in the right direction" PR 😊

### DIFF
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -8,12 +8,24 @@ use crate::protocol::{Attachment, Breadcrumb, Context, Event, Level, User, Value
 use crate::session::Session;
 use crate::Client;
 
+#[non_exhaustive]
+pub enum ScopeUpdate {
+    AddBreadcrumb(Breadcrumb),
+    ClearBreadcrumbs,
+    User(Option<User>),
+    SetExtra(String, Value),
+    RemoveExtra(String),
+    SetTag(String, String),
+    RemoveTag(String),
+}
+
 #[derive(Debug)]
 pub struct Stack {
     layers: Vec<StackLayer>,
 }
 
 pub type EventProcessor = Arc<dyn Fn(Event<'static>) -> Option<Event<'static>> + Send + Sync>;
+pub type ScopeListener = Arc<dyn Fn(&ScopeUpdate) + Send + Sync>;
 
 /// Holds contextual data for the current scope.
 ///
@@ -44,6 +56,7 @@ pub struct Scope {
     pub(crate) tags: Arc<HashMap<String, String>>,
     pub(crate) contexts: Arc<HashMap<String, Context>>,
     pub(crate) event_processors: Arc<Vec<EventProcessor>>,
+    pub(crate) scope_listeners: Arc<Vec<ScopeListener>>,
     pub(crate) session: Arc<Mutex<Option<Session>>>,
     pub(crate) span: Arc<Option<TransactionOrSpan>>,
     pub(crate) attachments: Arc<Vec<Attachment>>,
@@ -144,6 +157,7 @@ impl Scope {
 
     /// Deletes current breadcrumbs from the scope.
     pub fn clear_breadcrumbs(&mut self) {
+        self.notify_scope_listeners(|| ScopeUpdate::ClearBreadcrumbs);
         self.breadcrumbs = Default::default();
     }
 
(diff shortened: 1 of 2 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -8,12 +8,24 @@ use crate::protocol::{Attachment, Breadcrumb, Context, Event, Level, User, Value
 use crate::session::Session;
 use crate::Client;
 
+#[non_exhaustive]
+pub enum ScopeUpdate {
+    AddBreadcrumb(Breadcrumb),
+    ClearBreadcrumbs,
+    User(Option<User>),
```
Swatinem, line 15:
```suggestion
    SetUser(Option<User>),
```

sentry-core/src/scope/real.rs
```diff
@@ -8,12 +8,24 @@ use crate::protocol::{Attachment, Breadcrumb, Context, Event, Level, User, Value
 use crate::session::Session;
 use crate::Client;
 
+#[non_exhaustive]
+pub enum ScopeUpdate {
```
Swatinem, line 12: at least this type should live in the public API, independent of the `client` feature.

Swatinem (commented):
This is a good start.
What makes this a bit challenging is the fact that the Rust SDK has multiple hubs.

So this change by itself is nice, but the bigger things around it need to be considered if you were to attach the scope to a crash event. Which scope? Essentially the one for the "currently active Hub". But you need that info somehow. If you have that, you might as well access the hub/scope directly? Otherwise you have to stash all kinds of info away "somewhere" and get it using some kind of unique ID based on the currently active Hub.

Also if you work with updates instead of snapshots, it might become a bit complicated when we take Hub cloning / Scope pushing into account.

timfish:
> it might become a bit complicated when we take Hub cloning / Scope pushing into account.

I do agree that this kind of API is at odds with how hubs and the scope stack works. It also completely ignores sessions and transactions!

> the bigger things around it need to be considered if you were to attach the scope to a crash event. Which scope? 

Yes that is a problem. As with the Electron SDK, minidump submission occurs from another process and the async nature of IPC + minidump generation means that any synced scope is not guaranteed to match up correctly with when the event is captured. 

I double checked the JavaScript `scope.addScopeListener()` usage and it looks like it's only used for Electron but I could have missed something somewhere else. `sentry-dotnet` has a similar [`IScopeObserver`](https://getsentry.github.io/sentry-dotnet/api/Sentry.IScopeObserver.html) API which it says is "for the sync. of Scopes across SDKs" and further searching suggests it's used for the Unity SDK. 

So it appears this is a common pattern used to sync scope between SKDs or processes but in all cases it appears to ignore the scope stack and anything but the global hub!

>  If you have that, you might as well access the hub/scope directly?

Making the scope fields public would allow reading the scope directly, but it's still handy to know **when** the scope has changed and even handier to know exactly what changed.

Swatinem:
> Making the scope fields public would allow reading the scope directly

I wouldn’t make the fields public. Due to the real/noop scope optimization, that is impossible to begin with ;-)

Rather you could apply the scope to an (empty) event, since thats essentially what you want to achieve.

So yes, all in all this API is perfectly fine, its just hard to see how it can be used in a meaningful way.

timfish:
> I wouldn’t make the fields public. Due to the real/noop scope optimization, that is impossible to begin with ;-)

😬

> So yes, all in all this API is perfectly fine, its just hard to see how it can be used in a meaningful way.

Rather than try and describe what my plans were I realised it would be much quicker to try it out! 

Since I'm starting the Sentry SDK in both the main process and the crash reporting process, I planned to hook the scope changes in the main process (the client), serialise, send via IPC and then apply them to the global hub scope in the crash reporter process (the server) like this:
https://github.com/timfish/sentry-rust-minidump/pull/1/files

Then when a minidump event is sent from the crash reporter process, it will include that scope that was added via IPC messages.

szokeasaurusrex:
Closing as this PR is likely out of date and has been inactive for some time

PR #471, opened 2022-06-11
Replace ancient `lazy_static` crate with `once_cell` or `const` slices

MarijnS95:
Piggybacking on the [motivation in winit]: `lazy_static!` is a macro whereas `once_cell` achieves the same using generics.  Its implementation has also been [proposed for inclusion in `std`], making it easier to switch to a standardized version if/when that happens.  The author of that winit PR is making this change to many more crates, slowly turning the scales in favour of `once_cell` in most dependency trees.  Furthermore `lazy_static` hasn't published any updates for 3 years.

See also [the `once_cell` F.A.Q.].

In addition "constant" `Vec`tor allocations don't need to be wrapped in a `lazy_static!` macro call at all but can be replaced with true `const` slices (or `const` sized arrays, but those are slightly more tedious to construct).

[motivation in winit]: https://github.com/rust-windowing/winit/pull/2313
[proposed for inclusion in `std`]: https://github.com/rust-lang/rust/issues/74465
[the `once_cell` F.A.Q.]: https://docs.rs/once_cell/latest/once_cell/#faq

### DIFF
--- a/sentry-backtrace/src/trim.rs
+++ b/sentry-backtrace/src/trim.rs
@@ -2,31 +2,29 @@ use sentry_core::protocol::{Frame, Stacktrace};
 
 use crate::utils::function_starts_with;
 
-lazy_static::lazy_static! {
-    static ref WELL_KNOWN_SYS_MODULES: Vec<&'static str> = vec![
-        "std::",
-        "core::",
-        "alloc::",
-        "backtrace::",
-        "sentry::",
-        "sentry_core::",
-        "sentry_types::",
-        // these are not modules but things like __rust_maybe_catch_panic
-        "__rust_",
-        "___rust_",
-        // these are well-known library frames
-        "anyhow::",
-        "log::",
-    ];
+const WELL_KNOWN_SYS_MODULES: &[&str] = &[
+    "std::",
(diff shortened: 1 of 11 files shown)


sentry-backtrace/src/trim.rs
```diff
@@ -2,31 +2,29 @@ use sentry_core::protocol::{Frame, Stacktrace};
...
-        "sentry_core::",
-        "sentry_types::",
-        // these are not modules but things like __rust_maybe_catch_panic
-        "__rust_",
-        "___rust_",
-        // these are well-known library frames
-        "anyhow::",
-        "log::",
-    ];
+const WELL_KNOWN_SYS_MODULES: &[&str] = &[
```
Swatinem, line 5: nice! this could have been done independently anyway. No reason for these being `Vec`s

Swatinem (approved):
lgtm, thanks!

I have been following this for quite a while, but rather waiting for it to stabilize in std.

MarijnS95:
> I have been following this for quite a while, but rather waiting for it to stabilize in std.

Yeah, not sure when it lands but this brings us a step closer at least :)

PR #472, opened 2022-06-17
ci: Bump MSRV to 1.57

kamilogorek:
I has been released on `02 Dec 2021`, which is >6 months, which is our support policy.


loewenheim (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.27.0
```
Swatinem, line 3: I think this should have been `Unreleased` but doesn’t matter.

Swatinem (commented):
Two other interesting things we could do after this MSRV bump:
- Add the `rust-version` field to the manifest, which was added in 56: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field
- As well as update to edition 2021, which was also added in 56 🎉

CHANGELOG.md
kamilogorek: Fixed in envelope PR

PR #477, opened 2022-06-23
ref(session): Change session update logic to follow the spec

kamilogorek:
This changes the ordering of filters applied to the event, as well as when a session update happens. This is done in order to follow our unified spec:

```
### Session update filtering

Events may be dropped by our filtering mechanisms (sample rate, beforeSend, event processors or ignored exception types). Only events dropped due to sampling should update the session despite being dropped as we assume the event was dropped to save quota but would have been something the developer cares about. Events dropped due to other reasons should not update the session as we assume they are more likely to be dropped because the developer chooses to ignore them.

#### Filter order

The python SDK shall serve as a reference here. The order for filtering error events is:
1. Check for ignored exception types (a.k.a `ignore_errors`)
3. Apply scoped `event_processor` (a.k.a `error_processor`)
4. Apply global `event_processor`
2. Apply `before_send`
5. Update the session if an event made it this far
6. Apply sampling rate
```

Closes https://github.com/getsentry/sentry-rust/issues/461

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -158,14 +158,6 @@ impl Client {
         mut event: Event<'static>,
         scope: Option<&Scope>,
     ) -> Option<Event<'static>> {
-        if let Some(scope) = scope {
-            scope.update_session_from_event(&event);
-        }
-
-        if !self.sample_should_send() {
-            return None;
-        }
-
         // event_id and sdk_info are set before the processors run so that the
         // processors can poke around in that data.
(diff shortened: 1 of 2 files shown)


Swatinem (approved):
the clarifications are good, and do make sense, though applying all the processors in the case of sampled events is a bit of wasted work, but meh.

kamilogorek:
> though applying all the processors in the case of sampled events is a bit of wasted work

Agree, but spec is spec :(

PR #479, opened 2022-06-27
feat(profiling): Add support for Profiling

viglia:
This PR adds support for the profiling feature.

For this first iteration we will stick to the design implementation of the mobile profilers (`Android`/`iOS`) with one profile being linked to one transaction.

* adds `enable_profiling` option to enable/disable profiling
* adds `profiles_sample_rate` option to further sample profiles with respect to the transactions. This is dependent on `traces_sample_rate` and the probability of a profile being collected is: `traces_sample_rate * profiles_sample_rate` 
* only one profile at a time can be collected. If the profiler is running, no other profiles will be collected when other transactions start

### DIFF
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -25,7 +25,8 @@ client = ["rand"]
 # I would love to just have a `log` feature, but this is used inside a macro,
 # and macros actually expand features (and extern crate) where they are used!
 debug-logs = ["log_"]
-test = ["client"]
+test = ["client", "profiling"]
+profiling = ["pprof", "lazy_static", "mut_static", "build_id", "uuid", "sys-info", "findshlibs"]
 
 [dependencies]
 log_ = { package = "log", version = "0.4.8", optional = true, features = ["std"] }
@@ -33,7 +34,14 @@ once_cell = "1"
 rand = { version = "0.8.1", optional = true }
 sentry-types = { version = "0.27.0", path = "../sentry-types" }
 serde = { version = "1.0.104", features = ["derive"] }
-serde_json = "1.0.46"
+serde_json = { version = "1.0.46" }
+lazy_static = { version = "1.4.0", optional = true }
+mut_static = { version = "5.0.0", optional = true }
+pprof = { version = "0.10.0", optional = true }
+uuid = { version = "1.0.0", features = ["v4", "serde"], optional = true }
+sys-info = { version = "0.9.1", optional = true }
+build_id = { version = "0.2.1", optional = true }
+findshlibs = { version = "=0.10.2", optional = true }
 
 [dev-dependencies]
 # Because we re-export all the public API in `sentry`, we actually run all the
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -25,8 +25,8 @@ client = ["rand"]
 # I would love to just have a `log` feature, but this is used inside a macro,
 # and macros actually expand features (and extern crate) where they are used!
 debug-logs = ["log_"]
-test = ["client", "profiling"]
-profiling = ["pprof", "lazy_static", "mut_static", "build_id", "uuid", "sys-info", "findshlibs"]
+test = ["client"]
+profiling = ["pprof", "build_id", "uuid", "sys-info", "findshlibs"]
 
 [dependencies]
 log_ = { package = "log", version = "0.4.8", optional = true, features = ["std"] }
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -35,9 +35,6 @@ rand = { version = "0.8.1", optional = true }
 sentry-types = { version = "0.27.0", path = "../sentry-types" }
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = { version = "1.0.46" }
-lazy_static = { version = "1.4.0", optional = true }
-mut_static = { version = "5.0.0", optional = true }
-#pprof = { version = "0.10.0", optional = true }
 uuid = { version = "1.0.0", features = ["v4", "serde"], optional = true }
 sys-info = { version = "0.9.1", optional = true }
 build_id = { version = "0.2.1", optional = true }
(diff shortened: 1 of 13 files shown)


sentry-types/src/protocol/profile.rs
```diff
@@ -0,0 +1,110 @@
+use super::v7::{DebugMeta, TraceId};
+use serde::{Deserialize, Serialize, Serializer};
+use uuid::Uuid;
+
+fn serialize_id<S: Serializer>(uuid: &Uuid, serializer: S) -> Result<S::Ok, S::Error> {
+    serializer.serialize_some(&uuid.as_simple().to_string())
```
Swatinem, line 6:
```suggestion
    serializer.serialize_some(&uuid.as_simple())
```

`Simple` implements `Serialize`: https://docs.rs/uuid/latest/uuid/fmt/struct.Simple.html#impl-Serialize

sentry-types/src/protocol/envelope.rs
```diff
@@ -282,13 +294,19 @@ impl Envelope {
                     writeln!(writer)?;
                     continue;
                 }
+                EnvelopeItem::Profile(profile) => {
+                    profile.to_writer(&mut writer)?;
+                    writeln!(writer)?;
+                    continue;
```
Swatinem, line 300: If you don’t need the extra header properties (which I think you don’t), you can use the same code here and below:

`serde_json::to_writer`, combined with the match down below for the `item_type`.

sentry-core/src/performance.rs
```diff
@@ -594,6 +702,163 @@ fn parse_sentry_trace(header: &str) -> Option<SentryTrace> {
...
+        start_time_secs: rep
+            .timing
+            .start_time
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_secs(),
+        duration_nanos: rep.timing.duration.as_nanos() as u64,
+        samples,
+    };
+    use std::time::SystemTime;
```
Swatinem, line 839: interesting, I didn’t know you could put the `use` after the code that actually uses this.

sentry-core/src/performance.rs
```diff
@@ -273,6 +300,14 @@ pub(crate) struct TransactionInner {
     pub(crate) transaction: Option<protocol::Transaction<'static>>,
 }
 
+#[derive(Default)]
+#[cfg(all(feature = "profiling", not(target_os = "windows")))]
+struct ProfileInner {
+    transaction_id: String,
+    #[cfg(all(feature = "profiling", not(target_os = "windows")))]
+    profiler_guard: Option<pprof::ProfilerGuard<'static>>,
+}
```
Swatinem, line 309: The whole struct is already `cfg`-guarded, so no need to have the member as well.

Also, I would rather make the whole struct an `Option`, instead of an `Option` member and an empty string for the transaction.

In general, I would remove the mutable global altogether, and rather attach the `Option<ProfilesGuard>` onto the transaction. That way, you wouldn’t have to take the lock and do a transaction_id comparison on *every* transaction finish.

I would still have one global `AtomicBool` that says `PROFILER_RUNNING`.

You should be able to use a [compare_exchange](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html#method.compare_exchange)(false, true), then try to start your profiler. If that fails for some reason, just `store(false)` back into it.
Similarly, to stop profiling, you will only work on the `Transaction` itself, and if you *do* have an associated `ProfilerGuard`, you can stop it and `store(false)` as well. That way, you would only have a single CAS instead of potentially longer running locks.

sentry-core/Cargo.toml
```diff
@@ -25,15 +25,26 @@ client = ["rand"]
 # I would love to just have a `log` feature, but this is used inside a macro,
 # and macros actually expand features (and extern crate) where they are used!
 debug-logs = ["log_"]
-test = ["client"]
+test = ["client", "profiling"]
```
Swatinem, line 28: The `test` feature should not pull in `profiling`.

Swatinem (commented):
This looks promising. I scattered some suggestions around. Essentially I would love this to move into its own module, and have a `fn start_profiling() -> Option<ProfileGuard>`, and maybe a `fn stop_profiling(ProfileGuard) -> Profile (the envelope item)` as the two functions that you call from the `performance` module.

sentry-core/src/performance.rs
viglia: > The whole struct is already cfg-guarded, so no need to have the member as well

yep, I definitely forgot to remove it when I moved it at the struct level.

viglia: > In general, I would remove the mutable global altogether, and rather attach the Option<ProfilesGuard> onto the transaction. That way, you wouldn’t have to take the lock and do a transaction_id comparison on every transaction finish

I would love such implementation!! 
unfortunately, the `build` [method](https://docs.rs/pprof/latest/pprof/struct.ProfilerGuardBuilder.html#method.build) of `ProfilerGuardBuilder` returns a `ProfilerGuard<'static>`, which does not implement the `clone` trait and so I cannot add it as an Option field in the `Transaction` struct defined in `sentry-core/performance.rs` since `Transaction` derives `Clone`.

sentry-core/src/performance.rs
viglia: Right. This is a bit confusing. I'll move it to the top of the function `body`.

sentry-types/src/protocol/profile.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// The ID of the transaction this profile belongs to
+    pub transaction_id: Uuid,
+    /// The ID of the event
+    #[serde(serialize_with = "serialize_id")]
+    pub profile_id: Uuid,
+    /// Represents the profile collected
+    pub sampled_profile: SampledProfile,
+    #[serde(rename = "device_os_name")]
+    /// OS name
+    pub os_name: String,
```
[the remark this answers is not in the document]
viglia: yes, those are all required, except for `architecture` that I will change to Option as you suggest.

sentry-core/src/profiling.rs
```diff
@@ -0,0 +1,239 @@
...
+    for sample in rep.data.keys() {
+        let frames = sample
+            .frames
+            .iter()
+            .map(|frame| RustFrame {
+                instruction_addr: format!("{:p}", frame.ip()),
+            })
+            .collect();
+
+        samples.push(Sample {
```
Swatinem, line 190: You can also use a `rep.data.keys().map(…).collect()` instead of the `Vec::new` + `push`.

Swatinem (approved):

PR #481, opened 2022-07-08
fix(profiling): Add support to use frame-pointer on macos

viglia:
This solves a problem (probably due to unwind) on macos, where the profiler would
yield a `Trace/BPT trap`.

Instead of using the `backtrace` crate, when running on `macos`, using `frame-pointer`
is more recommended to get the backtrace.


Swatinem (commented):
I would additionally gate this behind a `aarch64` check. frame pointer omission on x64 macOS is still very common so trying to do frame pointer unwinding on x64 would give you wrong results or even worse crash your program.

PR #490, opened 2022-07-29
feat(profiling): Drop the profile when no sample were collected

viglia:
This check is already happening in Relay.

The idea is drop it already on the client side in order to avoid waste of resources on Relay.


Swatinem (approved):

Swatinem:
The formatting is all wrong though, not sure why rustfmt is not picking that up?

Swatinem:
Well, rustfmt probably can’t format anything inside of that pesky `with_client_impl!` macro :-(

PR #496, opened 2022-08-22
doc: add few verbs and punctuation

olksdr:
Adding few verbs and punctuations makes the reading the description a bit clearer.


Swatinem (approved):

olksdr:
Looks like some dependencies issues:
```
error: package `time v0.3.14` cannot be built because it requires rustc 1.59.0 or newer, while the currently active rustc version is 1.57.0
make: *** [check-no-default-features] Error 101
Error: Process completed with exit code 2.
```
but I do not see where this version is coming from. Locally Cargo.lock contains only following: 
```
 "time 0.2.27",
 "time 0.3.13",
```

@Swatinem  is there anything I can do here to help it to compile? 😄

Swatinem:
We have to bump the MSRV: #498

PR #497, opened 2022-08-30
feat(profiling): Use frame-pointer for stack unwinding and update to pprof 0.10.1

viglia:
Once we'll create a doc section for profiling we'll have to make clear that on Linux this requires the following extra steps:

1. recompiling build-std : `cargo +nightly run -Z build-std --target x86_64-unknown-linux-gnu`
2. force frame-pointers with rustflags: `RUSTFLAGS="-C force-frame-pointers=yes" cargo build --release`


Swatinem:
I’m still a bit hesitant to do this, because:

1. requiring people to use a nightly compiler and `build-std` is a bit overwhelming
2. you still depend on system libraries like `libpthread` to be compiled with frame pointers, which for a lot of linux distros is not the case. :-(

We can alternatively just expose this as another feature flag that just forwards the `frame-pointer` flag to `pprof`. That way the user still has control over the exact unwinding mechanism used.

viglia:
> We can alternatively just expose this as another feature flag that just forwards the frame-pointer flag to pprof. That way the user still has control over the exact unwinding mechanism used.

Yes, I can add a feature flag `frame-pointer` to forward to `pprof` this way we're not "forcing" people to use frame-pointers.

On linux though, as we have seen, using the default approach with `libunwind` will just produce bad results.

Swatinem (approved):
lgtm.

We probably need to bump MSRV because of our dependencies anyway. However the weak dependencies feature is quite new, not sure when we can actually start using that.

Swatinem (approved):

PR #504, opened 2022-09-21
feat(profiling): Update to new standard Profile format

viglia:
This PR implements the new standard format for a Profile that will be used across all the different SDKs.

Relay related PR: [#1462](https://github.com/getsentry/relay/pull/1462)

### DIFF
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -26,7 +26,7 @@ client = ["rand"]
 # and macros actually expand features (and extern crate) where they are used!
 debug-logs = ["log_"]
 test = ["client"]
-profiling = ["pprof", "build_id", "uuid", "sys-info", "findshlibs"]
+profiling = ["pprof", "build_id", "uuid", "sys-info", "findshlibs", "rustc_version_runtime", "libc"]
 frame-pointer = ["pprof?/frame-pointer"]
 
 [dependencies]
@@ -40,9 +40,12 @@ uuid = { version = "1.0.0", features = ["v4", "serde"], optional = true }
 sys-info = { version = "0.9.1", optional = true }
 build_id = { version = "0.2.1", optional = true }
 findshlibs = { version = "=0.10.2", optional = true }
+rustc_version_runtime = { version = "0.2.1", optional = true }
+chrono = { version = "0.4.22",  features = ["serde"], optional = true}
 
 [target.'cfg(target_family = "unix")'.dependencies]
 pprof = { version = "0.10.1", optional = true, default-features = false}
+libc = { version = "^0.2.66", optional = true }
 
 [dev-dependencies]
 # Because we re-export all the public API in `sentry`, we actually run all the
--- a/sentry-core/src/profiling.rs
+++ b/sentry-core/src/profiling.rs
@@ -1,12 +1,15 @@
+use std::collections::HashMap;
 use std::fmt;
 use std::sync::atomic::{AtomicBool, Ordering};
 
+use chrono::DateTime;
 use findshlibs::{SharedLibrary, SharedLibraryId, TargetSharedLibrary, TARGET_SUPPORTED};
 
 use sentry_types::protocol::v7::Profile;
 use sentry_types::protocol::v7::{
-    DebugImage, DebugMeta, RustFrame, Sample, SampledProfile, SymbolicDebugImage, TraceId,
-    Transaction,
+    DebugImage, DebugMeta, DeviceMetadata, OSMetadata, RuntimeMetadata, RustFrame, Sample,
+    SampleProfile, SymbolicDebugImage, ThreadMetadata, TraceId, Transaction, TransactionMetadata,
+    Version,
 };
 use sentry_types::{CodeId, DebugId, Uuid};
 
@@ -60,14 +63,9 @@ pub(crate) fn finish_profiling(
     transaction: &Transaction,
     profiler_guard: ProfilerGuard,
     trace_id: TraceId,
-) -> Option<Profile> {
-    let profile = match profiler_guard.0.report().build_unresolved() {
-        Ok(report) => Some(get_profile_from_report(
-            &report,
-            trace_id,
-            transaction.event_id,
-            transaction.name.as_ref().unwrap().clone(),
-        )),
+) -> Option<SampleProfile> {
+    let sample_profile = match profiler_guard.0.report().build_unresolved() {
+        Ok(report) => Some(get_profile_from_report(&report, trace_id, transaction)),
         Err(err) => {
             sentry_debug!(
                 "could not build the profile result due to the error: {}",
@@ -78,7 +76,7 @@ pub(crate) fn finish_profiling(
     };
 
     PROFILER_RUNNING.store(false, Ordering::SeqCst);
-    profile
+    sample_profile
 }
 
 /// Converts an ELF object identifier into a `DebugId`.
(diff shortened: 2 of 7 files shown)


sentry-types/src/protocol/v7.rs
```diff
@@ -1894,6 +1894,11 @@ impl fmt::Display for SpanStatus {
     }
 }
 
+#[cfg(all(feature = "profiling", target_family = "unix"))]
+fn is_zero(number: u64) -> bool {
```
Swatinem, line 1898: this seems to be unused

sentry-types/Cargo.toml
```diff
@@ -30,3 +30,4 @@ thiserror = "1.0.15"
 time = { version = "0.3.5", features = ["formatting", "parsing"] }
 url = { version = "2.1.1", features = ["serde"] }
 uuid = { version = "1.0.0", features = ["v4", "serde"] }
+chrono = { version = "0.4.22",  features = ["serde"] }
```
Swatinem, line 33: I wouldn’t really want to add back chrono.
We removed that in https://github.com/getsentry/sentry-rust/pull/409 as it was unmaintained for a long time. I think they found new maintainers in the meantime, but we should still get along fine with just `time`.

sentry-types/src/protocol/profile.rs
```diff
@@ -1,77 +1,143 @@
...
+pub struct TransactionMetadata {
+    #[serde(serialize_with = "serialize_id")]
+    /// Transaction ID
+    pub id: Uuid,
+    /// Transaction Name
+    pub name: String,
+    /// Trace ID
+    pub trace_id: TraceId,
+    /// Transaction start timestamp in nanoseconds
+    pub relative_start_ns: u64,
```
Swatinem, line 23: relative to what?

sentry-types/src/protocol/profile.rs
```diff
@@ -1,77 +1,143 @@
...
-    pub thread_name: String,
-    /// The thread id
+    /// ID of the relative stack
+    pub stack_id: u32,
+    /// Thread ID
     pub thread_id: u64,
-    /// Nanoseconds elapsed between when the profiler started and when this sample was collected
-    pub nanos_relative_to_start: u64,
+    /// Timestamp at which this sample was collected
+    pub relative_timestamp_ns: u64,
```
Swatinem, line 46: relative to what? the start of the profiling session, or relative to the transaction start?

sentry-types/src/protocol/profile.rs
```diff
@@ -1,77 +1,143 @@
...
+    /// Event ID or Profile ID
+    pub event_id: Uuid,
+    /// Platform
+    pub platform: String,
+    /// Collected profile
+    pub profile: Profile,
+    /// Release
+    pub release: String,
+    /// Timestamp at which the profiler started
+    pub timestamp: DateTime<Utc>,
```
Swatinem, line 138: This would be `SystemTime` then.

sentry-types/src/protocol/profile.rs
```diff
@@ -1,77 +1,143 @@
...
-    /// A string describing the architecture of the CPU that is currently in use
-    /// <https://doc.rust-lang.org/std/env/consts/constant.ARCH.html>
+    /// list of samples in a profile
+    pub samples: Vec<Sample>,
+    /// List of stacks: each stacks is a vec of indexed frames
+    pub stacks: Vec<Vec<u32>>,
+    /// List of frames
+    pub frames: Vec<RustFrame>,
+    /// Thread metadata
+    pub thread_metadata: HashMap<String, ThreadMetadata>,
```
Swatinem, line 67: what is the key here exactly? I wonder why it is not indexed using a `u32` similar to how stacks/frames works?

sentry-core/src/profiling.rs
```diff
@@ -171,72 +169,104 @@ pub fn debug_images() -> Vec<DebugImage> {
...
+                *address_to_frame_id
+                    .entry(instruction_addr.clone())
+                    .or_insert_with(|| {
+                        frames.push(RustFrame { instruction_addr });
+                        (frames.len() - 1) as u32
+                    })
             })
             .collect();
 
+        stacks.push(stack);
```
Swatinem, line 200: shouldn’t we also dedupe these?

sentry-core/src/profiling.rs
```diff
@@ -171,72 +169,104 @@ pub fn debug_images() -> Vec<DebugImage> {
...
+                let instruction_addr = format!("{:p}", frame.ip as *mut core::ffi::c_void);
                 #[cfg(not(feature = "frame-pointer"))]
-                instruction_addr: format!("{:p}", frame.ip()),
+                let instruction_addr = format!("{:p}", frame.ip());
+                *address_to_frame_id
+                    .entry(instruction_addr.clone())
+                    .or_insert_with(|| {
+                        frames.push(RustFrame { instruction_addr });
+                        (frames.len() - 1) as u32
+                    })
```
Swatinem, line 196: this calls for some `IndexSet` ;-)
I can pair on this to help with it.

sentry-core/src/profiling.rs
```diff
@@ -171,72 +169,104 @@ pub fn debug_images() -> Vec<DebugImage> {
...
-        version_code: build_id::get().to_simple().to_string(),
-    };
 
-    profile
+        event_id: uuid::Uuid::new_v4(), // double check this
+        release: format!(
+            "{} ({})",
+            env!("CARGO_PKG_VERSION"),
+            build_id::get().to_simple(),
+        ),
```
Swatinem, line 249: I believe `CARGO_PKG_VERSION` would be the version of the `sentry-core` crate in this case. Is that intentional or should this rather use the release as configured?

sentry-types/Cargo.toml
viglia: fair enough, but I'd need a way to connvert `SystemTime` to `ISO 8601` string representation to comply to what the backend expects (see example [here](https://github.com/getsentry/relay/pull/1462/files#diff-aacb2c0a29425780822d9231b633ab596c0a141611568e91de64955a8df0b86cR16)) and I'm not sure there is a way to do that with `time`.

ps.  checking at their Github release page I can see they started releasing new versions and fixing stuff from July onward.

sentry-types/src/protocol/profile.rs
viglia: good point, let me add that in the comments!

sentry-types/src/protocol/profile.rs
viglia: as above: good point, let me add that in the comments!

sentry-types/src/protocol/profile.rs
viglia: Why so? 

In the backend it expects something like this `"timestamp": "2022-09-01T09:45:00.000Z"` so [this](https://docs.rs/chrono/0.4.0/chrono/struct.DateTime.html) looked like the way to go (unless we are discarding _chrono_ as discussed in the comments above)

Swatinem: The relevant fields in `sentry-types` have custom serde serializers that are taking care of this, for example `#[serde(default = "SystemTime::now", with = "ts_rfc3339")]`. There is `sentry_types::utils::ts_rfc3339` that does this.

sentry-types/src/protocol/profile.rs
viglia: That's because json keys can only be strings in an object. Otherwise yes, a `u32` would have made more sense.

sentry-core/Cargo.toml
```diff
@@ -40,9 +40,12 @@ uuid = { version = "1.0.0", features = ["v4", "serde"], optional = true }
 sys-info = { version = "0.9.1", optional = true }
 build_id = { version = "0.2.1", optional = true }
 findshlibs = { version = "=0.10.2", optional = true }
+rustc_version_runtime = { version = "0.2.1", optional = true }
+chrono = { version = "0.4.22",  features = ["serde"], optional = true}
 
 [target.'cfg(target_family = "unix")'.dependencies]
 pprof = { version = "0.10.1", optional = true, default-features = false}
+libc = { version = "^0.2.66", optional = true }
```
audkar, line 48: (I am novice to Rust)

But if Rust is similar to C, then`libc` dependency is bad for portability. Consumers of such transitive dependency have to start thinking with which `libc` version executable was linked during compilation.

We are trying to avoid any Linux native sys dependencies and use Rust native alternatives (eg. `native-tls` -> `rustls`, etc). 

Maybe there are a Rust alternative to `libc` usage? If I understand it is used only to get threadId?

sentry-core/src/profiling.rs
viglia: we're moving to sentry SDK release. Good 👍🏻

sentry-core/Cargo.toml
audkar: > if you have a specific alternative in mind though, I'm happy to reconsider this.

No. Sadly I don't.

sentry-core/src/profiling.rs
```diff
@@ -188,12 +187,15 @@ fn get_profile_from_report(
...
-                    })
+                let rust_frame = RustFrame { instruction_addr };
+
+                address_to_frame_idx
+                    .get_index_of(&rust_frame)
+                    .unwrap_or_else(|| -> usize {
+                        address_to_frame_idx.insert(rust_frame);
+
+                        address_to_frame_idx.len() - 1
+                    }) as u32
```
Swatinem, line 198: This can be as simple as `address_to_frame_idx.insert_full(rust_frame).0 as u32`.

Also, I would recommend to just save the raw `const *()` in the `IndexSet`. that would avoid allocating a string for all of these, and would also likely speed up the `IndexSet` operations as well. Once you write them out into the report, thats the time you can `format!()` stuff.

sentry-core/src/profiling.rs
```diff
@@ -241,13 +243,12 @@ fn get_profile_from_report(
...
-            "{} ({})",
-            env!("CARGO_PKG_VERSION"),
-            build_id::get().to_simple(),
-        ),
-        timestamp: DateTime::from(rep.timing.start_time),
+        event_id: uuid::Uuid::new_v4(),
+        release: transaction
+            .release
+            .as_ref()
+            .map_or("".to_string(), |r| -> String { r.to_string() }),
```
Swatinem, line 250: maybe `transaction.release.unwrap_or_default().into()` would be easier to read here.

viglia: If I use that I get this error

```
cannot move out of `transaction.release` which is behind a shared reference
move occurs because `transaction.release` has type `std::option::Option<std::borrow::Cow<'_, str>>`, which does not implement the `Copy` trait
```

So I'd have to use a `clone`. Not sure what would be better frankly.

Swatinem: ah yes, `clone()` absolutely. you cant an allocated `String` anyway.

Swatinem (approved):

sentry-types/Cargo.toml
LeoniePhiline: RUSTSEC-2020-0071 matches again now. https://github.com/getsentry/sentry-rust/issues/371 should be re-opened.

PR #508, opened 2022-10-17
feat(transport): Add option to disable SSL verification

olksdr:
Adding this option will allow to use self-signed certificates in the user's self-hosted environments.

Using this option also allows to use any types of certs, even expired which can introduce vulnerabilities, since all the certificates are trusted by default. That's why we must make clear that it's a last resort if there is no possibility to use legitimate certificate (e.g. Let's Encrypt or similar).

related https://github.com/getsentry/relay/issues/1499


sentry/src/transports/surf.rs
```diff
@@ -28,7 +32,20 @@ impl SurfHttpTransport {
     }
 
     fn new_internal(options: &ClientOptions, client: Option<SurfClient>) -> Self {
-        let client = client.unwrap_or_else(SurfClient::new);
+        let client = if let Some(client) = client {
+            client
+        } else {
+            let mut http_client = http_client::isahc::IsahcClient::new();
```
Swatinem, line 38: if/else might be better here I think

Swatinem (approved):

PR #510, opened 2022-11-03
sentry-core: add traces_sampler client option

tommilligan:
This PR adds the `traces_sampler` option to `ClientOptions`, and respects it when making transaction sampling decisions as detailed in the SDK guidelines: https://develop.sentry.dev/sdk/performance/#sdk-configuration

There was some prior work on this topic in https://github.com/getsentry/sentry-rust/pull/288, however by now it's pretty stale so I started fresh.

Points for review:
- To implement `Debug` for this option I just print the function's memory address (if set); I can't think of something simple that would have more value to the user. [link](https://doc.rust-lang.org/std/ptr/macro.addr_of.html)
- If accepted, I would like to update the official docs with the new option, could someone point me in the right direction for that?
- Adding a new field to `ClientOptions` is a breaking change - I presume this necessitates a minor version bump for release?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Breaking Changes**:
 
 - The minimum supported Rust version was bumped to **1.60.0** due to requirements from dependencies. ([#498](https://github.com/getsentry/sentry-rust/pull/498))
+- Added the `traces_sampler` option to `ClientOptions`. This allows the user to customise sampling rates on a per-transaction basis. ([#510](https://github.com/getsentry/sentry-rust/pull/510))
 
 **Features**:
 
--- a/sentry-contexts/src/utils.rs
+++ b/sentry-contexts/src/utils.rs
@@ -26,7 +26,7 @@ mod model_support {
                 return None;
             }
 
-            let mut buf = vec![0u8; size as usize];
+            let mut buf = vec![0u8; size];
             let res = libc::sysctlbyname(
                 c_name.as_ptr() as _,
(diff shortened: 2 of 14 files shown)


tommilligan:
> > Adding a new field to `ClientOptions` is a breaking change - I presume this necessitates a minor version bump for release?
> 
> According to [this guideline](https://doc.rust-lang.org/cargo/reference/semver.html#struct-add-public-field-when-no-private), it even requires a major bump because you're adding a new public field, but no private field currently exists.

Right, agreed it's a breaking change - but all the `sentry*` packages are currently sub-v1, so I think a minor version bump is what's needed. I'll add a note in the changelog of the breakage anyway.

I don't like that adding a new config option breaks this for all users, but I think that's outside the scope of this PR.

loewenheim (approved):

Turbo87:
I've tried to make use of this new API, but I fail to understand how this is supposed to work. All fields on the `TransactionContext` struct are private, so even if the function gets passed such a struct there is still no information available to determine the sample rate for the transaction. Am I missing something? 🤔

tommilligan:
Hi - yes, agreed. This was actually merged by a maintainer before I'd fully tested it, so sorry for the incomplete feature.

I've opened a follow up PR already which supports inspection of the transaction context, plus adding arbitrary user data as supported in e.g. the Python SDK. Comments welcome over there! https://github.com/getsentry/sentry-rust/pull/512

This isn't a breaking change so should be able to be added with a patch release fairly quickly once merged.

Issue #520, opened 2022-11-21
Add support for "v0" (non-legacy) Rust symbol mangling


vorporeal:
Rust is changing their symbol mangling strategy; the new version is currently available via a compiler flag: `-C symbol-mangling-version=v0`.

See rust tracking issue here: https://github.com/rust-lang/rust/issues/60705

We'd like to use this new symbol mangling format, but Sentry doesn't seem to quite demangle these properly:

![image](https://user-images.githubusercontent.com/211194/203171543-11736a63-399c-48e5-b335-af828d48c1db.png)

Note in the above image that Sentry doesn't know which crate the stack frame is from, and is failing to strip (what appears to be) some hash from the crate name.

PR #525, opened 2022-11-28
fix: Support process v0 symbols in stack traces

Swatinem:
This will now correctly strip the crate hash from new v0 mangled symbols.

We may need to do the same in sentry itself when cleaning symbols that come from serverside symbolication.

fixes https://github.com/getsentry/sentry-rust/issues/520

### DIFF
--- a/sentry-backtrace/src/utils.rs
+++ b/sentry-backtrace/src/utils.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use once_cell::sync::Lazy;
 use regex::{Captures, Regex};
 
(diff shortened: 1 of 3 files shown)


sentry-backtrace/src/utils.rs
```diff
@@ -10,14 +12,23 @@ static HASH_FUNC_RE: Lazy<Regex> = Lazy::new(|| {
...
+
 static CRATE_RE: Lazy<Regex> = Lazy::new(|| {
     Regex::new(
         r#"(?x)
         ^
         (?:_?<)?           # trait impl syntax
         (?:\w+\ as \ )?    # anonymous implementor
         ([a-zA-Z0-9_]+?)   # crate name
-        (?:\.\.|::)        # crate delimiter (.. or ::)
+        (?:\.\.|::|\[)     # crate delimiter (.. or ::)
```
kamilogorek, line 31:
```suggestion
        (?:\.\.|::|\[)     # crate delimiter (.. or :: or [)
```

kamilogorek (approved):

PR #528, opened 2022-11-29
ref: Update CI Workflows

Swatinem:
- Update actions versions.
- Switch to cargo-llvm-cov.


Swatinem:
The "regression" in code coverage is primarily because we are not considering doctests for coverage anymore. Sentry-rust relies on doctests a lot compared to straight unit/integration tests.

kamilogorek (approved):

PR #533, opened 2023-01-06
sentry-core: make TransactionContext.trace_id readable

tommilligan:
I would like to add the Sentry trace_id to our internal logs to be able to correlate it with the sentry transaction. Currently this isn't publicly accessible anywhere.

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -180,6 +180,11 @@ impl TransactionContext {
         &self.op
     }
 
+    /// Get the Trace ID of this Transaction.
+    pub fn trace_id(&self) -> protocol::TraceId {
+        self.trace_id
+    }
+
     /// Get the custom context of this Transaction.
     pub fn custom(&self) -> Option<&CustomTransactionContext> {
         self.custom.as_ref()


kamilogorek (approved):
This is already possible in JS SDK, so I don't see a reason why it shouldn't be here.

tommilligan:
> This is already possible in JS SDK, so I don't see a reason why it shouldn't be here.

Agreed, thanks. I already access `trace_id` as publicly accessible in the Python SDK, so figured it would be fine to have here too.

Note that this PR only exposes it on the `TransactionContext`, I think it could be exposed on `Transaction` as well, but that looked like a bit more effort/design required.

kamilogorek:
> Note that this PR only exposes it on the TransactionContext, I think it could be exposed on Transaction as well, but that looked like a bit more effort/design required.

cc @Swatinem

Swatinem:
IMO its a good idea to also expose that on the `Transaction`, `Span` and `TransactionOrSpan` types.

tommilligan:
> IMO its a good idea to also expose that on the `Transaction`, `Span` and `TransactionOrSpan` types.

I'm happy to pick this up as a subsequent PR. The simplest option would be to expose the `TraceContext` using a getter, is it acceptable to expose the inner protocol structs in that way? https://github.com/getsentry/sentry-rust/blob/5013eb1fdca505ef9c7c28266afcf847da7c17b3/sentry-types/src/protocol/v7.rs#L1439

Alternatively I could look at exposing the fields individually, which would be quite boilerplate heavy.

tommilligan:
Opened a separate PR for exposing on the `Transaction`/`Span` types here: https://github.com/getsentry/sentry-rust/pull/534

Swatinem (approved):

tommilligan:
Is there a scheduled release cadence for the sentry family of crates? Would like to know when a release with this in is likely to be cut

Swatinem:
We don’t have a fixed cadence, rather adhoc. I triggered the automated release which needs to be approved and should be out shortly: https://github.com/getsentry/publish/issues/1766

tommilligan:
Amazing, thank you

PR #534, opened 2023-01-12
sentry-core: make TraceContext publicly readable

tommilligan:
Raised in review of https://github.com/getsentry/sentry-rust/pull/533#issuecomment-1375551161

This PR exposes the `TraceContext` stored on a `Transaction` or `Span`.

All data on the `TraceContext` is publicly readable in other SDKs, such as Python or Javascript.

- Rust struct: https://github.com/getsentry/sentry-rust/blob/5013eb1fdca505ef9c7c28266afcf847da7c17b3/sentry-types/src/protocol/v7.rs#L1439
- Python analagous class: https://github.com/getsentry/sentry-python/blob/d0eed0ee828684f22fe2a2b28b02cf7f4ce8c74a/sentry_sdk/tracing.py#L81

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -62,7 +62,7 @@ pub type CustomTransactionContext = serde_json::Map<String, serde_json::Value>;
 ///
 /// The Transaction Context defines the metadata for a Performance Monitoring
 /// Transaction, and also the connection point for distributed tracing.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct TransactionContext {
     #[cfg_attr(not(feature = "client"), allow(dead_code))]
     name: String,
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
its a bit unfortunate this is behind a lock so you need to clone :-(

but oh well

tommilligan:
> its a bit unfortunate this is behind a lock so you need to clone :-(
> 
> but oh well

Agreed. I think in future there would be scope for handling this better with something like parking_lot's `MutexGuard::map`, if you're open to changing the implementation of `Mutex`: https://docs.rs/lock_api/latest/lock_api/struct.MutexGuard.html#method.map

But I think for this case, a one-off clone is not too bad.

Swatinem:
I would be hesitant to give out more lock guards, with the recent deadlock I tracked down and fixed quite painfully: #536 

We should rather remove the need for locking in the first place somehow, though I have no good ideas at the moment.

PR #539, opened 2023-01-19
fix actix match name bug

wuerges:
# Issue

Actix `request.match_name()` is buggy. 
It doesn't care about the HTTP verb. It only looks at the `path`. Thus 2 routes with the same `path` will have the same match name. 

# Proposal

Previously, `format!("{} {}", req.method(), req.uri())` was already used as a fallback for the transaction name.
These changes make it as the default transaction name.

### DIFF
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -336,7 +336,8 @@ fn map_status(status: StatusCode) -> protocol::SpanStatus {
 
 /// Extract a transaction name from the HTTP request
 fn transaction_name_from_http(req: &ServiceRequest) -> String {
-    format!("{} {}", req.method(), req.uri())
+    let path_part = req.match_pattern().unwrap_or_else(|| "<none>".to_string());
+    format!("{} {}", req.method(), path_part)
 }
 
 /// Build a Sentry request struct from the HTTP request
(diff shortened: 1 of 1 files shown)


sentry-actix/src/lib.rs
```diff
@@ -336,14 +334,13 @@ fn map_status(status: StatusCode) -> protocol::SpanStatus {
...
-/// Build a Sentry request struct from the HTTP request
-fn sentry_request_from_http(request: &ServiceRequest, with_pii: bool) -> (Option<String>, Request) {
-    let transaction = if let Some(name) = request.match_name() {
-        Some(String::from(name))
-    } else {
-        request.match_pattern()
-    };
+/// Extract a transaction name from the HTTP request
+fn transaction_name_from_http(req: &ServiceRequest) -> String {
+    format!("{} {}", req.method(), req.uri())
```
Swatinem, line 339: I believe `req.uri()` might have too high of a cardinality and might even contain PII.

I believe `match_pattern` is the thing we want, is it documented to give you `/user/{id}/profile` for example: https://docs.rs/actix-web/latest/actix_web/struct.HttpRequest.html#method.match_pattern

`match_name` seems to refer to the function used for logging? Not sure, there is no example.

Adding a `req.method()` prefix does seem like a very good thing indeed.

Swatinem (commented):
NOTE to self: Axum recently got something similar which we might want to support with a feature flag as well:
https://docs.rs/axum/latest/axum/extract/struct.MatchedPath.html

sentry-actix/src/lib.rs
wuerges: Agree 100%.

I've changed the code to use `request.match_pattern()`, fixed the broken tests and added an extra one that does a POST with a pattern.

sentry-actix/src/lib.rs
```diff
@@ -487,7 +531,7 @@ mod tests {
         assert_eq!(events.len(), 2);
         for event in events {
             let request = event.request.expect("Request should be set.");
-            assert_eq!(event.transaction, Some("failing".into())); // Transaction name is the name of the function
+            assert_eq!(event.transaction, Some("GET /test".into())); // Transaction name is the name of the function
```
Swatinem, line 534:
```suggestion
            assert_eq!(event.transaction, Some("GET /test".into())); // Transaction name is the matcher of the route
```

:-)

Swatinem (approved):

PR #554, opened 2023-02-22
fix(sentry-types): Switch to checked version of `from_secs_f64`

olksdr:
This change make use of `try_from_secs_f64` (the checked version of `from_secs_f64`), avoiding panics if the incoming seconds is negative, overflows Duration or not finite.

Only one gotcha here, is that this function was added only in rust `1.66.0`, which is above current MSRV.


Swatinem (approved):

Swatinem:
> Only one gotcha here, is that this function was added only in rust `1.66.0`, which is above current MSRV.

I’m surprised that CI does not complain though, its using `1.60`

kamilogorek:
It was stabilized in 1.66, not added, so I wonder if we have incorrect compilation flags set somewhere

olksdr:
> It was stabilized in 1.66, not added, so I wonder if we have incorrect compilation flags set somewhere

That's what I meant 😅  thanks! 

How do you want to proceed with this? Maybe change the implementation a bit and for now just use previous function with additional checks on top of it to avoid panics? 
There maybe other places where the the call to this function must be changed. 

Since this one caused an incident for us https://github.com/getsentry/relay/pull/1872

kamilogorek:
Welp, MSRV CI was broken - https://github.com/getsentry/sentry-rust/pull/555
Note that this PR https://github.com/getsentry/sentry-rust/pull/549 also broke MSRV and it's now 1.65 😅

olksdr:
How about fix  from [c987539](https://github.com/getsentry/sentry-rust/pull/554/commits/c9875393d2ca62497dc18c30dbe9fc6b017d59e0) which does not use the new function, but just has additional checks in-place to make sure we do not panic.

kamilogorek:
@olksdr feel free to revert it back to your original fix, we bumped MSRV :)

PR #555, opened 2023-02-23
ci: Set default rust toolchain to MSRV for MSRV job

kamilogorek:
- `rustup toolchain install stable` calls are redundant due to `ubuntu-latest` already containins up-to-date stable - https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#rust-tools
- setting `RUSTFLAGS: -Dwarnings` is redundant, as all `make` jobs that are used already set it there
- `component add` uses `stable` by default
- calling `rustup default ${{ matrix.rust }}` for MSRV tests is required, as otherwise `stable` is used


.github/workflows/ci.yml
```diff
@@ -87,6 +87,7 @@ jobs:
         uses: actions/checkout@v3
 
       - run: rustup toolchain install ${{ matrix.rust }} --profile minimal --no-self-update
+      - run: rustup default ${{ matrix.rust }}
```
kamilogorek, line 90:
```suggestion
      - name: Install rust ${{ matrix.rust }} toolchain
        run: |
          rustup toolchain install ${{ matrix.rust }} --profile minimal --no-self-update
          rustup default ${{ matrix.rust }}
```

Swatinem (approved):
well, that explains why MSRV was not really enforced :-(
I believe it was me breaking that at some point…

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v3
 
-      - run: rustup toolchain install stable --profile minimal --component rustfmt --component clippy --no-self-update
```
Swatinem, line 18: We want the *latest* stable version whenever that is available. We don’t want to be at the mercy of whenever github feels like updating their Runners.

.github/workflows/ci.yml
```diff
@@ -76,17 +67,17 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         rust: [1.60.0]
 
-    name: Check / Test MSRV on ${{ matrix.os }}
+    name: Check/Test MSRV (${{ matrix.rust }}) on ${{ matrix.os }}
```
Swatinem, line 70: We removed the explicit version tagging, as its a pain in the rear because we would have to update branch protection rules in a "turn off, merge, then turn back on" fashion.

.github/workflows/ci.yml
```diff
@@ -30,20 +30,14 @@ jobs:
...
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [stable]
 
-    name: Check feature permutations using Rust ${{ matrix.rust }} on ${{ matrix.os }}
+    name: Check feature permutations using Rust stable on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
-    env:
-      RUSTFLAGS: -Dwarnings
```
Swatinem, line 39: While it sure is a pain that we have failing builds on every Rust update, that is quite intentional and we do want to deny warnings.

Swatinem (changes requested):

.github/workflows/ci.yml
kamilogorek: Oh, gotcha

.github/workflows/ci.yml
kamilogorek: > setting RUSTFLAGS: -Dwarnings is redundant, as all make jobs that are used already set it there

.github/workflows/ci.yml
kamilogorek: They have weekly cadence though, so not sure if it's that crucial tbh.
If we want to update it though, we'll need to change the default everywhere as well, as it's tied to version installed in the image.

.github/workflows/ci.yml
kamilogorek, line 70:
```suggestion
    name: Check / Test MSRV on ${{ matrix.os }}
```

.github/workflows/ci.yml
Swatinem: We do have a few steps that are not using the makefile targets

kamilogorek: Iiiii don't belive so? I removed it from workflows that run `make checkall, `make checkfast` and `make testfast`. All of which already set `-Dwarnings`. (`make testfail` doesnt but in this case `checkfast` would fail first)

Swatinem (approved):

PR #562, opened 2023-03-14
feat(profiling): remove a sample if it's the only sample in a thread

viglia:
This closes the following [issue](https://github.com/orgs/getsentry/projects/30/views/7?pane=issue&itemId=18343664)

### DIFF
--- a/sentry-core/src/profiling.rs
+++ b/sentry-core/src/profiling.rs
@@ -185,6 +185,8 @@ fn get_profile_from_report(
     let mut stacks: Vec<Vec<u32>> = Vec::with_capacity(rep.data.len());
     let mut address_to_frame_idx: IndexSet<_> = IndexSet::new();
     let mut thread_metadata: HashMap<String, ThreadMetadata> = HashMap::new();
+    // map form thread id to number of samples for that thread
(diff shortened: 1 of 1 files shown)


sentry-core/src/profiling.rs
```diff
@@ -221,6 +228,8 @@ fn get_profile_from_report(
             });
     }
 
+    samples.retain(|s| thread_samples.get(&s.thread_id).unwrap() > &1);
```
Swatinem, line 231:
```suggestion
    samples.retain(|s| thread_samples.get(&s.thread_id).unwrap_or_default() > &1);
```

Swatinem (approved):

sentry-core/src/profiling.rs
viglia: @Swatinem does it actually work in this case?

The compiler suggests it's not for `&u64`:
```
the trait bound `&u64: std::default::Default` is not satisfied
the trait `std::default::Default` is implemented for `u64`
```

Swatinem: Ah, you can do a `.get().copied().unwrap_or_default() > 1` in that case.

I mean the `unwrap` itself is pretty much unreachable, though the compiler can’t know that :-)

PR #564, opened 2023-04-06
Add default TlsConnector for ureq when native-tls is enabled

### LINKED ISSUE #563: Using ureq with native-tls is not working
### Environment

sentry "0.30.0"

### Steps to Reproduce

Run 
```rust
fn main() {
    env_logger::init();

    let _guard = sentry::init((
        "SENTRY ENDPOINT",
        sentry::ClientOptions {
            release: sentry::release_name!(),
            ..Default::default()
        },
    ));

    std::thread::spawn(|| {
        std::thread::sleep(std::time::Duration::from_secs(1));
        panic!("Everything is on fire!");
    });
    std::thread::sleep(std::time::Duration::from_secs(5));
}
```
with this config 
```
[package]
edition = "2021"
name = "mrp"
version = "0.1.0"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
env_logger = "0.10"
sentry = { version = "0.30.0", default-features = false, features = [
    "backtrace",
    "contexts",
    "panic",
    "ureq",
    "native-tls",
    # "rustls",  <- if we use rustls and comment native-tls it works
] }

```

### Expected Result

the `panic!` event makes it to sentry

### Actual Result

No event gets generated.

Upon further inspection it seems that `ureq` provides no default when using it with `native-tls` (see [native-tls feature description](https://docs.rs/ureq/latest/ureq/#features) and in [the code](https://github.com/algesten/ureq/blob/2e7bcf255ac777f8f9867d683bc9b88dcefd5854/src/lib.rs#L377-L402))

jose-acevedoflores:
closes #563

### DIFF
--- a/sentry/src/transports/ureq.rs
+++ b/sentry/src/transports/ureq.rs
@@ -43,16 +43,18 @@ impl UreqHttpTransport {
         let agent = agent.unwrap_or_else(|| {
             let mut builder = AgentBuilder::new();
 
-            if options.accept_invalid_certs {
-                #[cfg(feature = "native-tls")]
-                {
-                    let tls_connector = TlsConnector::builder()
-                        .danger_accept_invalid_certs(true)
-                        .build()
-                        .unwrap();
-                    builder = builder.tls_connector(Arc::new(tls_connector));
+            #[cfg(feature = "native-tls")]
+            {
+                let mut tls_connector_builder = TlsConnector::builder();
+
+                if options.accept_invalid_certs {
+                    tls_connector_builder.danger_accept_invalid_certs(true);
                 }
 
+                builder = builder.tls_connector(Arc::new(tls_connector_builder.build().unwrap()));
+            }
+
+            if options.accept_invalid_certs {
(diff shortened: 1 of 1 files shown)


sentry/src/transports/ureq.rs
```diff
@@ -43,16 +43,18 @@ impl UreqHttpTransport {
...
+                if options.accept_invalid_certs {
+                    tls_connector_builder.danger_accept_invalid_certs(true);
                 }
 
+                builder = builder.tls_connector(Arc::new(tls_connector_builder.build().unwrap()));
+            }
+
+            if options.accept_invalid_certs {
                 #[cfg(feature = "rustls")]
                 {
```
Swatinem, line 59: Lets see if the compiler will complain about this `if` being empty when the feature is not enabled.

Swatinem (approved):

Swatinem:
I opened up https://github.com/getsentry/sentry-rust/pull/565 to fix the clippy errors.

Issue #541, opened 2023-01-31
Axum integration


Meemaw:
Sentry integration for [axum](https://github.com/tokio-rs/axum)

Swatinem:
We use axum ourself in symbolicator, and since it is based on tower, you can use the tower integration as we do here:

https://github.com/getsentry/symbolicator/blob/54d7e0850fea2af12fc83788495f1d7d5792c1eb/crates/symbolicator/src/endpoints/mod.rs#L34-L36

The only missing piece is maybe using [`MatchedPath`](https://docs.rs/axum/latest/axum/extract/struct.MatchedPath.html) for the transaction name, as it is currently falling back to the full uri which can have very high cardinality and is thus a bad fit for it.

I mentioned this recently as a similar thing was fixed in the actix integration: https://github.com/getsentry/sentry-rust/pull/539#pullrequestreview-1265513290

I believe adding a feature flag to then use that `MatchedPath` should be straight forward. Together with a usage example would be enough for actix users I believe.

PR #570, opened 2023-04-24
sentry-tower: Use `axum::MatchedPath` for transaction names

Turbo87:
If the `axum-matched-path` feature of `sentry-tower` is used, it will enable the corresponding `matched-path` feature of `axum` and use the `MatchedPath` request extension to generate the transaction name. As before, the HTTP method is also still included in the transaction name.

Resolves https://github.com/getsentry/sentry-rust/issues/541

/cc @Swatinem

### DIFF
--- a/sentry-tower/Cargo.toml
+++ b/sentry-tower/Cargo.toml
@@ -14,8 +14,10 @@ rust-version = "1.66"
 
 [features]
 http = ["dep:http", "pin-project", "url"]
+axum-matched-path = ["axum/matched-path"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


Turbo87:
for context: this is supposed to replace https://github.com/rust-lang/crates.io/blob/8b72b4ad67bab12574f7eb8704f8997c047054f1/src/middleware/sentry.rs, but it has the advantage that the `trace_sampler` will get the `MatchedPath` value too, instead of the full URL, which makes it easier to make sampling decisions based on the URL.

Swatinem (approved):

PR #572, opened 2023-04-25
sentry: Add `tower-axum-matched-path` feature

Turbo87:
#570 introduced the corresponding feature on the `sentry-tower` crate, but it isn't really useable directly from the `sentry` crates, which I assume is the recommended way of using these integration crates?

This PR introduces a `tower-axum-matched-path` feature on the top-level `sentry` crate to fix this. I'm not a huge fan of the long feature name, but I couldn't think of anything more suitable 😅 

/cc @Swatinem


Swatinem (approved):
well yes, in hindsight, splitting out every integration from the main crate might not have been the best decision :-D

PR #577, opened 2023-05-18
feat(crons): Add monitor check-in types to sentry-types

evanpurkhiser:
Adds types that define monitor check-in envelopes.

This is essentially a copy from https://github.com/getsentry/relay/blob/master/relay-monitors/src/lib.rs#L39-L142

We would like to use this in sentry-cli to generate monitor check-ins

### DIFF
--- a/sentry-types/src/protocol/monitor.rs
+++ b/sentry-types/src/protocol/monitor.rs
@@ -1,0 +1,113 @@
+use serde::{Deserialize, Serialize, Serializer};
+use uuid::Uuid;
+
+/// Represents the status of the monitor check-in
+#[derive(Clone, Copy, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum MonitorCheckInStatus {
+    /// Check-in had no issues during execution.
+    Ok,
+    /// Check-in failed or otherwise had some issues.
+    Error,
+    /// Check-in is expectred to complete.
+    InProgress,
+    /// Monitor did not check in on time.
+    Missed,
+    /// No status was passed.
+    #[serde(other)]
+    Unknown,
+}
+
+/// Configuration object of the monitor schedule.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+#[serde(tag = "type")]
+pub enum MonitorSchedule {
+    /// A Crontab schedule allows you to use a standard UNIX crontab style schedule string to
+    /// configure when a monitor check-in will be expected on Sentry.
+    Crontab {
+        /// The crontab syntax string defining the schedule.
+        value: String,
+    },
+    /// A Interval schedule allows you to configure a periodic check-in, that will occur at an
+    /// interval after the most recent check-in.
+    Interval {
+        /// The interval value.
+        value: u64,
+        /// The interval unit of the value.
+        unit: MonitorIntervalUnit,
+    },
+}
+
+/// The unit for the interval schedule type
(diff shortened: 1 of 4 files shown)


kamilogorek:
That's a good start. The next step would be to add a new `EnvelopeItem` type with all corresponding traits to the `Envelope` implementation. Once it's there, you should be able to create a `Envelope::new` instance, populate it with your monitor item and send in the same way we do with any other envelopes in the cli - https://github.com/getsentry/sentry-cli/blob/master/src/commands/send_envelope.rs

sentry-types/src/protocol/monitor.rs
```diff
@@ -0,0 +1,113 @@
...
+#[serde(rename_all = "snake_case")]
+#[serde(tag = "type")]
+pub enum Schedule {
+    /// A Crontab schedule allows you to use a standard UNIX crontab style schedule string to
+    /// configure when a monitor check-in will be expected on Sentry.
+    Crontab {
+        /// The crontab syntax string defining the schedule.
+        value: String,
+    },
+    /// A Interval scheudle allows you to configure a periodic checkin, that will occur at an
```
Swatinem, line 32:
```suggestion
    /// A Interval schedule allows you to configure a periodic checkin, that will occur at an
```

sentry-types/src/protocol/monitor.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct MonitorConfig {
+    /// The monitor schedule configuration.
+    pub schedule: Schedule,
+
+    /// How long (in minutes) after the expected checkin time will we wait until we consider the
+    /// checkin to have been missed.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub checkin_margin: Option<u64>,
+
+    /// How long (in minutes) is the checkin allowed to run for in in_rogress before it is
```
Swatinem, line 71:
```suggestion
    /// How long (in minutes) is the checkin allowed to run for in `in_progress` before it is
```

sentry-types/src/protocol/monitor.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// The environment to associate the check-in with.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub environment: Option<String>,
+
+    /// Duration of this check-in since it has started in seconds.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub duration: Option<f64>,
+
+    /// monitor configuration to support upserts. When provided a monitor will be created on Sentry
+    /// upon reciving the first checkin.
```
Swatinem, line 107:
```suggestion
    /// Monitor configuration to support upserts. When provided a monitor will be created on Sentry
    /// upon receiving the first checkin.
```

sentry-types/src/protocol/monitor.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    /// Identifier of the monitor for this check-in.
+    pub monitor_slug: String,
+
+    /// Status of this check-in. Defaults to `"unknown"`.
+    pub status: CheckInStatus,
+
+    /// The environment to associate the check-in with.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub environment: Option<String>,
```
Swatinem, line 100: just curious as I know nothing about the crons product: do we not need/want a `release` as well?

evanpurkhiser: It's not something we support yet, or afaik have a plan to support

evanpurkhiser:
I would like to wait to merge this in once I can confirm this is everything I need to get checkins sent via sentry-cli to relay 👍

Swatinem (approved):

PR #579, opened 2023-05-19
Fix rate-limiting/filtering of raw envelopes

Swatinem:
It turns out, raw envelopes were never ever send out via transports that support rate-limiting, as the `filter` would discard raw envelopes.


sentry/examples/tracing-demo.rs
```diff
@@ -34,21 +34,28 @@ fn main_span1() {
...
 
 #[tracing::instrument]
-fn thread_span1() {
+fn thread_span1(_arg: &str) {
     thread::sleep(Duration::from_millis(200));
 }
 
 #[tracing::instrument]
-fn main_span2() {
+fn main_span2(_arg: SomeArgument) {
```
Swatinem, line 52: these are just some drive-by changes I had laying around that will log the arguments via the `tracing::instrument` macro for testing, you can ignore these :-)

sentry-types/src/protocol/envelope.rs
```diff
@@ -213,12 +215,14 @@ impl Envelope {
     where
         I: Into<EnvelopeItem>,
     {
+        let item = item.into();
+
         let Items::EnvelopeItems(ref mut items) = self.items else {
-            eprintln!("WARNING: This envelope contains raw items. Adding an item is not supported.");
+            if item != EnvelopeItem::Raw {
```
kamilogorek, line 221: How is `cargo fmt` not complaining here? 👀

kamilogorek (approved):

sentry-types/src/protocol/envelope.rs
Swatinem: because `let else` is still not supported after all this time :-(

kamilogorek (approved):

PR #584, opened 2023-05-23
ci: Fix linting issues for nightly release

kamilogorek:
They are failing right now with recent release.


sentry-backtrace/src/integration.rs
```diff
@@ -18,7 +18,7 @@ pub struct ProcessStacktraceIntegration;
 impl ProcessStacktraceIntegration {
     /// Creates a new Integration to process stacktraces.
     pub fn new() -> Self {
-        Self::default()
+        Self {}
```
Swatinem, line 21: actually, unit structs don’t need braces, I’m surprised noone complains about that :-D

Swatinem (approved):

PR #585, opened 2023-05-24
feat(tracing): Improve structure for tracing errors

jan-auer:
Changes how the tracing integration tracks errors with messages and provides more structured data for fields passed via tracing. 

First, this introduces the ability to set Sentry tags via a `tags.*` prefix. The prefix is removed and does no longer occur in the tracing context. The following example will create a tag called `foo`:

```rust
tracing::error!(tags.foo = "bar", "this is an error");
```

Next, the custom context is renamed to `Rust Tracing Fields`. This is closer to the original naming in the `tracing` crate and also disambiguates this data more from Sentry tags.

The top-level exception passed to tracing is now marked with a `tracing` mechanism to annotate in the UI how the exception was captured.

Finally, if both a message and an error are passed to tracing, the message gets moved to a synthetic exception on the top of the stack. Most importantly, this allows for proper grouping and display in the issue stream. Additionally, if `attach_stacktrace` is enabled, the stack trace is placed on the synthetic exception rather than the thread interface, so it is shown inline with the chained exception.

## Example Rendering

Due to the synthetic exception, the title of the issue is now the location of the capture. The captured message is shown underneath. This is also the case in the title bar on issue details:

Screenshot 2023-05-24 at 14 22 05

The exception stack now shows an entry for `tracing::error!` with the logged message, marked with the `tracing` mechanism. The stack trace is captured through the backtrace integration and points to the capture:

Screenshot 2023-05-24 at 14 22 20

If the source errors had backtraces, those would be shown underneath.

## Background

The tracing integration converts fields containing a `dyn Error` into Sentry exception records. This also works with error sources and creates a stack of nested exceptions on Sentry. Additionally, the formatted message from the `error!` invocation is added as a message. Finally, if `attach_stacktrace` is enabled on the client, the integration also adds a stacktrace. 

For both grouping and rendering, this is suboptimal for the following reasons:
- The logged message usually contains a more specific or useful description than the logged error. However, Sentry does not render the attached message in issue stream, which makes it hard to disambiguate places where the same error is tracked.
- Neither the message nor the stack trace contribute to grouping if there is an exception list. The exceptions are generic, however, which means unrelated problems would be grouped together. 
- Likewise, the message does not end up in the issue title.
- The Sentry UI currently does not even display the exception list if there is a message and a stack trace on the issue. This will have to be fixed in the product, however.

A synthetic exception in Sentry is an exception record that was created from something that's not a language-level exception. It can carry a stack trace and a value, but unlike regular exceptions, the "type" has no meaning and is not used for grouping or issue title. Instead, Sentry will use the source location from the stack trace to for both.

With this, we receive the behavior that is most likely desirable in such a scenario:
- A chain of exceptions is shown with the logged message and stack trace on top, pointing to the log location.
- Grouping does not use the synthetic "tracing::error!" type, it uses the source location.
- Issue details show a descriptive location and the message.

### DIFF
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -1,8 +1,8 @@
 use std::collections::BTreeMap;
 use std::error::Error;
 
-use sentry_core::protocol::{Event, Exception, Value};
-use sentry_core::{event_from_error, Breadcrumb, Level};
+use sentry_core::protocol::{Event, Exception, Mechanism, Thread, Value};
+use sentry_core::{event_from_error, Breadcrumb, Hub, Level};
 use tracing_core::field::{Field, Visit};
 use tracing_core::{span, Subscriber};
 use tracing_subscriber::layer::Context;
@@ -106,9 +106,27 @@ pub fn breadcrumb_from_event(event: &tracing_core::Event) -> Breadcrumb {
     }
 }
 
+fn tags_from_event(fields: &mut BTreeMap<String, Value>) -> BTreeMap<String, String> {
+    let mut tags = BTreeMap::new();
+
+    fields.retain(|key, value| {
+        let Some(key) = key.strip_prefix("tags.") else { return true };
+        let string = match value {
+            Value::String(s) => std::mem::take(s),
+            _ => return true,
+        };
+
+        tags.insert(key.to_owned(), string);
+
+        false
+    });
+
+    tags
+}
+
 fn contexts_from_event(
     event: &tracing_core::Event,
-    event_tags: BTreeMap<String, Value>,
+    fields: BTreeMap<String, Value>,
 ) -> BTreeMap<String, sentry_core::protocol::Context> {
     let event_meta = event.metadata();
     let mut location_map = BTreeMap::new();
@@ -123,10 +141,10 @@ fn contexts_from_event(
     }
 
     let mut context = BTreeMap::new();
-    if !event_tags.is_empty() {
+    if !fields.is_empty() {
         context.insert(
-            "Rust Tracing Tags".to_string(),
-            sentry_core::protocol::Context::Other(event_tags),
+            "Rust Tracing Fields".to_string(),
+            sentry_core::protocol::Context::Other(fields),
         );
     }
     if !location_map.is_empty() {
@@ -143,12 +161,13 @@ pub fn event_from_event<S>(event: &tracing_core::Event, _ctx: Context<S>) -> Eve
 where
     S: Subscriber + for<'a> LookupSpan<'a>,
 {
-    let (message, visitor) = extract_event_data(event);
+    let (message, mut visitor) = extract_event_data(event);
 
     Event {
         logger: Some(event.metadata().target().to_owned()),
         level: convert_tracing_level(event.metadata().level()),
         message,
+        tags: tags_from_event(&mut visitor.json_values),
         contexts: contexts_from_event(event, visitor.json_values),
         ..Default::default()
     }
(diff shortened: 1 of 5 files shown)


jan-auer:
I'm planning to follow up with a separate PR that improves docs around tracking `dyn Error` with tracing for Sentry.

sentry-tracing/src/converters.rs
```diff
@@ -163,13 +182,53 @@ where
...
+    } = visitor;
+
+    if !exceptions.is_empty() && message.is_some() {
+        #[allow(unused_mut)]
+        let mut thread = Thread::default();
+
+        #[cfg(feature = "backtrace")]
+        if let Some(client) = sentry_core::Hub::current().client() {
+            if client.options().attach_stacktrace {
+                thread = sentry_backtrace::current_thread(true);
```
jan-auer, line 198: These two changes I'm not happy with. In order to attach the stack trace, I need to depend on `sentry_backtrace` and also grab the client to check its options. 

The backtrace integration could detect this case (attach_stacktrace enabled + top-level exception is synthetic) and write there. Would that be a better alternative? It composes better, but I'm not sure if we're happy with these semantics.

sentry/tests/test_tracing.rs
```diff
@@ -106,7 +106,7 @@ fn test_tracing() {
 
 #[tracing::instrument(fields(span_field))]
 fn function() {
-    tracing::Span::current().record("span_field", "some data");
+    tracing::Span::current().record("span_field", &"some data");
```
jan-auer, line 109: If I remove this, I receive a compilation error:

```
the size for values of type `str` cannot be known at compilation time
the trait `std::marker::Sized` is not implemented for `str`
```

However, the linter now wants to remove this borrow. It looks like we're linting with whatever rust version is installed on the GHA workers without installing a specific version. Is that intended?

jan-auer: In CI, this passes again, but locally this continues to give me a compilation error.

Swatinem:
Some questions from my end:

> Neither the message nor the stack trace contribute to grouping if there is an exception list. The exceptions are generic, however, which means unrelated problems would be grouped together.

While the issue title that ends up in sentry is crap, the separate grouping depending on the chained errors is actually a good thing, at least for the use-case in symbolicator, where we get different groups for different errors with broken debug files. If all those would be grouped as a single issue, that would be worse. Though I have to look up what mechanism we end up using to capture these in symbolicator.

> A synthetic exception in Sentry is an exception record that was created from something that's not a language-level exception. It can carry a stack trace and a value, but unlike regular exceptions, the "type" has no meaning and is not used for grouping or issue title. Instead, Sentry will use the source location from the stack trace to for both.

Depending on the question related to `sentry-backtrace`: We could also just put the tracing location in there as the a single-frame stack trace maybe, in case when we don’t have a *proper* stack trace.

sentry-tracing/src/converters.rs
Swatinem: I believe its a fair thing to do. However turning on `attach_stacktrace` is a rather hard hammer, and may have high overhead if you just want to log random messages, and not necessarily errors.

As I said in another comment. Another alternative would be to add the `"Rust Tracing Location"` as it is now as the stack trace here. That is definitely better than it being hidden somewhere in a random context.

Swatinem (approved):

jan-auer:
> separate grouping depending on the chained errors is actually a good thing

Agreed, and this is also what's happening with the changes in this PR. Even more so, it now groups by the stack trace (ie. the log location) _plus_ the chained exceptions (and their stack traces if they have any).

sentry-tracing/src/converters.rs
jan-auer: I was thinking about adding the tracing location, too. However, it doesn't have enough information to create even a single stack frame that is eligible for grouping, most notably because of the missing function name. 

We could also pass this as an option to the tracing integration and pass it down to the converter function, so you can control this separately. At least in Relay, we've enabled stack traces for everything (they are just added for error messages anyway, otherwise we're not tracking stuff in Sentry), so the behavior is intended. This is a good consideration for what attach_stacktraces is actually supposed to mean; and I think in this specific case (we have an exception entry) it's seems exactly as intended.

PR #588, opened 2023-06-06
fix: Allow for Contexts data type to be infered from the key value

kamilogorek:
~Currently, tests are failing due to visitor going more than one level deep and nested objects effectively getting top-level key assigned to them.~

~Or rather, after more investigation, it simply puts a `type` field into the flattened `other` map, as `type` is not removed; it normally is, as it's used as `tag = "type"` during deserialization.~

tl;dr - we want both variants to be deserializable (currently first example is failing due to missing `type` key):

``` json
{
  "contexts": {
    "device": {
      "name": "fancy name for a fancy server",
      "arch": "amd64",
      "num_cpu": 8
    },
}
```

```json
{
  "contexts": {
    "device": {
      "type": "device",
      "name": "fancy name for a fancy server",
      "arch": "amd64",
      "num_cpu": 8
    },
}
```

Closes https://github.com/getsentry/sentry-rust/issues/483

### DIFF
--- a/sentry-types/src/protocol/v7.rs
+++ b/sentry-types/src/protocol/v7.rs
@@ -16,8 +16,7 @@ use std::str;
 use std::time::SystemTime;
 
 use self::debugid::{CodeId, DebugId};
-use serde::Serializer;
-use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


loewenheim (approved):
Oof, that's pretty gnarly. Respect for getting this done.

PR #592, opened 2023-06-15
chore(deps): bump rustls

utkarshgupta137:
ureq 2.7 is using rustls 0.21 now, so sentry won't build if both `rustls` & `ureq` are enabled.


utkarshgupta137:
@Swatinem Can you do a release, our CIs are broken.

PR #603, opened 2023-08-07
docs: `debug-images` should be listed as a default feature

timfish:
`debug-images` was moved to the default features but the docs were not updated


Swatinem (approved):
lgtm, looks like we need some crate maintenance though

Swatinem:
I cherry-picked this into https://github.com/getsentry/sentry-rust/pull/610 which has now been released.

PR #606, opened 2023-08-23
Bump `webpki-roots` to ^0.25.1

marcusirgens:
See https://github.com/rustls/webpki-roots/compare/v/0.23.1..08972db98bafca0bb1ed682b018f0e37352a2767.

Removes a transitive dependency on `webpki`, which is currently pulling in vulnerable code to one of our projects.

See https://rustsec.org/advisories/RUSTSEC-2023-0052.

The latest tagged version of `sentry` (`0.31.5`) still pulls `webpki-roots`@`0.22.6`, which pulls `webpki`@`^0.22`.

PR #609, opened 2023-08-28
Update webpki-roots to fix RUSTSEC-2023-0052

FSMaxB:
webpki-roots 0.23 still depends on webpki 0.22 which is affected by a denial of service vulnerability, see https://rustsec.org/advisories/RUSTSEC-2023-0052.html

Opening this PR despite #606 because that one doesn't fix the compilation error that the update brings with it.


Swatinem:
Thanks for opening this PR. I have cherry-picked #606 into https://github.com/getsentry/sentry-rust/pull/610 which also includes necessary MSRV bumps to actually pass CI, so I will rather go ahead with that.

PR #614, opened 2023-09-14
Use `rand` instead of `getrandom` for IDs

Swatinem:
The `getrandom` crate uses syscalls to get *true* randomness from the OS. We do not need that high quality of randomness, we do not even need cryptographically secure randomness.

This switches to the `rand` crate which uses a reseeding PRNG which is still cryptographically secure though.

This is much faster, and we can switch to an even simpler PRNG if it is not fast enough still.

### DIFF
--- a/sentry-types/Cargo.toml
+++ b/sentry-types/Cargo.toml
@@ -22,11 +22,11 @@ protocol = []
 
 [dependencies]
 debugid = { version = "0.8.0", features = ["serde"] }
-getrandom = "0.2.3"
 hex = "0.4.3"
+rand = "0.8.5"
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = "1.0.46"
 thiserror = "1.0.15"
 time = { version = "0.3.5", features = ["formatting", "parsing"] }
 url = { version = "2.1.1", features = ["serde"] }
-uuid = { version = "1.0.0", features = ["v4", "serde"] }
+uuid = { version = "1.0.0", features = ["serde"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2818,8 +2818,8 @@ name = "sentry-types"
 version = "0.31.6"
 dependencies = [
  "debugid",
- "getrandom 0.2.10",
  "hex",
+ "rand 0.8.5",
  "serde",
  "serde_json",
  "thiserror",
--- a/sentry-core/src/api.rs
+++ b/sentry-core/src/api.rs
@@ -17,9 +17,9 @@ use crate::{Hub, Integration, IntoBreadcrumbs, Scope};
 ///
 /// ```
 /// use sentry::protocol::{Event, Level};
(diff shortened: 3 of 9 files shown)


loewenheim (approved):

jan-auer (approved):

sentry-types/Cargo.toml
```diff
@@ -22,11 +22,11 @@ protocol = []
...
-getrandom = "0.2.3"
 hex = "0.4.3"
+rand = "0.8.5"
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = "1.0.46"
 thiserror = "1.0.15"
 time = { version = "0.3.5", features = ["formatting", "parsing"] }
 url = { version = "2.1.1", features = ["serde"] }
-uuid = { version = "1.0.0", features = ["v4", "serde"] }
+uuid = { version = "1.0.0", features = ["serde"] }
```
MarijnS95, line 32: Note that, because `Uuid` is _publicly_ reexported, this could be considered a breaking change because downstream crates are also calling `Uuid::new_v4` on the reexported type, and are now breaking because the function is no longer available:

https://github.com/EmbarkStudios/sentry-contrib-rust/pull/27

Maybe this shouldn't have been published as a non-breaking `0.31.7` release?

(On a related sidenote, I haven't noticed this in my crate hierarchy because other dependencies were still turning on the `v4` fetaure)

Swatinem: That is a very good point!
Maybe it does make sense to set up `cargo-semver-checks` or similar tooling that would ideally catch these problems.

MarijnS95: That might be nice; it's been on the back of my mind for a while for the crates I maintain, but have never got to use it.

For this case, are there any plans to "correct" this (adding back `v4`, yanking the past 2 patch releases) or shall we let it slide?
Because it is reexported, we cannot mark `new_v4()` as `deprecated` to make users aware of the new function, unfortunately.

MarijnS95: @Swatinem anything to follow up on here?

Swatinem: Sorry this flew under the radar for such a long time.

> For this case, are there any plans to "correct" this (adding back v4, yanking the past 2 patch releases) or shall we let it slide?

given that so much time has passed, "let it slide" was the answer here.

TBH, we don’t really have a good story for yanking releases, and would rather roll forward in such cases. And since then we have also bumped the minor (aka semver major) version.

I do feel the pain, as I also use `reqwest::Url` instead of `url::Url` in way too many places.

Another problem here is that the SDK has to do "semver major" bumps all the time, also because of all the integration crates. A concrete example here is also that our own team was not able to update the SDK because it bumped the `tower` version which forced that version update on the consumers as well.

Long story short I’m afraid I don’t have a good solution to all this :-(

PR #615, opened 2023-09-14
Micro-optimize Scope access

Swatinem:
Instead of having the compiler do bounds-checks and generate panic-ing code, just make that case impossible by making sure there is always a top scope.


loewenheim (approved):
Making illegal states unrepresentable FTW

PR #618, opened 2023-10-17
Add experimental metrics implementation

Swatinem:
This adds:
- A new `statsd` EnvelopeItem
- A new `MetricAggregator` which is implemented similarly to the Python prototype implementation 
- The Client instantiates a metric flusher and starts it in the background
- A `Metric` type that allows building and sending metrics, which is the primary user-facing API
- A `cadence`-compatible `SentryMetricSink` that makes it easier to integrate Sentry into existing environments
- Internal types that are mostly adapted from `relay-metrics`

This PR does not yet include support for code locations and span linking; both will follow at a later stage.

### DIFF
--- a/sentry-core/src/cadence.rs
+++ b/sentry-core/src/cadence.rs
@@ -147,7 +147,7 @@ mod tests {
         assert_eq!(envelopes.len(), 1);
 
         let mut items = envelopes[0].items();
-        let Some(EnvelopeItem::Metrics(metrics)) = items.next() else {
+        let Some(EnvelopeItem::Statsd(metrics)) = items.next() else {
             panic!("expected metrics");
         };
         let metrics = std::str::from_utf8(metrics).unwrap();
--- a/sentry-core/src/cadence.rs
+++ b/sentry-core/src/cadence.rs
@@ -1,0 +1,90 @@
+use std::sync::Arc;
+
+use cadence::MetricSink;
+
+use crate::{Client, Hub};
+
+/// A [`cadence`] compatible [`MetricSink`].
+///
+/// This will ingest all the emitted metrics to Sentry as well as forward them
+/// to the inner [`MetricSink`].
+#[derive(Debug)]
+pub struct SentryMetricSink<S> {
+    client: Arc<Client>,
+    sink: S,
+}
+
+impl<S> SentryMetricSink<S> {
+    /// Creates a new [`SentryMetricSink`], wrapping the given [`MetricSink`].
+    pub fn try_new(sink: S) -> Result<Self, S> {
+        let hub = Hub::current();
+        let Some(client) = hub.client() else {
+            return Err(sink);
+        };
+
+        Ok(Self { client, sink })
+    }
+}
+
+impl<S> MetricSink for SentryMetricSink<S>
+where
+    S: MetricSink,
+{
+    fn emit(&self, metric: &str) -> std::io::Result<usize> {
+        self.client.add_metric(metric);
+        self.sink.emit(metric)
+    }
+
+    fn flush(&self) -> std::io::Result<()> {
+        if self.client.flush(None) {
+            self.sink.flush()
+        } else {
+            Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "Flushing Client failed",
+            ))
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use cadence::{Counted, Distributed};
+    use sentry_types::protocol::latest::EnvelopeItem;
+
+    use crate::test::with_captured_envelopes;
+
+    use super::*;
+
+    #[test]
+    fn test_basic_metrics() {
+        let envelopes = with_captured_envelopes(|| {
+            let sink = SentryMetricSink::try_new(cadence::NopMetricSink).unwrap();
+
+            let client = cadence::StatsdClient::from_sink("sentry.test", sink);
+            client.count("some.count", 1).unwrap();
+            client.count("some.count", 10).unwrap();
+            client
+                .count_with_tags("count.with.tags", 1)
+                .with_tag("foo", "bar")
+                .send();
+            client.distribution("some.distr", 1).unwrap();
+            client.distribution("some.distr", 2).unwrap();
+            client.distribution("some.distr", 3).unwrap();
+        });
+        assert_eq!(envelopes.len(), 1);
+
+        let mut items = envelopes[0].items();
+        let Some(EnvelopeItem::Metrics(metrics)) = items.next() else {
+            panic!("expected metrics");
+        };
+        let metrics = std::str::from_utf8(metrics).unwrap();
+
+        println!("{metrics}");
+
+        assert!(metrics.contains("sentry.test.count.with.tags:1|c|#foo:bar|T"));
+        assert!(metrics.contains("sentry.test.some.count:11|c|T"));
+        assert!(metrics.contains("sentry.test.some.distr:1:2:3|d|T"));
+        assert_eq!(items.next(), None);
+    }
+}
--- a/sentry-core/src/cadence.rs
+++ b/sentry-core/src/cadence.rs
@@ -17,12 +17,10 @@ pub struct SentryMetricSink<S> {
 impl<S> SentryMetricSink<S> {
     /// Creates a new [`SentryMetricSink`], wrapping the given [`MetricSink`].
     pub fn try_new(sink: S) -> Result<Self, S> {
-        let hub = Hub::current();
-        let Some(client) = hub.client() else {
-            return Err(sink);
-        };
-
-        Ok(Self { client, sink })
+        match Hub::current().client() {
+            Some(client) => Ok(Self { client, sink }),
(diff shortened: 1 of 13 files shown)


sentry-core/src/metrics.rs
```diff
@@ -0,0 +1,404 @@
...
+                    match buckets.entry(key) {
+                        Entry::Occupied(mut entry) => {
+                            let _ = entry.get_mut().merge(value);
+                        }
+                        Entry::Vacant(entry) => {
+                            entry.insert(value);
+                        }
+                    }
+                }
+                _ => {
```
loewenheim, line 286:
```suggestion
                Ok(Task::Shutdown) => {
```

sentry-core/src/metrics.rs
```diff
@@ -0,0 +1,404 @@
...
+        if let Some(parsed_metric) = parse(metric) {
+            let _ = self.sender.send(Task::SendMetrics(parsed_metric));
+        }
+    }
+
+    pub fn flush(&self) {
+        let _ = self.sender.send(Task::Flush);
+    }
+
+    fn worker_thread(receiver: Receiver<Task>, shutdown: Arc<AtomicBool>, transport: TransportArc) {
```
loewenheim, line 250: Why does this have the `shutdown` parameter in addition to the `Shutdown` variant of `Task`? Wouldn't sending `Task::Shutdown` be enough?

sentry-core/src/metrics.rs
Swatinem: Not quite, there is also `RecvTimeoutError::Disconnected`, in which case shutting down is the only viable thing to do.

sentry-core/src/metrics.rs
```diff
@@ -0,0 +1,404 @@
...
+
+struct GaugeValue {
+    last: f64,
+    min: f64,
+    max: f64,
+    sum: f64,
+    count: u64,
+}
+enum BucketValue {
+    Counter(f64),
```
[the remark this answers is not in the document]
Swatinem: That is what `relay-metrics` uses: https://getsentry.github.io/relay/relay_metrics/type.CounterType.html
So I just stuck with that. Admittedly it does not make as much sense.

sentry-core/src/metrics.rs
Swatinem: Thats a good point, though I copied most of this code from the transport that has the same logic. I can’t remember why that was, but also I don’t want to change it too much tbh.

sentry-core/src/metrics.rs
loewenheim: Ah, I didn't catch that `RecvTimeoutError` has two variants.

sentry-core/src/metrics.rs
Swatinem: well, it looks like the tests were not happy with the way that explicit forceful shutdown and the channel interacted with each other, so I removed that.
The difference to the transport is that the transport can have long running http requests, and you can explicitly flush with a timeout, and then force-shutdown.

sentry-core/src/metrics.rs
```diff
@@ -0,0 +1,387 @@
...
+    Flush,
+    Shutdown,
+}
+
+pub struct MetricFlusher {
+    sender: SyncSender<Task>,
+    handle: Option<JoinHandle<()>>,
+}
+
+const FLUSH_INTERVAL: Duration = Duration::from_secs(10);
```
jan-auer, line 173: This constant is used for two separate concerns, I'd suggest to separate them: 
 1. The bucketing interval, to which timestamps will be rounded when reporting metrics upstream.
 2. The interval for flushing buckets upstream.

Right now, I do not see a strong reason why those two have to be different, but it is important not to change the bucketing interval once we choose a different flush interval.

sentry-core/src/metrics.rs
```diff
@@ -0,0 +1,387 @@
...
+                    let rounded_timestamp = (key.timestamp / rounding_interval) * rounding_interval;
+
+                    key.timestamp = rounded_timestamp;
+
+                    match buckets.entry(key) {
+                        Entry::Occupied(mut entry) => {
+                            let _ = entry.get_mut().merge(value);
+                        }
+                        Entry::Vacant(entry) => {
+                            entry.insert(value);
```
jan-auer, line 267: When inserting metrics with high volume, the total size of cached buckets could grow larger than what we can send to the server. In the Python SDK, we therefore count the number of data points inserted with this ruleset:

- gauge creation: +5, all others: +1
- add to set or distribution: +1, others +0

Then, when we cross a hard-coded threshold, we flush immediately after adding to ensure that outgoing envelopes to not grow indefinitely.

sentry-core/src/client.rs
```diff
@@ -308,11 +322,22 @@ impl Client {
         }
     }
 
+    #[cfg(feature = "UNSTABLE_metrics")]
+    pub(crate) fn send_metric(&self, metric: &str) {
+        if let Some(ref flusher) = *self.metrics_flusher.read().unwrap() {
+            flusher.send_metric(metric)
```
jan-auer, line 328: I would propose a slight change in the signatures here. Apart from using cadence, we should offer a convenient programmatic interface to capture metrics from code directly. For inspiration, have a look at `sentry_sdk.metrics` in the Python SDK.

1. One basic function to capture metrics with structured inputs: name, unit, value, and tags. This could be a function `capture_metric(&self, metric: Metric)`. See [this](https://github.com/getsentry/relay/blob/e7d779eeebcfe7dd3556bfd37be44428c4815dbb/relay-metrics/src/protocol.rs#L321) type from an older version of Relay for an example. Note how there is a type `MetricValue` separate from `BucketValue` for the user. 
2. Convenience functions for `.count` / `.incr`, `.gauge`, `.time` etc -- either named like in the Sentry SDK or like in cadence.
3. Implement `FromStr for Metric` to support parsing plain statsd payloads (which cannot contain packed values), as required by the cadence sink sadly.
4. The cadence sink could parse into `Metric`, then invoke `capture_metric` on `Hub::current`. This way, the CPU for parsing statsd payload occurs at call site, not in the central background thread which might backlog easier.

sentry-core/src/metrics.rs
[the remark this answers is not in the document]
jan-auer: This allows the user to count / accumulate things that are not whole numbers.

jan-auer:
I've updated the implementation to address above comments. The current state is still missing a few things, however:

 1. There's no shift while flushing
 2. The cadence sink test is flakey
 3. There are no tests of the metrics module
 4. The public API needs a review, we should probably add top-level utility functions

sentry-core/src/metrics.rs
```diff
@@ -0,0 +1,1171 @@
...
+///  1. The rounded UNIX timestamp of buckets.
+///  2. The metric buckets themselves with a corresponding timestamp.
+///
+/// This structure allows for efficient dequeueing of buckets that are older than a certain
+/// threshold. The buckets are dequeued in order of their timestamp, so the oldest buckets are
+/// dequeued first.
+type BucketMap = BTreeMap<Timestamp, HashMap<BucketKey, BucketValue>>;
+
+#[derive(Debug)]
+struct SharedAggregatorState {
```
loewenheim, line 344: Nit: this struct belongs further down with the other aggregator types.

sentry-core/src/units.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+}
+
+impl std::str::FromStr for MetricUnit {
+    type Err = ParseMetricUnitError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(match s {
+            "nanosecond" | "ns" => Self::Duration(DurationUnit::NanoSecond),
+            "microsecond" => Self::Duration(DurationUnit::MicroSecond),
```
loewenheim, line 55: Could do `μs` here, but I don't know if that's common.

loewenheim (approved):
Some nits, but otherwise looks good

sentry-core/src/units.rs
jan-auer: Good point, `μ` is an ASCII character so we could add this as an alias. We'd have to add this to the spec and Relay too, though.

sentry-core/src/cadence.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+impl MetricSink for SentryMetricSink {
+    fn emit(&self, string: &str) -> std::io::Result<usize> {
+        if let Ok(metric) = Metric::parse_statsd(string) {
+            if let Some(ref client) = self.client {
+                client.add_metric(metric);
+            } else if let Some(client) = Hub::current().client() {
+                client.add_metric(metric);
+            }
```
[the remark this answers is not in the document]
jan-auer: Unfortunately, this method would have to either always clone an `Arc`, or return an enum. That's why I left it as copy-paste.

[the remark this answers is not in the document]
loewenheim: Ah, right. Thought you could get a reference to the `Hub::current().client()`, but that's temporary.

PR #622, opened 2023-10-25
docs(tower): Mention how to enable http feature from sentry crate

flub:
It is common to have this integration pulled in via the tower feature of the sentry crate.  In this case it is convenient to know about the shortcut of the tower-http feature instead of having to add a direct dependency to sentry-tower.


flub:
See #623 for the CI failure, that's not caused by this PR.

Swatinem (approved):

PR #623, opened 2023-10-25
fix(docs): Fix some doc errors that slipped in

flub:


Swatinem (approved):
good to see you :-)

yes I believe these are new lints in Rust now. I also fixed that in another PR of mine which was not yet merged.

PR #627, opened 2023-11-28
Support for axum 0.7.1

benjaminjellis:
Hi,

I'm not sure if you accept contributions so feel free to close this if not. 

Axum recently update to `0.7.0` which doesn't compile using sentry `0.31.8`, as it turns out this is simple to fix by bumping the http dependency in `sentry-tower` to `1.0.0` to match that in axum.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,7 +40,7 @@ dependencies = [
  "flate2",
  "futures-core",
  "h2",
- "http",
+ "http 0.2.9",
  "httparse",
  "httpdate",
  "itoa",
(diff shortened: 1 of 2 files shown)


Swatinem (approved):
I’m a bit surprised this is the only thing that needs bumping, I thought there was more in the tower ecosystem that had version bumps due to hyper.

TimDiekmann:
I don't know the typical release flow for sentry, is it just the release workflow dispatch? If so, would it be possible to trigger it?

indietyp:
Also it appears like http 0.2 is still in the dependency tree, axum and actix seem to need updating as well. 🤔

Swatinem:
I did a ton more dependency updates in https://github.com/getsentry/sentry-rust/pull/628

Issue #617, opened 2023-09-27
Propagate tracing span attributes in sentry events


samscott89:
Hey,

Thanks for providing Rust support. We're happily using `sentry` and `sentry-tracing` crates. Mostly all working well.

We expose a ton of helpful debugging information through span attributes. As far as I can tell, these are captured for performance monitoring, but aren't provided as context when capturing events (via `tracing::error!` for example).

Is there any way that any span data captured could also be included in events emitted by `sentry-tracing`?

Swatinem:
I don’t think this is currently possible. Other members of the team are prototyping ways to rewamp the performance monitoring / tracing APIs to make such things easier in the future, but it is still very early in that process.

samscott89:
Cool, looking forward to seeing what happens there. Let me know if you want any feedback/testers, we'd be open to trying stuff out.

tsionyx:
Hello. I just [made the implementation](https://github.com/getsentry/sentry-rust/pull/629) of this feature. Would be glad if someone looks at it and maybe suggest some improvements.

thomaseizinger:
Unfortuantely, this feature appears to be tied to a `span_filter` that records the span which in turn might spam many spans to Sentry's Tracing feature. It would be great if we can record fields from spans in breadcrumbs while at the same time, only submit certain spans to Sentry.

thomaseizinger:
> It would be great if we can record fields from spans in breadcrumbs while at the same time, only submit certain spans to Sentry.

The way to achieve this is to use a `span_filter` that allows all spans, paired with a `traces_sampler` that inspects the current span and computes an appropriate sampling rate (i.e. 0.0 for discard or 1.0 for include).

PR #629, opened 2023-12-02
tracing: send spans' attributes along with the event ones

tsionyx:
Implementing https://github.com/getsentry/sentry-rust/issues/617

Adds a `enable_span_attributes()` method for the `sentry_tracing::SentryLayer` that activates the desired behaviour.
The root span's attributes can only be propagated once the traces enabled (e.g. through `ClientOptions::traces_sample_rate`).

### DIFF
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -147,7 +147,10 @@ impl Visit for FieldVisitor {
 }
 
 /// Creates a [`Breadcrumb`] from a given [`tracing_core::Event`]
-pub fn breadcrumb_from_event<S>(event: &tracing_core::Event, _ctx: Option<Context<S>>) -> Breadcrumb
+pub fn breadcrumb_from_event<'context, S>(
+    event: &tracing_core::Event,
+    _ctx: impl Into<Option<Context<'context, S>>>,
+) -> Breadcrumb
 where
     S: Subscriber + for<'a> LookupSpan<'a>,
 {
@@ -220,11 +223,14 @@ fn contexts_from_event(
 }
 
 /// Creates an [`Event`] from a given [`tracing_core::Event`]
-pub fn event_from_event<S>(event: &tracing_core::Event, ctx: Option<Context<S>>) -> Event<'static>
+pub fn event_from_event<'context, S>(
+    event: &tracing_core::Event,
+    ctx: impl Into<Option<Context<'context, S>>>,
+) -> Event<'static>
 where
     S: Subscriber + for<'a> LookupSpan<'a>,
 {
-    let (message, mut visitor) = extract_event_data(event, ctx);
+    let (message, mut visitor) = extract_event_data(event, ctx.into());
 
     Event {
         logger: Some(event.metadata().target().to_owned()),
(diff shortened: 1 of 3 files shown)


tsionyx:
```rust
use sentry::{
    integrations::tracing::{self as sentry_tracing, EventFilter},
    ClientOptions,
};
use tracing::{span, warn, Level};
use tracing_subscriber::{layer::SubscriberExt as _, util::SubscriberInitExt as _};

const DSN: &str = "MY_DSN";

fn main() {
    let _guard = sentry::init((
        DSN,
        ClientOptions {
            traces_sample_rate: 1.0,
            ..Default::default()
        },
    ));

    let sentry_layer = sentry_tracing::layer()
        .enable_span_attributes()
        .event_filter(|md| match *md.level() {
            Level::ERROR | Level::WARN => EventFilter::Event,
            _ => EventFilter::Ignore,
        });

    tracing_subscriber::registry()
        .with(sentry_layer)
        .with(tracing_subscriber::fmt::layer())
        .init();

    // create very simple spans hierarchy
    let span = span!(Level::ERROR, "root span", span_int = 42, span_str = "hello");
    span.in_scope(|| {
        span!(
            Level::WARN,
            "child span",
            span_num = 144,
            span_str = "world"
        )
        .in_scope(|| {
            warn!(event_data = 3.1417, "The event message");
        })
    });
}
```


gives the following output in the dashboard

image

sentry-tracing/src/converters.rs
```diff
@@ -174,11 +220,11 @@ fn contexts_from_event(
 }
 
 /// Creates an [`Event`] from a given [`tracing_core::Event`]
-pub fn event_from_event<S>(event: &tracing_core::Event, _ctx: Context<S>) -> Event<'static>
+pub fn event_from_event<S>(event: &tracing_core::Event, ctx: Option<Context<S>>) -> Event<'static>
```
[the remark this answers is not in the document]
tsionyx: I updated the type from `Option<Context>` to `impl Into<Option<Context>>`. This modification restores the capability to provide either a direct `Context` or an `Option<Context>`. As a result, the breaking change has been mitigated.

sentry-tracing/src/converters.rs
```diff
@@ -104,8 +147,14 @@ impl Visit for FieldVisitor {
 }
 
 /// Creates a [`Breadcrumb`] from a given [`tracing_core::Event`]
-pub fn breadcrumb_from_event(event: &tracing_core::Event) -> Breadcrumb {
```
Swatinem, line 107: This is a breaking change anyway because this fn signature is gaining a new parameter. So might as well just go directly with an `Option`, as in `impl Into` is a bit heavy handed.

tsionyx: Since I don't need it anyway, I can remove this new argument from this particular `breadcrumb_from_event`. I put it here before only to make the uniform-looking signatures.

tsionyx: Done, restored the initial `breadcrumb_from_event` signature.

dtheodor:
Yes, adding a span filter to allow spans while keeping the traces_sample_rate to 0 works, thanks

PR #632, opened 2023-12-27
chore(deps): bump `anyhow` and disable its `backtrace` feature

LunaBorowska:
Since anyhow 1.0.77, backtrace feature is not needed for Rust 1.65+ (https://github.com/dtolnay/anyhow/releases/tag/1.0.77). Because sentry has a higher MSRV than that, it's no longer necessary to enable this feature.


Swatinem:
It might not be necessary, but I doubt it hurts to keep it?

I would have to take a closer look at this in any case.

lcian (approved):

PR #646, opened 2024-03-21
fix(ClientInitGuard): ensure drop can only be called once

lbarthon:
In some cases, we might want / need to clone the ClientInitGuard, or maybe even send it in many threads. The only way to get it to work today is to call `std::mem::forget` on it, which is not ideal, as it will never be dropped. This is because the first time the ClientInitGuard will be dropped, the session will be ended, and the client closed.

This can be fixed by checking the # of references to the guard in the drop. By doing so, we can ensure that we close everything only when the last reference to it is dropped. This makes cases where we want to clone it around possible, without using `std::mem::forget`, and thus allowing us to keep the drop behaviour, when needed.

### DIFF
--- a/sentry/src/init.rs
+++ b/sentry/src/init.rs
@@ -28,14 +28,16 @@ impl ClientInitGuard {
 
 impl Drop for ClientInitGuard {
     fn drop(&mut self) {
-        if self.is_enabled() {
-            sentry_debug!("dropping client guard -> disposing client");
-        } else {
-            sentry_debug!("dropping client guard (no client to dispose)");
+        if Arc::strong_count(&self.0) == 1 {
+            if self.is_enabled() {
+                sentry_debug!("dropping client guard -> disposing client");
+            } else {
+                sentry_debug!("dropping client guard (no client to dispose)");
+            }
+            // end any session that might be open before closing the client
+            crate::end_session();
+            self.0.close(None);
         }
-        // end any session that might be open before closing the client
-        crate::end_session();
-        self.0.close(None);
     }
 }
 


Swatinem:
I don’t think this will work at all.

The `Arc<Client>` is part of every Hub, and is implicitly cloned when spawning new threads.

So not only does the *current* thread have one reference to the configured `Hub`, so does every other thread, plus the `ClientInitGuard` itself.

Which all means that you will never really get into the situation where you end up with a single leftover reference.

I was also confused that you say you are `clone`ing the `ClientInitGuard`, which does not implement `Clone` itself.

In either case, if you really want to have shared ownership, the simplest solution is to just wrap the `ClientInitGuard` in yet another `Arc` and make sure you clone that and pass it to other threads.

lbarthon:
Thanks for getting back to this quickly.

I just looked it up, and I was able to clone the `ClientInitGuard`, which didn't clone the `ClientInitGuard`, but the inner `Client`, as it deref the `ClientInitGuard` and then clones it. This is why my implementation was wrong.

I was able to go around that by just calling `std::mem::forget`, as per the doc, but that's something I'd rather avoid. Couldn't we change the init to return an `Arc<ClientInitGuard>`, so that we can keep the guard in many threads if need be, and keep the current behaviour? The `ClientInitGuard` will be dropped when all references to the Arc would be.

Swatinem:
To be quite frank, this is a very niche usecase. Usually you `sentry::init` in the main thread, before calling into the rest of the application.

As I said, you can wrap it in an `Arc` on your end, as changing it in `sentry::init` itself does not provide any benefit to the 99% use-case, and would rather be more confusing.

PR #649, opened 2024-04-04
Update h2 crate because of RUSTSEC-2024-0332

eaufavor:
https://rustsec.org/advisories/RUSTSEC-2024-0332.html

H2 client is [not affected](https://seanmonstar.com/blog/hyper-http2-continuation-flood/) but this update will make things like cargo audit happy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1536,9 +1536,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
lgtm, though it has no effect on downstream users, as a `cargo update` on their end will pull in the latest version.

PR #654, opened 2024-04-16
Add embedded-svc based http transport

madmo:
This small pr adds support for sending envelopes using the embedded-svc http transport available on the esp32 microcontroler.

With this in place it is possible to use sentry in rust (with std) based esp32 projects.

### DIFF
--- a/sentry/src/transports/embedded_svc_http.rs
+++ b/sentry/src/transports/embedded_svc_http.rs
@@ -1,0 +1,64 @@
+use crate::{sentry_debug, ClientOptions, Transport};
+use embedded_svc::http::client::Client as HttpClient;
+use esp_idf_svc::{http::client::EspHttpConnection, io::Write};
+
+/// Transport using the embedded-svc http client
+pub struct EmbeddedSVCHttpTransport {
+    options: ClientOptions,
+}
+
+impl EmbeddedSVCHttpTransport {
+    /// Creates a new transport
+    pub fn new(options: &ClientOptions) -> Self {
+        Self {
+            options: options.clone(),
+        }
+    }
+}
+
+impl EmbeddedSVCHttpTransport {
+    fn send_envelope(
+        &self,
+        envelope: sentry_core::Envelope,
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let dsn = self
+            .options
+            .dsn
+            .as_ref()
+            .ok_or_else(|| "No DSN specified")?;
+        let user_agent = &self.options.user_agent;
+        let auth = dsn.to_auth(Some(user_agent)).to_string();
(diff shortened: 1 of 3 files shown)


sentry/src/transports/embedded_svc_http.rs
```diff
@@ -0,0 +1,64 @@
...
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        let dsn = self
+            .options
+            .dsn
+            .as_ref()
+            .ok_or_else(|| "No DSN specified")?;
+        let user_agent = self.options.user_agent.clone();
+        let auth = dsn.to_auth(Some(&user_agent)).to_string();
+        let headers = [("X-Sentry-Auth", auth.as_str())];
+        let url = dsn.envelope_api_url().to_string();
```
Swatinem, line 32:
```suggestion
        let user_agent = &self.options.user_agent;
        let auth = dsn.to_auth(Some(user_agent)).to_string();
        let headers = [("X-Sentry-Auth", auth.as_str())];
        let url = dsn.envelope_api_url().as_str();
```

Swatinem (approved):
lgtm, left a small suggestion for improvement

Swatinem:
well, our CI runs with `--all-features`, and it looks like that one can’t just simply build this crate on an arbitrary CI system :-(

madmo:
Thanks for the fast response! I will look into adding a sample esp32 project, so that the checks can pass.

madmo:
Had some issues adding a esp32 project to the workspace:

* Would have polluted the workspace root with platform specific files
* Would have required building for esp32, but esp32 does not support all other targets

To get it "working" I made the embedded-svc-http feature do nothing if not compiling for a espidf based target. This has the side effect of effectively disabling CI for espidf.

One solution would be to have a separate ci run with only the embedded-svc-http and a espidf target. Does that sound like an acceptable solution?

Swatinem:
> To get it "working" I made the embedded-svc-http feature do nothing if not compiling for a espidf based target. This has the side effect of effectively disabling CI for espidf.

This is perfect. Given it wouldn’t compile anyway with the feature flag, but without the proper `target_os`, this should be fine.

> One solution would be to have a separate ci run with only the embedded-svc-http and a espidf target. Does that sound like an acceptable solution?

If it can be done with minimal effort, sure. Otherwise, I would rather treat this as a "tier3" platform or however you want to call it. In the sense that it exists, but "we" as the crate maintainers do not guarantee that it won’t break at some point in the future.
We have a similar policy in `sentry-native` for platforms where we have no realistic way to test it ourselves.

madmo:
Keeping it as tier3 is fine for me. If I find some spare time, I will try and create ci setup, but in an other PR.

PR #658, opened 2024-05-23
ref(metrics): Add normalization and update set metrics hashing

elramen:
Add metrics normalization in accordance with [the metrics developer docs](https://develop.sentry.dev/sdk/metrics/#normalization). 
Add metric name, unit, and tag truncation to adhere to [the metrics user docs](https://docs.sentry.io/product/metrics/#maximum-lengths). 
Add `to_envelope` for a single `Metric` instance to facilitate sending metrics from sentry-cli.
Change hash function to crc32 as described [here](https://develop.sentry.dev/sdk/metrics/#sets), this ensures compatibility between different SDKs using the same metric.
Use `clone_from` instead of `clone` and `clone_into` instead of `to_owned` in a couple of places as suggested by clippy.

### DIFF
--- a/sentry-core/src/metrics/mod.rs
+++ b/sentry-core/src/metrics/mod.rs
@@ -54,9 +54,6 @@ use std::sync::{Arc, Mutex};
 use std::thread::{self, JoinHandle};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use normalization::normalized_name::NormalizedName;
-use normalization::normalized_tags::NormalizedTags;
-use normalization::normalized_unit::NormalizedUnit;
 use sentry_types::protocol::latest::{Envelope, EnvelopeItem};
 
 use crate::client::TransportArc;
@@ -534,11 +531,11 @@ impl Metric {
             .as_secs();
         let data = format!(
             "{}@{}:{}|{}|#{}|T{}",
-            NormalizedName::from(self.name.as_ref()),
-            NormalizedUnit::from(self.unit.to_string().as_ref()),
+            normalization::normalize_name(self.name.as_ref()),
+            normalization::normalize_unit(self.unit.to_string().as_ref()),
             self.value,
             self.value.ty(),
-            NormalizedTags::from(&self.tags),
+            normalization::normalize_tags(&self.tags),
             timestamp
         );
         EnvelopeItem::Statsd(data.into_bytes()).into()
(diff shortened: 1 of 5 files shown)


Swatinem (commented):
The code looks correct and has good test coverage, so no problem there.

My main concerns are:
- performance, quite a bit so
- dependencies

At the very least, you should follow the performance best practices of the `regex` crate here: https://docs.rs/regex/latest/regex/#avoid-re-compiling-regexes-especially-in-a-loop

I would consider that a blocker, as you are compiling *the same* `regex` over and over again in every single call.

Both for performance and dependencies: Do we *really* care that tags are limited to N *graphemes*? Why not unicode `char`s or even bytes? As in a bunch of cases (keys mainly?) we are filtering for ASCII chars anyway, iterating over graphemes is absolute overkill.

Otherwise, there is a bunch of `String` allocation happening all over the place which might be avoidable. Those are just a drop in the bucket compared to the `regex` topic mentioned above, so it might not be worth micro-optimizing the last drop of performance out of this.

Speaking of micro-optimization, this very much reminds me of the work I did recently in https://github.com/rust-lang/rust/pull/121150 which I see is about to land today 🎉

sentry-core/src/metrics/mod.rs
```diff
@@ -510,6 +523,24 @@ impl Metric {
...
     }
+
+    /// Convert the metric into an [`Envelope`] containing a single [`EnvelopeItem::Statsd`].
+    pub fn to_envelope(self) -> Envelope {
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs();
+        let data = format!(
+            "{}@{}:{}|{}|#{}|T{}",
```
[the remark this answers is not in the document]
Swatinem: well actually, ignore me :-D I figured out we are defaulting to `none` as unit, and always attach an `environment` tag unconditionally.

Swatinem:
"catching panics" is not really a thing because people can (and often do) use `panic = "abort"`, not to mention that it also has perf overhead.

`floor/ceil_char_boundary` exists in theory but is nightly only:
https://doc.rust-lang.org/std/primitive.str.html#method.floor_char_boundary
The docs also contain a nice example related to graphemes.

Depending on what our goal here is (bytes, chars or graphemes), you might as well just copy over the underlying implementation from `std`, or just iterate over `chars()` which I believe might be the simplest solution here.

loewenheim (commented):
I have one criticism, and it's admittedly nitpicky: I don't think the `From` impls on `NormalizedName/Tags/Unit` are appropriate. Specifically, they are neither lossless nor value-preserving, as per https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from. I think free functions `normalize_name/tags/unit` exported from the `normalization` module would be better here.

loewenheim (approved):

sentry-core/Cargo.toml
```diff
@@ -31,9 +31,12 @@ UNSTABLE_cadence = ["dep:cadence", "UNSTABLE_metrics"]
 
 [dependencies]
 cadence = { version = "0.29.0", optional = true }
+crc32fast = "1.4.0"
+itertools = "0.13.0"
```
Swatinem, line 35: are we still using this?

Swatinem (approved):

sentry-core/Cargo.toml
Swatinem: similar to https://github.com/getsentry/sentry-rust/pull/659, we should make both these dependencies optional.
Also instead of manually sorting the metric tags, how about you switch to a `BTreeMap` which is sorted by definition?

PR #665, opened 2024-06-27
Tower and Actix: Filter out sensitive headers

FSMaxB:
I've noticed that when constructing a `sentry_core::protocol::Request`, all headers are copied even if they are marked as sensitive.

This PR filters out sensitive headers.

### DIFF
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -355,6 +355,7 @@ fn sentry_request_from_http(request: &ServiceRequest, with_pii: bool) -> Request
         headers: request
             .headers()
             .iter()
+            .filter(|(_, v)| !v.is_sensitive())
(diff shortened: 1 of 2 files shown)


Swatinem (approved):

Swatinem:
Thank you, I wasn’t aware so far that some headers are marked as sensitive and there is API for that.

FSMaxB:
I'm not sure whether any are set as sensitive by default, but there is a middleware in `tower-http` that can be used to set headers as sensitive for example: https://docs.rs/tower-http/0.5.2/tower_http/sensitive_headers/index.html

Ten0:
Looks like it might be better to instead write them as "<sensitive>", so that people don't start wondering why their headers aren't sent when they in fact are but are just redacted from the logs.

PR #671, opened 2024-07-19
chore: export `BeforeCallback` type

GuillaumeLagrange:
I needed the type during sentry init for conditional callback depending on context, and ended up having to duplicate this type to pass it around. Figured it would be much better to actually import it but it is currently not possible.

Cheers


Swatinem (approved):

GuillaumeLagrange:
@Swatinem fixed the linting, sorry!

Swatinem (approved):

Issue #677, opened 2024-08-05
`TransportThread::flush` returns incorrect value


manifest:
Description for the [`flush` function of the `Transport` trait](https://github.com/getsentry/sentry-rust/blob/aa17af4ca29d684123b783472759212a74f65963/sentry-core/src/transport.rs#L21) states that ["If the queue was successfully drained, the return value should be `true`."](https://github.com/getsentry/sentry-rust/blob/aa17af4ca29d684123b783472759212a74f65963/sentry-core/src/transport.rs#L19-L20), but implementation of the trait on [`tokio_thread::TransportThread` do the oposite](https://github.com/getsentry/sentry-rust/blob/aa17af4ca29d684123b783472759212a74f65963/sentry/src/transports/tokio_thread.rs#L99). [The same error for `thread::TransportThread`](https://github.com/getsentry/sentry-rust/blob/aa17af4ca29d684123b783472759212a74f65963/sentry/src/transports/thread.rs#L83).

For the user of the library, that means that `Client::flush` and `Client::close` return incorrect values. 

### Expected Result

`TransportThread::flush` returns `true` If the queue was successfully drained.

### Actual Result

`TransportThread::flush` returns `false` If the queue was successfully drained.

manifest:
[PR](https://github.com/getsentry/sentry-rust/pull/678) with a fix for the issue.

manifest:
Fixed in #678

PR #678, opened 2024-08-05
Fix `Transport::flush` implementation for `TransportThread`

manifest:
The PR for the [issue](https://github.com/getsentry/sentry-rust/issues/677).


Swatinem (approved):
nice catch!

PR #687, opened 2024-09-26
feat(sentry-core): add support for w3c traceparent

saiintbrisson:
This change does three things:
1. Add a function to allow creating a `TransactionContext` with a custom Trace ID without a parent span ID.
  My company is testing traces on Sentry in parallel to our current trace solution, for ease of use, we'd like to keep using the same trace ID on both platforms for now. Currently, my approach is to create a `sentry-trace` header with a zeroed span ID, but the UI has this bright red alert telling me that the parent span ID is invalid, and this is bugging me. I doubt I'm the only one who has faced this issue before.
2. Implements support for the `traceparent` as stated in getsentry/team-sdks#41. `sentry-trace` still takes precedence. It does **not** emit the header, though. Should it? It only allows for parsing for now.
3. Creates a builder allowing to customize the creation of a transaction.

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -69,12 +69,15 @@ pub struct TransactionContext {
     op: String,
     trace_id: protocol::TraceId,
     parent_span_id: Option<protocol::SpanId>,
+    span_id: protocol::SpanId,
     sampled: Option<bool>,
     custom: Option<CustomTransactionContext>,
 }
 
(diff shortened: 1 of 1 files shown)


saiintbrisson:
I need to figure some stuff out in #686 before doing this one.

Swatinem (approved):

saiintbrisson:
@Swatinem this one is also ready to be reviewed.

PR #692, opened 2024-10-03
Bump rustls to 0.23.13

qdrs:


Swatinem (approved):
lets see if that works, I remember the `surf` transport requiring some outdated version of some crate, might have been a different one though.

viglia (approved):

PR #696, opened 2024-10-10
Implement record of tagged span data in tracing integration

### LINKED ISSUE #653: sentry-tracing: Fields prefixed with tags. are not sent as tags within spans
### Environment

What version are you running? Etc.

0.32.2

### Steps to Reproduce

1. Create tracing span:

```
tracing::span!(
    tracing::level::INFO
    "request",
    tags.http.url = field::Empty,
    tags.protocol = field::Empty,
    tags.request_id = field::Empty,
    user_agent = field::Empty,
    tags.http.method = field::Empty,
    tags.http.status_code = field::Empty,
)
```

3. For the duration of requests, these fields will be filled with data
4. Unfortunately when I open sentry to see event, it shows no tags and instead all fields are sent as Additional Data

I see in code how it checks for tags presence
https://github.com/getsentry/sentry-rust/blob/1df2e4ca3cc678695bc47e90489c3f83847b9abb/sentry-tracing/src/converters.rs#L155-L179

But it seems not to be applied to transactions?

### Expected Result

See tags in sentry

### Actual Result

Fields are sent as additional data

DoumanAsh:
This is prototype to attach tagged data for tracing spans
Please let me know if it is suitable and I'll see about unit tests if necessary

Fixes #653

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::ops::{Deref, DerefMut};
 use std::sync::{Arc, Mutex, MutexGuard};
@@ -425,6 +426,20 @@ impl<'a> TransactionData<'a> {
             Box::new(std::iter::empty())
         }
     }
+
+    /// Set some extra information to be sent with this Transaction.
+    pub fn set_data(&mut self, key: Cow<'a, str>, value: protocol::Value) {
+        if let Some(transaction) = self.0.transaction.as_mut() {
+            transaction.extra.insert(key.into(), value);
+        }
+    }
+
+    /// Set some extra information to be sent with this Transaction.
+    pub fn set_tag(&mut self, key: Cow<'_, str>, value: String) {
+        if let Some(transaction) = self.0.transaction.as_mut() {
+            transaction.tags.insert(key.into(), value);
+        }
+    }
 }
 
 impl Transaction {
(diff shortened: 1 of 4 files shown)


Swatinem (approved):

viglia (approved):

nrxus:
What's the reason behind putting these methods in `TransactionData` instead of directly in `Transaction`, `Span`, and `TransactionOrSpan`? This way it looks like `TransactionData`'s intent was as an iterator of transaction data not as a way to add more data into it (especially since there is a `set_data` already on the other three structs I mentioned). 

Additionally, it looks like this is a change in behavior? Previously data would be added to the `Span` directly if we were in an inner span (nor the root transaction) but now data will be added to the transaction directly?

DoumanAsh:
> What's the reason behind putting these methods in TransactionData

Because acquiring lock every time you need to add data point is pointless

> Additionally, it looks like this is a change in behavior?

There is no change in behavior, data is added in the same places as before

DoumanAsh:
Is this crate still maintained?

Swatinem:
We often "abused" doctests to act as unit tests for a lot of things. Even though it would be good to have tests for this, I can also merge this without to speed things up here a bit.

DoumanAsh:
Is there any utility to create some sort of dummy/test setup for sentry in order to create transactions and read its data later?
I'm wondering if there are some tests exists that check transaction creation via global hub?

Swatinem:
There is a test utility that will create a new sentry with a testing transport that captures all the envelopes/events happening within a scope. You should be able to use that.

DoumanAsh:
Created PR with basic test as follow up https://github.com/getsentry/sentry-rust/pull/703

PR #699, opened 2024-10-18
Allow settings tags directly on transactions/spans

nrxus:
currently the only tags added to transactions are the tags on the scope at the time you finish a transaction. Sometimes, however, it is useful to have tags that aren't for the scope but just for the transaction. 

I am using the same signature as the equivalent method in `Scope`. 

I know there are currently other PRs related to tags in transactions but I think those are more specific to the tracing integration so having methods  to do it explicitly in a transaction (or span) is still useful.

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -260,6 +260,14 @@ impl TransactionOrSpan {
         }
     }
 
+    /// Sets a tag to a specific value.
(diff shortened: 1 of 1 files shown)


Swatinem (approved):

Swatinem:
Ah yes, this slipped my radar, as I don’t think anyone at the moment really maintains this repo/SDK actively.

PR #700, opened 2024-10-21
Refactoring tracing span activation/deactivation

### LINKED ISSUE #675: `tracing::Instrument::instrument`ed futures don't propagate to Sentry
It seems that the `sentry_tracing` `Layer` doesn't take into account the `on_enter` event provided by `tracing` for `Instrumented` futures:
https://github.com/getsentry/sentry-rust/blob/8cc4848f115152af6f2ad66d7aeb17bd3bbe38fd/sentry-tracing/src/layer.rs#L153

This implies that Sentry scopes are not properly entered and exited as `tracing` scopes are entered/exited.

It seems that the `Span`'s existence and `finish` should be controlled by the existing `on_new_span`/`on_close`, but scope configuration should be tied to `on_enter`/`on_exit` instead of also being controlled by `on_new_span` and `on_close`.

DoumanAsh:
This is proposal how to improve tracing layer to handle concurrency better

This changes logic of transaction activation/deactivation from being tied to ctor/dtor and instead tied to enter/exit

The reason being is that if you have multiple transactions running concurrently, there is no way to distinguish for sentry this scenario as it just creates new child transaction every time parent is present

Partially Fixes #675

### DIFF
--- a/sentry-tracing/src/layer.rs
+++ b/sentry-tracing/src/layer.rs
@@ -268,8 +268,6 @@ where
         // This comes from typically the `fields` in `tracing::instrument`.
         record_fields(&sentry_span, data);
 
-        sentry_core::configure_scope(|scope| scope.set_span(Some(sentry_span.clone())));
-
         let mut extensions = span.extensions_mut();
         extensions.insert(SentrySpanData {
             sentry_span,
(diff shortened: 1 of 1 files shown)


DoumanAsh:
Closed because #686 already fixes

PR #701, opened 2024-10-23
fix: Don't block on sending envelopes (ureq/curl transports)

pjenvey:
This is a followup to #546 (for #543):

Any transport is susceptible to #543, where sending too many events to Sentry can block. So the `thread::TransportThread` impl should similarly drop events instead of blocking, just as the similar `tokio_thread::TransportThread` began doing with the fix in #546.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**:
+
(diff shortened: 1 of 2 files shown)


Swatinem (approved):

viglia (approved):

pjenvey:
@Swatinem or @viglia could you get this merged? Thank you!

PR #708, opened 2024-11-19
feat(sentry-tracing): add span fields to breadcrbums

thomaseizinger:
The fields of a span are equally helpful in an event as well as in the breadcrumbs leading up to it. In fact, in an application that makes heavy use of spans to add contextual information to messages, the log messages themselves without the span fields are often almost useless.


thomaseizinger:
Unfortuantely, this is only one half of the story. We also need to find a solution to https://github.com/getsentry/sentry-rust/issues/617#issuecomment-2487058619 otherwise only spans that have been sampled are present.

Swatinem (approved):

Swatinem:
I think this needs a rebase

thomaseizinger:
> I think this needs a rebase

Done!

thomaseizinger:
> Unfortuantely, this is only one half of the story. We also need to find a solution to [#617 (comment)](https://github.com/getsentry/sentry-rust/issues/617#issuecomment-2487058619) otherwise only spans that have been sampled are present.

Any thoughts on this? I did some digging through the codebase and I think we could solve this by sampling spans _later_, i.e. keep them around initially and only sample and discard them when they are about to be submitted to Sentry.

The problem with the current approach is that the code added in this PR does not do anything unless one also samples all spans. That on the other hand though can create A LOT of spans in Sentry which may not be desired. Ideally, I think the two features of sampling spans to be sent to Sentry and capturing the fields within spans for breadcrumbs and events are orthogonal.

Swatinem:
A test for `tracing` was recently added here: https://github.com/getsentry/sentry-rust/commit/00cb40faa5a31d16e033e257c0d482be256923f0
You might be able to use the `with_captured_envelopes_options` function: https://github.com/getsentry/sentry-rust/blob/05fc7de25ed545bb5ff9348f586eb6ef0d535bc3/sentry-core/src/test.rs#L120

Otherwise I’m a bit out of the loop when it comes to how the SDK APIs and are supposed to work :-(

PR #709, opened 2024-11-19
fix: disable rustls features by default

thomaseizinger:
As of rustls 0.23, `aws-lc-rc` is the default crypto backend of rustls. In order to allow binaries to choose, which crypto backend to use, libraries need to disable rustls' default features.


Swatinem (approved):

thomaseizinger:
The CI failure appears unrelated to my changes. That is likely broken on `main` too?

Swatinem:
Yes it most likely is due to a Rust update.

Swatinem:
Thank you for taking care of this ❤️

Issue #706, opened 2024-11-11
Allow `set_user` to update current user instead of replacing it


thomaseizinger:
In our application, not all user-identifying pieces of information are known at the same time. We would like to use `set_user` to add as much information as we can to the user context.

At present, `sentry` only allows the whole user to be set at once, meaning we have to track that information separately and update the user context every time we learn new information.

The simplest way to solve this would be if we could expose a `user` getter. That would allow the application to see, what the current user is and update it with the existing fields and add new ones to it. Alternatively, a `user_mut` could be exposed as well. If that would return `&mut Option<User>` then one could use the `get_or_insert` family of APIs to set / update the current user.

PR #715, opened 2024-12-19
feat: allow retrieving the current user of a scope

thomaseizinger:
In our application, not all information of the current user is known immediately upfront. The user ID is loaded directly on startup but the account that they log in to is only known later.

Currently, we have to separately track this information because we always have to set the entire user at once. To avoid this, this PR adds a `user` getter to get the current user from the scope. This solves our use-case because we can get the current user and produce a new one that has additional fields and set it again.

Resolves: #706.

### DIFF
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -197,6 +197,11 @@ impl Scope {
         self.user = user.map(Arc::new);
     }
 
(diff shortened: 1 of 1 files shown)


thomaseizinger:
CI failures are addressed in https://github.com/getsentry/sentry-rust/pull/716.

Swatinem (approved):

PR #718, opened 2025-01-02
sentry-tower: Update `axum` dependency to v0.8

Turbo87:
see https://tokio.rs/blog/2025-01-01-announcing-axum-0-8-0


Swatinem (approved):
can you check why axum 0.7 is still being pulled into the dependencies?

Turbo87:
> can you check why axum 0.7 is still being pulled into the dependencies?

`tonic` is (optionally) used by `sentry-tower` and that is using `axum`, but unfortunately they haven't released a version with axum 0.8 support yet. is there a better way to handle this situation?

Swatinem:
Probably not in that case.

PR #719, opened 2025-01-02
fix: Ensure Send+Sync bounds on SentryLayer and SentryService

rubdos:
Updates PhantomData usage in SentryLayer and SentryService to explicitly ensure Send+Sync trait bounds using the fn pointer pattern as documented in the Rust Nomicon.

Sync axum 0.8 (also, see #718), the [Router layer call seems to require `Sync` bounds](https://github.com/tokio-rs/axum/commit/52fd139a8).

https://github.com/tokio-rs/axum/pull/2473

### DIFF
--- a/sentry-tower/src/lib.rs
+++ b/sentry-tower/src/lib.rs
@@ -196,7 +196,9 @@ where
     H: Into<Arc<Hub>>,
 {
     provider: P,
-    _hub: PhantomData<(H, Request)>,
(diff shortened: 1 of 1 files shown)


jplatte (approved):

Swatinem (approved):
Adding a test like in #721 would be nice

Swatinem:
I went with #721 instead, as it was ready to merge. Thanks for your contribution nonetheless :-)

PR #731, opened 2025-02-02
feat(actix): capture HTTP request body

pacifistes:
## Summary
This PR adds the ability to capture HTTP request bodies in the Sentry Actix middleware, allowing for better debugging and error tracking capabilities. The feature is configurable through the `max_request_body_size` option in `ClientOptions`.

## Changes
- Added `MaxRequestBodySize` enum to control request body capture behavior:
  - `None`: Don't capture request bodies (default)
  - `Small`: Capture up to 1000 bytes
  - `Medium`: Capture up to 10000 bytes
  - `Always`: Capture entire body
- Added request body capture support for `JSON` and `form-urlencoded` content types
- Implemented body capture and restoration logic to maintain request integrity
- Added necessary dependencies: serde_json, actix-http

## Implementation Details
- Request bodies are only captured for content types `application/json` and `application/x-www-form-urlencoded`
- The body is read from the request stream and then restored to maintain middleware transparency
- Size limits are enforced based on the configured `max_request_body_size` option
- Empty string is returned for bodies exceeding size limits


sentry-core/src/clientoptions.rs
```diff
@@ -40,6 +40,24 @@ pub enum SessionMode {
     Request,
 }
 
+/// The maximum size of an HTTP request body that the SDK captures.
+///
+/// Only request bodies that parse as JSON or form data are currently captured.
+/// See the [Documentation on attaching request body](https://develop.sentry.dev/sdk/expected-features/#attaching-request-body-in-server-sdks)
+/// and the [Documentation on handling sensitive data](https://develop.sentry.dev/sdk/expected-features/data-handling/#sensitive-data)
+/// for more information.
+#[derive(Clone, PartialEq)]
```
Swatinem, line 49: This could probably be `Copy` and `Default`.

Also, another variant with `Explicit(usize)` would also be nice.

It would also be nice if this enum had a `len()` method or similar defined right on the enum.
That way, the actual size limit is close to the enum and its documentation, and not in a completely different part of the code.

sentry-actix/src/lib.rs
```diff
@@ -191,21 +196,63 @@ where
...
 }
 
 /// The middleware for individual services.
 pub struct SentryMiddleware<S> {
-    service: S,
+    service: Rc<S>,
     inner: Sentry,
 }
 
+async fn should_capture_request_body(headers: &HeaderMap) -> bool {
```
Swatinem, line 211: this doesn’t need to be async

sentry-actix/src/lib.rs
```diff
@@ -191,21 +196,63 @@ where
...
+    Ok::<_, PayloadError>(body)
+}
+
+async fn capture_request_body(
+    req: &mut ServiceRequest,
+    max_size: MaxRequestBodySize,
+) -> Option<String> {
+    let request_body = body_from_http(req).await.unwrap();
+    let request_body_str = String::from_utf8_lossy(&request_body).into_owned();
+
```
Swatinem, line 243: you should do the size check before you are doing the utf-8 conversion. Otherwise this is just wasted work in case it fails the size-check.

sentry-actix/src/lib.rs
```diff
@@ -191,21 +196,63 @@ where
...
+/// Extract a body from the HTTP request
+async fn body_from_http(req: &mut ServiceRequest) -> Result<BytesMut, PayloadError> {
+    let mut stream = req.take_payload();
+
+    let mut body = BytesMut::new();
+    while let Some(chunk) = futures_util::StreamExt::next(&mut stream).await {
+        let chunk = chunk?;
+        body.extend_from_slice(&chunk);
+    }
```
Swatinem, line 229: I don’t think capturing the whole payload ahead of invoking the actual handler is not a good idea.

Although this is less likely for json / urlencoded payloads, the downstream middleware might expect to process the body in a streaming fashion. Plus, the body might be *huge*, and buffering it into memory might cause problems.
Again, this is highly unlikely to be a concern for json bodies though.

Swatinem:
Another question would be: Is the intention to capture *up to* the configured limit, or to *discard* any payloads that exceed the limit.

sentry-actix/src/lib.rs
```diff
@@ -191,21 +196,63 @@ where
...
+
+async fn capture_request_body(
+    req: &mut ServiceRequest,
+    max_size: MaxRequestBodySize,
+) -> Option<String> {
+    let request_body = body_from_http(req).await.unwrap();
+    let request_body_str = String::from_utf8_lossy(&request_body).into_owned();
+
+    match max_size {
+        MaxRequestBodySize::Small if request_body_str.len() < 1_000 => Some(request_body_str),
```
lcian, line 245: Let's also make these inclusive as e.g. 1k is supposed to be the max, not 999

lcian:
> Another question would be: Is the intention to capture _up to_ the configured limit, or to _discard_ any payloads that exceed the limit.

The goal here is to discard payloads that exceed the limit. In some SDKs (e.g. Python) there is some more complex logic to handle payloads that exceed the max but in others (e.g. PHP) it's simply discarded.
So for now I think we can go for the simpler option.

sentry-actix/src/lib.rs
Swatinem: that sounds reasonable

sentry-actix/src/lib.rs
pacifistes: You are right, I changed it.

sentry-core/src/clientoptions.rs
pacifistes: Thanks for the feedback! I've added the `Explicit` variant and created an `is_within_size_limit` method on the enum to keep the size validation logic close to its definition.

Regarding `Copy` - I kept `Clone` because it's required by `ClientOptions`. If you know a way to make it work with `Copy`, I'm open to suggestions.

sentry-actix/src/lib.rs
pacifistes: It's Done by the new implementation of `should_capture_request_body` method now checks the `Content-Length`, `Content-Type`, and ensures that `transfer_encoding` does not include `chunked`. However, with this update, I no longer retain the empty string when the body is too large, meaning it won’t be displayed in the Sentry issue. Should I modify this to keep the empty string in those cases?

sentry-actix/src/lib.rs
pacifistes: Fixed inside the new method `is_within_size_limit`.

sentry-actix/src/lib.rs
```diff
@@ -191,21 +196,72 @@ where
...
 }
 
 /// The middleware for individual services.
 pub struct SentryMiddleware<S> {
-    service: S,
+    service: Rc<S>,
     inner: Sentry,
 }
 
+async fn should_capture_request_body(
```
Swatinem, line 211: this does not need to be `async`

sentry-core/src/clientoptions.rs
Swatinem: You can make this `Copy` independent of the `ClientOptions`.

sentry-actix/src/lib.rs
```diff
@@ -180,7 +185,7 @@ impl Default for Sentry {
 
 impl<S, B> Transform<S, ServiceRequest> for Sentry
 where
-    S: Service<ServiceRequest, Response = ServiceResponse<B>, Error = Error>,
+    S: Service<ServiceRequest, Response = ServiceResponse<B>, Error = Error> + 'static,
```
Swatinem, line 188: Why exactly is the `'static` bound and the `Rc` needed?
If it is indeed unavoidable, better to use `Arc`, as we are talking about (potentially multithreaded) servers here.

sentry-actix/src/lib.rs
pacifistes: Oups did it again :grimacing: . Fixed.

sentry-actix/src/lib.rs
```diff
@@ -258,21 +318,27 @@ where
...
-            }
-            scope.add_event_processor(move |event| Some(process_event(event, &sentry_req)));
-            parent_span
-        });
+        let svc = self.service.clone();
+        async move {
+            let mut req = req;
 
-        let fut = self.service.call(req).bind_hub(hub.clone());
+            if max_request_body_size != MaxRequestBodySize::None
```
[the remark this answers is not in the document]
pacifistes: True. I changed it.

sentry-actix/src/lib.rs
```diff
@@ -191,21 +196,72 @@ where
...
+    }
+    let (_, mut orig_payload) = actix_http::h1::Payload::create(true);
+    orig_payload.unread_data(body.clone().freeze());
+    req.set_payload(actix_web::dev::Payload::from(orig_payload));
+
+    Ok::<_, PayloadError>(body)
+}
+
+async fn capture_request_body(req: &mut ServiceRequest) -> String {
+    let request_body = body_from_http(req).await.unwrap();
```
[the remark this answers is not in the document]
pacifistes: From what i see the different errors are: 
```rust
/// A set of errors that can occur during payload parsing.
#[derive(Debug, Display)]
#[non_exhaustive]
pub enum PayloadError {
    /// A payload reached EOF, but is not complete.
    #[display(fmt = "payload reached EOF before completing: {:?}", _0)]
    Incomplete(Option<io::Error>),

    /// Content encoding stream corruption.
    #[display(fmt = "can not decode content-encoding")]
    EncodingCorrupted,

    /// Payload reached size limit.
    #[display(fmt = "payload reached size limit")]
    Overflow,

    /// Payload length is unknown.
    #[display(fmt = "payload length is unknown")]
    UnknownLength,

    /// HTTP/2 payload error.
    #[cfg(feature = "http2")]
    #[display(fmt = "{}", _0)]
    Http2Payload(::h2::Error),

    /// Generic I/O error.
    #[display(fmt = "{}", _0)]
    Io(io::Error),
}
```
So i think it would be really unlikely that the method `body_from_http` would return an error. More importantly, if we do get an error, we've already consumed the request payload via `take_payload()`. At this point, the request would be in an invalid state anyway and couldn't be processed correctly.
but again, i think it would have fail whatever the case if we have error during the payload extraction. I can put the unwrap on the chunk if you think it's better, or maybe return an empty string if there is an error 🤷‍♂️ .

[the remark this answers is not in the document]
lcian: Let's not use unwrap but set `request_body` to empty string if this errors. The SDK should never panic.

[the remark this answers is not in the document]
pacifistes: Done

sentry-actix/src/lib.rs
pacifistes: I needed the RC + 'static because otherwhise i was getting lifetime issues. Wasn't 100% sure if i should use `Rc` or `Arc` after u asked so i ask on the Actix discord and the lead of Actix confirm that `Rc` was the right choice.
source: https://discord.com/channels/771444961383153695/1337163852118491136

lcian (approved):

lcian:
Setting the option to `Medium` by default as that's what all of the other SDKs do, the develop docs were outdated.

lcian:
@pacifistes thanks for the contribution! 😄

Issue #733, opened 2025-02-03
Document how to attach data onto a span/transaction


lcian:
Document how a user can add arbitrary `data` attributes to a span & transaction.
As almost all SDKs, besides JS, do not yet expose a `setAttribute` API, we want to document `setData` instead.

This is important, as we constantly add more users to the Trace Explorer Beta.

lcian:
The SDK does not support `scope.get_transaction` and therefore, compared to the other SDKs, it's not possible to retrieve the currently active transaction to set data attributes on it.
For now, we will just document what is possible with the current API.

PR #738, opened 2025-02-05
feat(core): add `scope.get_transaction` API

lcian:
This adds a `scope.get_transaction` API to get the currently active transaction from a Scope.
This will make it possible for customers to get the currently active transaction to add data attributes to it, which can then be used for queries in the Trace Explorer.
It also aligns the SDK with other SDKs that have this API.
Needed for https://github.com/getsentry/sentry-rust/issues/733

### DIFF
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -3,6 +3,7 @@ use std::collections::{HashMap, VecDeque};
 use std::fmt;
 use std::sync::{Arc, Mutex, PoisonError, RwLock};
 
+use crate::performance;
 use crate::performance::TransactionOrSpan;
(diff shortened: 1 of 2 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -341,6 +341,14 @@ impl Scope {
         self.span.as_ref().clone()
     }
 
+    /// Returns the currently active transaction.
+    pub fn get_transaction(&self) -> Option<TransactionOrSpan> {
```
szokeasaurusrex, line 345:
```suggestion
    pub fn get_transaction(&self) -> Option<Transaction> {
```

`get_transaction` should always return a transaction, so we should enforce this through the type system.

szokeasaurusrex (commented):
see suggestions before merging

PR #739, opened 2025-02-05
feat(core): transaction.set_data sets data on TraceContext

lcian:
The Rust SDK currently sets data on `transaction.extra` when calling `transaction.set_data` on a `Transaction`.
This change makes it so that data is instead set on the `TraceContext` i.e. `transaction.context.trace.data` as other SDKs do (see https://github.com/getsentry/team-sdks/issues/95). 

Required for https://github.com/getsentry/sentry-rust/issues/733

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -526,32 +526,32 @@ pub struct Transaction {
     pub(crate) inner: TransactionArc,
 }
 
-/// Iterable for a transaction's [`extra` field](protocol::Transaction::extra).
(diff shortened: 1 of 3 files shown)


lcian:
Changed existing APIs that access/mutate Transaction data (used in the `tracing` integration) to change data attributes (`TraceContext.data`) instead of `transaction.extra`.
Changed comments to reflect the new naming convention of "data attributes" to refer to `TraceContext.data`.

Swatinem (approved):
lgtm! its good to have someone who actually knows how all of this is supposed to work within SDKs :-)

PR #746, opened 2025-03-01
Add option to enable propagation of parent span tags

DoumanAsh:
I would like to suggest option to allow event's spans to automatically insert first available tag as tag within event, rather than propagating it as additional data

Old behavior can be used together with this feature too

But I often find it inconvenient that events may not contain tags from within span (especially if spans get rate limited at sentry)
What do you think about this addition?
Let me know if you'd like me to make some adjustments to PR

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -625,6 +625,22 @@ pub struct Transaction {
     pub(crate) inner: TransactionArc,
 }
 
+/// Iterable for a transaction's [tags attributes](protocol::Transaction::tags).
+pub struct TransactionTags<'a>(MutexGuard<'a, TransactionInner>);
+impl<'a> TransactionTags<'a> {
+    #[inline]
+    /// Returns iterator over available tags, if transaction is sampled
+    pub fn iter(&'a self) -> Option<impl Iterator<Item = (&'a String, &'a String)> + 'a> {
+        self.0.transaction.as_ref().map(|tx| tx.tags.iter())
+    }
+
+    #[inline]
+    /// Converts self into data pointer
+    pub fn into_data(self) -> TransactionData<'a> {
+        TransactionData(self.0)
+    }
+}
+
 /// Iterable for a transaction's [data attributes](protocol::TraceContext::data).
 pub struct TransactionData<'a>(MutexGuard<'a, TransactionInner>);
 
@@ -744,6 +760,12 @@ impl Transaction {
     /// a deadlock.
     pub fn data(&self) -> TransactionData {
         TransactionData(self.inner.lock().unwrap())
+    }
+
+    /// Returns an iterating accessor to the transaction's
+    /// [tags attributes](protocol::Transaction::tags).
+    pub fn tags(&self) -> TransactionTags {
+        TransactionTags(self.inner.lock().unwrap())
     }
 
     /// Get the TransactionContext of the Transaction.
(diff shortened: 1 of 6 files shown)


lcian:
Hi @DoumanAsh , thanks for the PR and sorry for the wait here!
Could you please provide an example that shows the difference in behavior with `SpanPropagation::Attributes` and `SpanPropagation::All`?

DoumanAsh:
Would test case suffice or what sort of example you'd like to see?

DoumanAsh:
Actually I kinda forgot about it, but looking back at implementation, event stores fields in context `Rust Tracing Fields`
Would it be able to store tags at all...

lcian:
No problem. 
I would like to see an example using top level APIs of `sentry-rust`/`tracing` and how the result is different in the envelope/UI.
I tried some stuff for myself combining the span/message APIs of both and using both attributes and tags in my spans but I was never able to hit this code path: https://github.com/DoumanAsh/sentry-rust/blob/7344609dcbfee4c9d855a72ad28a40ecda390d5c/sentry-tracing/src/converters.rs#L111C1-L115C18

I don't clearly see how, but maybe this might be affected by https://github.com/getsentry/sentry-rust/commit/3bcdda39ba4c0ad3e6cdce60a0ce1f8e224ce644
In that commit we've changed `transaction.data()` and `transaction.set_data()` to operate on the `trace context` instead of `extra` (which is correct from the point of view of SDKs)

DoumanAsh:
@lcian Actually I know what is problem is and it is very trivial mistake on my part
But my PC is broken right now so I will try to get to it as soon as I can to get this PR working and will add test to showcase tags

lcian:
> @lcian Actually I know what is problem is and it is very trivial mistake on my part But my PC is broken right now so I will try to get to it as soon as I can to get this PR working and will add test to showcase tags

Nice, thanks!

DoumanAsh:
@lcian I refactored my code to correctly look up span's tags on new event (this of course not possible for breadcrumbs, but events/exceptions should be working as it carries tags)
https://github.com/DoumanAsh/sentry-rust/blob/span_propagate_tags/sentry-tracing/tests/smoke.rs#L22

DoumanAsh:
It is kinda not really used without context so I made it together
But if you prefer it separately sure

DoumanAsh:
Rebased and refactored context passing
https://github.com/getsentry/sentry-rust/pull/746/commits/0e59852946c9b631880c6f786923a2c0593389b9

DoumanAsh:
@lcian do you plan to merge it?

lcian:
Thanks for the PR and your patience @DoumanAsh.
I've looked into it just now.

In general we don't provide this kind of API in other SDKs and personally I find the `FieldVisitor` approach hard to reason about.

What I think should be merged is an API to be able to get tags from a `Span` or `Transaction` (and also from the "top-level" `TransactionOrSpan`). Right now we just have `set_tag` on those types but no way to retrieve the tags you set.

Imagining we provide that API, I believe you would be able to accomplish your usecase with a `before_send` similar to this:
```rust
            before_send: Some(Arc::new(|mut event| {
                if let Some(span) = sentry::configure_scope(|scope| scope.get_span()) {
                     for (key, value) in span.tags().iter() {
                         event.tags.insert(key.into(), value.into());
                     }
                }
                Some(event)
            })),
```

Please let me know if that works for you. 
One thing I haven't reasoned about is what would happen with nested spans with my approach, not sure if it would match exactly the behavior implemented in this PR.

DoumanAsh:
Very well, I will need to look into it, but I probably will prefer to have proper span's tags propagation so I will just use my own fork for sentry

PR #749, opened 2025-03-11
feat: introduce `release-health` feature

### LINKED ISSUE #748: Gate release health behind a feature flag
Some targets (e.g. `wasm32-wasip1`) don't support threads.
By default, the SDK is compiled with session release health features: `sentry-core/src/sessions.rs`.
We should create a feature flag that enables/disables these features so that the SDK can be compiled for those targets by disabling the flag (similar to how a user can opt to create a custom transport and use none of the existing ones).
The feature should be enabled by default to respect existing behavior.

Features implemented in `sentry-core/src/sessions.rs` are exposed in the `start_session` function in `sentry-core/src/hub.rs` and `sentry-core/src/api.rs`. The API is also used in `sentry/src/init.rs` to enable session tracking by default. These usages should also be gated behind the new feature flag.

pepperoni505:
Resolves #748 by introducing a new feature enabled by default named `release-health`


sentry-core/src/client.rs
```diff
@@ -60,10 +62,15 @@ impl fmt::Debug for Client {
...
     fn clone(&self) -> Client {
         let transport = Arc::new(RwLock::new(self.transport.read().unwrap().clone()));
+
+        #[cfg(feature = "release-health")]
         let session_flusher = RwLock::new(Some(SessionFlusher::new(
             transport.clone(),
             self.options.session_mode,
         )));
+        #[cfg(not(feature = "release-health"))]
+        let session_flusher = RwLock::new(None);
```
Swatinem, line 72: it might be a bit more work, but avoiding the `Client.session_flusher` field altogether would be slightly more efficient.

Swatinem (approved):

PR #754, opened 2025-03-19
chore(msrv): `cargo update` and bump MSRV to 1.81

lcian:
The MSRV is actually 1.81, I didn't run `cargo update` the last time :smile:


Swatinem (approved):
In general I’m 👍🏻 on bumping msrv.
Though the very latest cargo with `resolver=3` is "msrv aware", meaning that it picks msrv compatible versions when doing a `cargo update`, etc.

Though that latest cargo version is 1.85, and requires that explicit resolver setting for workspaces, or the 2024 edition for crates, which I think we are still quite far away from.

PR #755, opened 2025-03-19
fix(tracing): wrap error with synthetic mechanism only if attaching stacktrace

### LINKED ISSUE #723: Errors reported through `sentry-tracing` have unhelpful titles
### Environment

**Version:** `0.36.0`

Sentry set up with the `sentry-tracing` integration.

### Steps to Reproduce

1. Followed instructions in the docs to track errors using the `tracing` macros: https://docs.rs/sentry-tracing/0.36.0/sentry_tracing/#tracking-errors
2. Logged an error along the lines of `tracing::error!(error = &custom_error as &dyn Error, "my operation failed - ...");`
    1. tried with `let custom_error = std::io::Error::new(std::io::ErrorKind::Other, "Custom ioErrorKind Other Error");`
    2. tried with custom `thiserror` based error struct
3. Logged a "control" error using `error!("my operation failed - simple error!(\"my operation failed\") has helpful title");`

### Expected Result

Errors to show up in Sentry with a useful title, either based on the original message, or some version of the error struct name

### Actual Result

In the Sentry dashboard, the errors reported using the `error =` capturing method had titles that had nothing to do with the error but rather referenced the `tracing` crate. Errors reported without the `error = ` capturing method show up with the message in the title.

Image

lcian:
### Description

With this change, we wrap an `Error` captured with e.g. `tracing::error` with the synthetic mechanism for `tracing` only if we effectively have a stack trace to be reported in the error, i.e. feature `backtrace` is enabled and `attach_stacktrace` is enabled.

Without this change, using `tracing::error!(error = <err>, <message>)` without the `backtrace` feature enabled would result in Sentry issues with `<unknown>` as title, as shown in https://github.com/getsentry/sentry-rust/issues/723#issuecomment-2642497057.

Closes #723 

### Example

```rust
fn a() {
    let err = "NaN".parse::<u32>().unwrap_err();
    tracing::error!(error = std_dyn_err(&err), "error with error");
}

fn b() {
    let err = File::open("non_existent_file.txt").unwrap_err();
    tracing::error!(error = std_dyn_err(&err), "error with error");
}

fn c() {
    let err = File::open("non_existent_file.txt").unwrap_err();
    sentry::capture_error(&err);
}
```

### Before

Screenshot 2025-03-19 at 17 51 22

### After

Screenshot 2025-03-19 at 19 41 49

`b` and `c` will now be grouped under a single issue. This makes sense as they are carrying an equivalent `Error`. In absence of stack frames, we cannot distinguish between the two.
`b` will still include the additional message, the `mechanism: tracing` indicator and the `Rust Tracing Context`.
Screenshot 2025-03-19 at 17 41 32

While `c` will not include this extra info as usual.

Screenshot 2025-03-19 at 17 41 50

### DIFF
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeMap;
 use std::error::Error;
 
-use sentry_core::protocol::{Event, Exception, Mechanism, Thread, Value};
+use sentry_core::protocol::{Event, Exception, Mechanism, Value};
 use sentry_core::{event_from_error, Breadcrumb, Level, TransactionOrSpan};
 use tracing_core::field::{Field, Visit};
 use tracing_core::Subscriber;
@@ -21,6 +21,7 @@ fn convert_tracing_level(level: &tracing_core::Level) -> Level {
     }
 }
 
+#[allow(unused)]
 fn level_to_exception_type(level: &tracing_core::Level) -> &'static str {
     match *level {
         tracing_core::Level::TRACE => "tracing::trace!",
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
🎉 nice! this has bothered me for a long time for the internal symbolicator errors we have. I remember this broke at some point because of messing with these fields, and we never figured out how to fix that.

PR #756, opened 2025-03-20
ref(backtrace): add entries and extra logic for in-app detection

### LINKED ISSUE #717: `in_app_exclude` does not work as intended.
### Environment

What version are you running? Etc.
0.34
### Steps to Reproduce

1. To exclude some of the generic crates that are not related to the core code path I was using `in_app_exclude ` but it does not seem to exclude anything. On the docs it says: https://docs.rs/sentry/0.17.0/sentry/struct.ClientOptions.html#structfield.in_app_include:~:text=in_app_exclude%3A%20Vec,are%20never%20in_app
I tried something like this:

```
in_app_exclude: vec![
            "backtrace",
            "futures",
            "futures_core",
            "hyper",
            "tokio",
            "tracing",
            "warp",
 ],
```

But it still does not exclude them for the stack trace. 

### Actual Result
Image

This has been really annoying when trying to find the root cause of the issue. We prefer to use the sdk and not the issue grouping tab. Seems like a bug in the code?

lcian:
# Description

Closes #717

### DIFF
--- a/sentry-backtrace/src/utils.rs
+++ b/sentry-backtrace/src/utils.rs
@@ -95,8 +95,9 @@ pub fn demangle_symbol(s: &str) -> String {
 
 /// Checks whether the function name starts with the given pattern.
 ///
-/// In trait implementations, the original type name is wrapped in "_< ... >" or "<F as ...>"
-/// and colons are replaced with dots. This function accounts for differences while checking.
+/// In trait implementations, the original type name is wrapped in "_< ... >" and colons are
+/// replaced with dots. This function accounts for differences while checking.
+/// The `<F as ` pattern is a special case that can often be observed in frames involving futures-rs traits.
 pub fn function_starts_with(mut func_name: &str, mut pattern: &str) -> bool {
     if pattern.starts_with('<') {
         while pattern.starts_with('<') {
--- a/sentry-backtrace/src/utils.rs
+++ b/sentry-backtrace/src/utils.rs
@@ -95,14 +95,16 @@ pub fn demangle_symbol(s: &str) -> String {
 
 /// Checks whether the function name starts with the given pattern.
 ///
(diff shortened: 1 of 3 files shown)


sentry-backtrace/src/utils.rs
```diff
@@ -111,7 +114,10 @@ pub fn function_starts_with(mut func_name: &str, mut pattern: &str) -> bool {
             }
         }
     } else {
-        func_name = func_name.trim_start_matches('<').trim_start_matches("_<");
+        func_name = func_name
+            .trim_start_matches("<F as ")
```
lcian, line 118: Specific case that can be observed in some raw frames related to `futures_core`, `futures_util`, `axum`, `hyper`, and others
It's specific for those cases as in general we will see the form `<X as Y>` for trait impls but I don't see any downside with adding this.

Swatinem (approved):
detecting this in the SDK requires these symbols to be present at runtime, which is often not the case when one is using a stripped release build.

Ideally the border frame should be added to our default stack trace enhancement rules that are part of serverside grouping, and are also applied after serverside symbolication.

lcian:
> detecting this in the SDK requires these symbols to be present at runtime, which is often not the case when one is using a stripped release build.
> 
> Ideally the border frame should be added to our default stack trace enhancement rules that are part of serverside grouping, and are also applied after serverside symbolication.

I'll do some more research and submit a PR to add more frames there.
There's at least one rule missing as currently when the SDK sends `rust_begin_unwind` as in app it's not overridden by the backend.

Just to document the change, the reason it's sent as in app right now is because in `trim_stacktrace` we trim on the first match.
If could happen that there's both `std::sys::backtrace::__rust_end_short_backtrace` and `core::panicking::panic_fmt` but we trim on the first of the two, leaving `rust_begin_unwind` as in_app between the two.
Screenshot 2025-03-20 at 16 45 44

Issue #745, opened 2025-02-28
anyhow errors are grouped incorrectly


danechitoaie:
### Environment
Sentry 0.36.0

Any idea would could cause different errors logged from different parts of the app with different messages (using `anyhow::anyhow!()`) to be incorrectly grouped?

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/d43088e6-e61a-4f30-9414-6deddd4991af" />

See the above print screen. 
Why is that error grouped with the others when it's different error.

One was logged with 
```rust
let err_msg = "Embeddings request failed!";
let err_anyhow = anyhow::anyhow!(err_msg);
sentry::integrations::anyhow::capture_anyhow(err_anyhow);
```

and the other with
```rust
let err_msg = "Create thread request failed!";
let err_anyhow = anyhow::anyhow!(err_msg);
sentry::integrations::anyhow::capture_anyhow(err_anyhow);
```

Can this be because both errors are anyhow errors just with different messages? And somehow they have same fingerprint and Sentry groups them by that?

Would me setting a custom fingerprint help?

lcian:
Hey @danechitoaie, could you try scrolling to the bottom of the issue page? There will be a section you can expand called "Event Grouping Information" which will give you more information around this.
Feel free to share it here if you need more help to understand it or if you think we might have a bug :)

danechitoaie:
Hi.
Trying to provide more details in a simplified way:

Given the following code:
```rust
    let err1 = anyhow::anyhow!("Some error!");
    sentry::integrations::anyhow::capture_anyhow(&err1);

    let err2 = anyhow::anyhow!("Other error!");
    sentry::integrations::anyhow::capture_anyhow(&err2);
```

In Sentry they appear like this:

<img width="1143" alt="Image" src="https://github.com/user-attachments/assets/12e74821-7b03-49ed-b43a-91f5c2b7c529" />

with the following "Event Grouping Information":

<img width="1160" alt="Image" src="https://github.com/user-attachments/assets/57e43abd-73f7-46f0-861f-1cede839d18f" />


So I'm not sure if this is the intended way. Basically they both are simple anyhow errors, but they are different errors generated in different place in the code, so in my opinion I think they should not be grouped.


Anyway, in my case I decided to use a custom `scope.set_fingerprint(...);` which now logs them separately, but I think that should be the default way... so maybe the issue/bug it's just with anyhow errors.

lcian:
[...]

PR #759, opened 2025-03-26
fix(anyhow): attach stacktrace only if anyhow error has backtrace

lcian:
This is similar to https://github.com/getsentry/sentry-rust/pull/755 but for the anyhow integration

Without this change, if we're running without `RUST_BACKTRACE` set, then `backtrace` will have value `"<disabled>"`.
We would try to parse it and get an empty vec as a result.
The consequence is that all anyhows would be grouped in a single issue due to having the same (empty) stack trace.

After the change, there will be no stack trace, so grouping will happen according to the error message.

Side effect: when running with `attach_stacktrace: true`, `sentry-backtrace` will kick in and provide a backtrace (this didn't happen before because we had a stack trace, even though it was an empty one).
It could be argued whether we want this or not.

Closes https://github.com/getsentry/sentry-rust/issues/745

### DIFF
--- a/sentry-anyhow/src/lib.rs
+++ b/sentry-anyhow/src/lib.rs
@@ -72,7 +72,12 @@ pub fn event_from_error(err: &anyhow::Error) -> Event<'static> {
         // exception records are sorted in reverse
         if let Some(exc) = event.exception.iter_mut().last() {
             let backtrace = err.backtrace();
-            exc.stacktrace = sentry_backtrace::parse_stacktrace(&format!("{backtrace:#}"));
+            if matches!(
+                backtrace.status(),
+                std::backtrace::BacktraceStatus::Captured
+            ) {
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
nice! thanks for fixing this and providing a good explanation.

Maybe it might be worth still checking `exc.stacktrace` if its empty, and clearing it in that case?

lcian:
Sorry the stacktrace is not empty, it's going to have a single frame which will contain either `disabled` or `unsupported`.
We wrongly parse that and believe it's a frame.
This is due to the fact that anyhow uses some custom logic to format the backtrace https://github.com/dtolnay/anyhow/blob/master/src/backtrace.rs. Luckily they provide `backtrace.status()` to check it.
In general we already return `None` in `Stacktrace::from_frames_reversed` if the vec of frames is empty, so I don't think we need to check twice.

Issue #702, opened 2024-10-28
Errors are silently discarded if tracing event gets converted to a message


thomaseizinger:
### Environment

- Version 0.34.0

### Steps to Reproduce

1. Create an event-filter that maps WARN logs to Sentry events: https://github.com/firezone/firezone/blob/51d92265f4163967c0045f67495954515da15c62/rust/logging/src/lib.rs#L102-L115
2. Attach errors as `&dyn std::error::Error` in a WARN log as [advised](https://docs.rs/sentry-tracing/latest/sentry_tracing/#tracking-errors) in the documentation: https://github.com/firezone/firezone/blob/51d92265f4163967c0045f67495954515da15c62/rust/gateway/src/eventloop.rs#L80

### Expected Result

To receive a Sentry event that contains the error.

### Actual Result

A Sentry event that only contains the log message but not the error:

![Image](https://github.com/user-attachments/assets/3782ad18-de4f-4ebd-8c9f-68f509e81652)

---

I've been digging through the source code here and the problem is that when a tracing event gets converted to a Sentry event, all fields apart from what is prefixed with `tags.` gets silently dropped: https://github.com/getsentry/sentry-rust/blob/a1481d42ae7a9a1f3f396d6948b0e61354a20ba4/sentry-tracing/src/converters.rs#L212-L230

This is pretty confusing because the documentation clearly states:

> To track [error structs](https://doc.rust-lang.org/nightly/core/error/trait.Error.html), assign a reference to error trait object as field in one of the logging macros. By convention, it is recommended to use the ERROR level and assign it to a field called error, although the integration will also work with all other levels and field names.

What is perhaps missing here is a foot-note that it will work with all levels as long as you configure an event filter that converts the event to a sentry exception! However, instead of fixing the documentation here, I think the proper fix is to actually make this work the way the documentation states and attach the exception to the message, even if the log level is not ERROR.

thomaseizinger:
> However, instead of fixing the documentation here, I think the proper fix is to actually make this work the way the documentation states and attach the exception to the message, even if the log level is not ERROR.

I tried to fix this but realised that this essentially just makes "messages" and "exceptions" the same?

thomaseizinger:
> > However, instead of fixing the documentation here, I think the proper fix is to actually make this work the way the documentation states and attach the exception to the message, even if the log level is not ERROR.
> 
> I tried to fix this but realised that this essentially just makes "messages" and "exceptions" the same?

[...]

PR #768, opened 2025-04-10
refactor(tracing): remove `EventFilter::exception` and always attach exception

lcian:
Before this change, when capturing something with e.g. `tracing::warn!(error = std_dyn_err(&e), "Something went wrong");` the following would happen based on the `EventFilter`:
- `EventFilter::Exception` -> `exception_from_event` (includes `error` in the `exception` field of the `Event`)
- `EventFilter::Event` -> `event_from_event` (ignores `error`)
- `EventFilter::Breadcrumb` -> `breadcrumb_from_event` (ignores `error`)
  So, both `Event` and `Breadcrumb` would counterintuitively ignore the field.

With this change:
- we remove `EventFilter::Exception` and `exception_from_event` in favor of `EventFilter::Event` and `event_from_event`, which will behave as the old `exception_from_event` (attaching `error`, if any)
- we attach the `error` field as structured data in breacrumbs. It's attached as a list of errors from the outermost to the innermost, mimicking the way it would be reported in a Sentry issue.

Closes [https://github.com/getsentry/sentry-rust/issues/702](https://github.com/getsentry/sentry-rust/issues/702)


Swatinem (approved):
just a heads up that this will be a breaking change.

PR #774, opened 2025-04-22
feat: reduce stack-size of threads

thomaseizinger:
Certain environments like iOS are very resource-constrained. A network extension on iOS is allowed to at most consume 50MB of RAM. By default, a thread in Rust allocates 2MB of RAM for the stack of the new thread. This can be tweaked if one knows that the particular thread doesn't need as much RAM.

### DIFF
--- a/sentry-core/src/session.rs
+++ b/sentry-core/src/session.rs
@@ -213,6 +213,7 @@ mod session_impl {
             let worker_shutdown = shutdown.clone();
             let worker = std::thread::Builder::new()
                 .name("sentry-session-flusher".into())
+                .stack_size(100 * 1024)
                 .spawn(move || {
                     let (lock, cvar) = worker_shutdown.as_ref();
                     let mut shutdown = lock.lock().unwrap();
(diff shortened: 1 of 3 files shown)


thomaseizinger:
I am opening this as a starting point for a discussion. The implementation is purposely very minimal for now. I am happy to extend with comments, consts etc.

I wanted to discuss first if this is something that you'd be willing to merge.

thomaseizinger:
I checked all of these threads on a high-level and none of them looked like they would consume a lot of stack-size. So I set it to 100KB for all of them which should be plenty for what we are doing in these threads.

Swatinem (commented):
I think this is reasonable. As you commented, running a full tokio runtime for usage with `reqwest` is still quite heavy on stack usage though.

thomaseizinger:
> I think this is reasonable. As you commented, running a full tokio runtime for usage with `reqwest` is still quite heavy on stack usage though.

500KB seems to allow CI to pass. I can experiment a bit more to see if we can get it any lower although it is already a great improvement :)

I am not too familiar with the test setup in this repo. Are they exhaustive enough that we hit all scenarios of stack usage and can therefore trust that green CI means we have enough stack for all situations?

Swatinem:
TBH, I wouldn’t put too much trust in them. In particular, I’m not even sure whether we are even testing the http interface at all, or just relying on a mock transport for tests mostly.

PR #776, opened 2025-04-24
Replace `once_cell` with `std::sync::LazyLock`

FalkWoldmann:


Swatinem (approved):
lgtm, but we should check whether it aligns with our advertised MSRV.

otherwise, as the `Cargo.lock` shows, we still depend on it transitively, and probably will for a while, as these changes are really slow to propagate through the ecosystem.

FalkWoldmann:
As far as I can tell, the README mentions an MSRV of 1.81, and since LazyLock stabilized in 1.80, it seems like it should be okay —does that sound right? :)

Swatinem:
Yes. Our CI was also green, so lets go :-)

PR #777, opened 2025-04-25
Update GH issue templates for Linear compatibility

stephanie-anderson:


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
```
Swatinem, line 1: I would personally prefer to still have blank issues, but okay :-)

Swatinem (approved):

PR #778, opened 2025-04-29
chore: update issue templates with blank issue and Discord link

lcian:


Swatinem (approved):

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,9 @@
 blank_issues_enabled: false
```
Swatinem, line 1: why not just turn this to `true`? because you would still want to auto-assign the `labels`?

lcian: This way we can auto-assign the labels and also provide the "Reserved for maintainers" description, both of which I don't think are possible with `black_issues_enabled: true`.

PR #787, opened 2025-05-07
fix: use `release-health` flag in `sentry-actix` and remove it from subcrates where unneeded

lcian:
Removes the `release-health` feature from `sentry-tower` and `sentry-tracing`, where it was used unnecessarily.

It's still a feature flag, enabled by default, in `sentry-actix`, as it's used there to create a session per request.
This was the default behavior previously, with no way of opting out, so it makes sense to keep it enabled by default there.
https://github.com/getsentry/sentry-rust/blob/3cc461a279d5a6ac45ea73268344d01318d8719e/sentry-actix/Cargo.toml#L16
I've fixed it to actually be used in the code because compilation would fail otherwise due to the `hub.start_session();` call.

There is no change to the `sentry` crate, where this flag is part of the default features as intended: https://github.com/getsentry/sentry-rust/blob/3cc461a279d5a6ac45ea73268344d01318d8719e/sentry/Cargo.toml#L30

Note that `sentry` has no way to bring in `sentry-actix` via feature flags.
Is there a reason for that?
I'll create a separate PR to enable it, similarly to other integrations.

### DIFF
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -45,8 +45,6 @@
 //! The actix middleware will automatically start a new session for each request
 //! when `auto_session_tracking` is enabled and the client is configured to
 //! use `SessionMode::Request`.
-//! The `release-health` feature flag also needs to be enabled on either
-//! `sentry` or `sentry-actix` for this to work.
 //!
 //! ```
 //! let _sentry = sentry::init(sentry::ClientOptions {
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -294,6 +294,7 @@ where
             .map(|client| client.options().max_request_body_size)
             .unwrap_or(MaxRequestBodySize::None);
         if track_sessions {
+            #[cfg(feature = "release-health")]
             hub.start_session();
         }
         let with_pii = client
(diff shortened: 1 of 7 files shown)


sentry-actix/src/lib.rs
```diff
@@ -294,6 +296,7 @@ where
             .map(|client| client.options().max_request_body_size)
             .unwrap_or(MaxRequestBodySize::None);
         if track_sessions {
```
Swatinem, line 298: you should probably move the `#[cfg]` up to where you define this field. I believe the client options `auto_session_tracking` is also not available with this feature, right?

lcian: At the moment the option is available even without the flag, I've just added a doc line on it to make it clear that the feature flag needs to be enabled for that option to have any effect.

But, I think both `auto_session_tracking` and `session_mode` in `ClientOptions` should instead not be available if the feature is not enabled.
This way, if someone uses those options but their deps don't bring in `sentry-core/release-health`, it will fail to compile. Better than no-opping silently.

I will update the PR.

sentry-core/src/session.rs
```diff
@@ -432,6 +432,7 @@ mod session_impl {
         }
 
         #[test]
+        #[cfg(feature = "release-health")]
```
Swatinem, line 435: I think we are still running this test in CI, are we?

Swatinem (approved):

sentry-core/src/session.rs
lcian: Actually the whole `session.rs` is wrapped in:
```
#[cfg(feature = "release-health")]
mod session_impl {
```
So effectively these extra `cfg` macros inside it are useless. Removed them.

Swatinem (approved):

PR #790, opened 2025-05-07
docs: update docs including OTEL and other integrations

lcian:
Adds docs for OTEL, basically what's already in `sentry-opentelemetry/src/lib.rs`.

Also mentions the `actix` and `tracing` integrations in the main README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,9 +60,9 @@ This workspace contains various crates that provide support for logging events a
 
   An integration for the `log` and `env_logger` crate.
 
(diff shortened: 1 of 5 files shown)


lcian:
> you do not strictly need to update these manually. the release workflow is using `cargo-readme` to update it based on the toplevel crate docs.

Ah that's great, I didn't know that :). 
I will delete the README.md from the crate docs then so that it's automatically populated.

PR #792, opened 2025-05-07
fix(otel): set `span.op` to `default` by default

lcian:
According to https://develop.sentry.dev/sdk/telemetry/traces/span-operations/ 
`span.op` is supposed to be high quality and low cardinality, also because it's queryable by EAP.
It makes no sense to set it equal to the span name.
Let's just set it to `default` for now, this way it's going to be hidden in the UI and only the span name will be shown.
In the future we will try to infer a more meaningful op from span attributes.

### DIFF
--- a/sentry-opentelemetry/src/processor.rs
+++ b/sentry-opentelemetry/src/processor.rs
@@ -92,19 +92,17 @@ impl SpanProcessor for SentrySpanProcessor {
         let mut span_map = self.span_map.lock().unwrap();
 
         let mut span_description = String::new();
-        let mut span_op = String::new();
         let mut span_start_timestamp = SystemTime::now();
         let mut parent_sentry_span = None;
         if let Some(data) = span.exported_data() {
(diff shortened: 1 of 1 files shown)


adinauer (approved):

Swatinem (approved):
can you also do the same for the `tracing` integration? I believe we do the same thing there.

lcian:
I'm thinking back about this, it should not be merged, it's actually fine to use `span.name`.
Unless the span name is generated dynamically, it's going to be low cardinality.
And it's the only way to distinguish different spans when using the Trace Explorer. 
If we set everything to default, then we won't be able to group/filter by span op at all.
For example, in the tracing integration this works quite well because when using the `instrument` macro on a function, the `span.op` will be the function name.
This is very useful because that way you can group or filter by a specific function.

lcian:
Actually it's possible to query on the span name as well (`span.description`), so this might make sense, will have to check what the consensus is on this one.

PR #794, opened 2025-05-08
fix(otel): fix doctests

lcian:
The doctests show usage from the point of view of the `sentry` crate, as we are importing the subcrate as `use sentry::integrations::opentelemetry as sentry_opentelemetry;`.
Therefore the first doctest will not compile.
The second will also not compile because we have the imports in the first.
Let's just use `ignore` on both of them.

### DIFF
--- a/sentry-opentelemetry/Cargo.toml
+++ b/sentry-opentelemetry/Cargo.toml
@@ -27,7 +27,7 @@ opentelemetry_sdk = { version = "0.29.0", default-features = false, features = [
 opentelemetry-semantic-conventions = "0.29.0"
 
 [dev-dependencies]
-sentry = { version = "0.37.0", path = "../sentry", features = ["test"] }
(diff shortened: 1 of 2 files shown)


Swatinem (commented):
you can still run these doctests by pulling in the main `sentry` crate as a dev dependency.
we use this pattern in the doctests for `sentry-core` itself.

We are abusing dependency cycles in a very creative way here. Not really a fan of it tbh, but it kinda works.

Swatinem (approved):

lcian:
> you can still run these doctests by pulling in the main `sentry` crate as a dev dependency. we use this pattern in the doctests for `sentry-core` itself.
> 
> We are abusing dependency cycles in a very creative way here. Not really a fan of it tbh, but it kinda works.

Ah cool, we were actually already pulling `sentry` as a dev dependency for the `test` feature, didn't think that pulling `opentelemetry` itself would work :)

PR #811, opened 2025-05-19
feat(core): implement Tracing without Performance

### LINKED ISSUE #796: Tracing without Performance
### Problem Statement

The SDK doesn't implement Tracing without Performance (see https://github.com/getsentry/team-sdks/issues/5), which is a prerequisite for e.g. a proper implementation of https://github.com/getsentry/sentry-rust/issues/785

### Solution Brainstorm

Implement it.

lcian:
Implements Tracing without Performance ([spec](https://develop.sentry.dev/sdk/telemetry/traces/tracing-without-performance/)).

This is done by having an additional `PropagationContext` attribute on the scope that holds a "fake" `trace_id` and `span_id`.
When applying the scope to an event, if there is no ongoing span set on the scope, we create the `trace` context from the `PropagationContext`.
Additionally, adds an API `scope.iter_trace_propagation_headers` that should be preferred over something like:
```rust
sentry::configure_scope(|scope| {
	if let Some(span) = scope.get_span() {
		headers = Some(span.iter_headers());
	} 
});
```
because it will use the `PropagationContext` as a fallback.

Closes #796

Also makes a minor adjustment to trace and span ids where we now use the `Display` impl for `Debug`, otherwise we were getting u8 arrays when using e.g. `dbg!` with the derived one which is not very useful.

### DIFF
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -778,11 +778,11 @@ impl Transaction {
     /// span's/transaction's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         let inner = self.inner.lock().unwrap();
-        let trace = SentryTrace {
-            trace_id: inner.context.trace_id,
-            span_id: inner.context.span_id,
-            sampled: Some(inner.sampled),
-        };
+        let trace = SentryTrace::new(
+            inner.context.trace_id,
+            inner.context.span_id,
+            Some(inner.sampled),
+        );
         TraceHeadersIter {
             sentry_trace: Some(trace.to_string()),
         }
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -476,8 +476,8 @@ impl TransactionOrSpan {
     }
 
     /// Returns the headers needed for distributed tracing.
-    /// [`crate::Scope::iter_trace_propagation_headers`] is preferred if the intention is to get
-    /// the distributed tracing headers for the currently active span/transaction.
+    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
+    /// span's/transaction's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         match self {
             TransactionOrSpan::Transaction(transaction) => transaction.iter_headers(),
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -776,8 +776,8 @@ impl Transaction {
     }
 
     /// Returns the headers needed for distributed tracing.
-    /// [`crate::Scope::iter_trace_propagation_headers`] is preferred if the intention is to get
-    /// the distributed tracing headers for the currently active transaction.
+    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
+    /// span's/transaction's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         let inner = self.inner.lock().unwrap();
         let trace = SentryTrace(
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -1030,8 +1030,8 @@ impl Span {
     }
 
     /// Returns the headers needed for distributed tracing.
-    /// [`crate::Scope::iter_trace_propagation_headers`] is preferred if the intention is to get
-    /// the distributed tracing headers for the currently active span.
+    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
+    /// span's/transaction's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         let span = self.span.lock().unwrap();
         let trace = SentryTrace(span.trace_id, span.span_id, Some(self.sampled));
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -476,7 +476,7 @@ impl TransactionOrSpan {
     }
 
     /// Returns the headers needed for distributed tracing.
-    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
+    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active
     /// span's/transaction's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         match self {
@@ -776,7 +776,7 @@ impl Transaction {
     }
 
     /// Returns the headers needed for distributed tracing.
-    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
+    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active
     /// span's/transaction's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         let inner = self.inner.lock().unwrap();
@@ -1030,7 +1030,7 @@ impl Span {
     }
 
     /// Returns the headers needed for distributed tracing.
-    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
+    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active
     /// span's/transaction's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         let span = self.span.lock().unwrap();
--- a/sentry-core/src/performance.rs
+++ b/sentry-core/src/performance.rs
@@ -475,7 +475,7 @@ impl TransactionOrSpan {
 
     /// Returns the headers needed for distributed tracing.
     /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active
-    /// span's/transaction's distributed tracing headers.
+    /// trace's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         match self {
             TransactionOrSpan::Transaction(transaction) => transaction.iter_headers(),
@@ -775,7 +775,7 @@ impl Transaction {
 
     /// Returns the headers needed for distributed tracing.
     /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active
-    /// span's/transaction's distributed tracing headers.
+    /// trace's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         let inner = self.inner.lock().unwrap();
         let trace = SentryTrace::new(
@@ -1029,7 +1029,7 @@ impl Span {
 
     /// Returns the headers needed for distributed tracing.
     /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active
-    /// span's/transaction's distributed tracing headers.
+    /// trace's distributed tracing headers.
     pub fn iter_headers(&self) -> TraceHeadersIter {
         let span = self.span.lock().unwrap();
         let trace = SentryTrace::new(span.trace_id, span.span_id, Some(self.sampled));
@@ -1141,7 +1141,7 @@ pub struct TraceHeadersIter {
 }
 
 impl TraceHeadersIter {
-    #[cfg_attr(not(feature = "client"), allow(dead_code))]
+    #[cfg(feature = "client")]
     pub(crate) fn new(sentry_trace: String) -> Self {
         Self {
             sentry_trace: Some(sentry_trace),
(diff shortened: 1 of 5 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -19,6 +21,12 @@ pub struct Stack {
 
 pub type EventProcessor = Arc<dyn Fn(Event<'static>) -> Option<Event<'static>> + Send + Sync>;
 
+#[derive(Debug, Default, Clone, Copy)]
+pub(crate) struct PropagationContext {
+    trace_id: TraceId,
+    span_id: SpanId,
+}
```
lcian, line 28: Note that this new type is very similar to the already existing `SentryTrace` type, but in theory `PropagationContext` could also contain data related to the DSC (baggage) which we don't support yet, so we make it a separate type.

sentry-core/src/performance.rs
```diff
@@ -476,6 +476,8 @@ impl TransactionOrSpan {
     }
 
     /// Returns the headers needed for distributed tracing.
+    /// [`crate::Scope::iter_trace_propagation_headers`] is preferred if the intention is to get
+    /// the distributed tracing headers for the currently active span/transaction.
```
szokeasaurusrex, line 480: [nit] try to make this comment a bit more concise:

```suggestion
    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
    /// span's/transaction's distributed tracing headers.
```

sentry-core/src/performance.rs
```diff
@@ -774,6 +776,8 @@ impl Transaction {
     }
 
     /// Returns the headers needed for distributed tracing.
+    /// [`crate::Scope::iter_trace_propagation_headers`] is preferred if the intention is to get
+    /// the distributed tracing headers for the currently active transaction.
```
szokeasaurusrex, line 780: [nit] here as well

```suggestion
    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
    /// span's/transaction's distributed tracing headers.
```

sentry-core/src/performance.rs
```diff
@@ -1026,6 +1030,8 @@ impl Span {
     }
 
     /// Returns the headers needed for distributed tracing.
+    /// [`crate::Scope::iter_trace_propagation_headers`] is preferred if the intention is to get
+    /// the distributed tracing headers for the currently active span.
```
szokeasaurusrex, line 1034: [nit] also here

```suggestion
    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active 
    /// span's/transaction's distributed tracing headers.
```

sentry-core/src/performance.rs
```diff
@@ -1143,8 +1149,12 @@ impl Iterator for TraceHeadersIter {
 
 /// A container for distributed tracing metadata that can be extracted from e.g. the `sentry-trace`
 /// HTTP header.
-#[derive(Debug, PartialEq)]
-pub struct SentryTrace(protocol::TraceId, protocol::SpanId, Option<bool>);
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct SentryTrace(
+    pub(crate) protocol::TraceId,
+    pub(crate) protocol::SpanId,
+    pub(crate) Option<bool>, // sampled
```
szokeasaurusrex, line 1156: If we are making this struct's items crate-public, we might want to convert it into a normal struct with named fields. Or, add a constructor rather than making the fields accessible (the constructor probably would only need to take the trace ID and span ID, and it would be default set the sampled flag to `None`).

sentry-core/src/scope/real.rs
```diff
@@ -357,4 +369,33 @@ impl Scope {
...
+            trace_id: self.propagation_context.trace_id,
+            span_id: self.propagation_context.span_id,
+            ..Default::default()
+        };
+        event.contexts.insert("trace".into(), context.into());
+    }
+
+    /// Returns the headers needed for distributed tracing.
+    pub fn iter_trace_propagation_headers(&self) -> TraceHeadersIter {
+        if let Some(span) = self.get_span() {
```
szokeasaurusrex, line 388: I'd recommend returning an opaque type here; that way, you can change the underlying `TraceHeadersIter` struct to a different type at any time without it being a breaking API change.

```suggestion
    pub fn iter_trace_propagation_headers(&self) -> impl Iterator<Item = (&'static str, String)> {
```

sentry-core/src/scope/real.rs
szokeasaurusrex: I'd honestly suggest using the existing `SentryTrace` type until whenever we need to add the DSC and make it a separate type.

Or, is there another reason why we should have this struct now?

sentry-core/src/performance.rs
```diff
@@ -1130,7 +1136,7 @@ impl Span {
 /// This currently only yields the `sentry-trace` header, but other headers
 /// may be added in the future.
 pub struct TraceHeadersIter {
-    sentry_trace: Option<String>,
+    pub(crate) sentry_trace: Option<String>,
```
szokeasaurusrex, line 1139: Since it looks like you are only using this field to create a new `TraceHeadersIter`, I'd suggest keeping the field private and adding a `pub(crate)` constructor (`pub(super)` might also be sufficient).

```suggestion
    sentry_trace: Option<String>,
```

The constructor could look like:

```rust
pub(super) fn new(sentry_trace: String) {
    Self { sentry_trace: Some(sentry_trace) }
}
```

sentry-types/src/protocol/v7.rs
```diff
@@ -1368,6 +1368,12 @@ impl fmt::Display for SpanId {
     }
 }
 
+impl fmt::Debug for SpanId {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "{}", self)
+    }
+}
```
szokeasaurusrex, line 1375: The Debug should probably indicate the type of this value. For example:


```suggestion
impl fmt::Debug for SpanId {
    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
        write!(fmt, "SpanId({})", self)
    }
}
```

Also, minor nit: as far as I can tell, this change appears to be unrelated to the rest of the changes in this PR, so likely it would have been better to split it off as a separate PR. I think it is okay to keep it here this time, but for the future, it is best to try to keep PRs limited in scope to only one change.

sentry-types/src/protocol/v7.rs
```diff
@@ -1415,6 +1421,12 @@ impl fmt::Display for TraceId {
     }
 }
 
+impl fmt::Debug for TraceId {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(fmt, "{}", self)
```
szokeasaurusrex, line 1426: Likewise here, add the type to the debug output:


```suggestion
        write!(fmt, "TraceId({})", self)
```

This likely also belongs in a separate PR with the SpanId change

szokeasaurusrex (commented):
Looks good overall, I have some (mostly minor) suggestions though

sentry-core/src/scope/real.rs
```diff
@@ -357,4 +369,33 @@ impl Scope {
...
+        let context = TraceContext {
+            trace_id: self.propagation_context.trace_id,
+            span_id: self.propagation_context.span_id,
+            ..Default::default()
+        };
+        event.contexts.insert("trace".into(), context.into());
+    }
+
+    /// Returns the headers needed for distributed tracing.
+    pub fn iter_trace_propagation_headers(&self) -> TraceHeadersIter {
```
szokeasaurusrex, line 387: I'd recommend returning an opaque type here; that way, you can change the underlying `TraceHeadersIter` struct to a different type at any time without it being a breaking API change.

```suggestion
    pub fn iter_trace_propagation_headers(&self) -> impl Iterator<Item = (&'static str, String)> {
```

sentry-core/src/scope/real.rs
lcian: The only justification to have this separate struct would be that this one is `pub(crate)` and the other one is `pub` (because it needs to be used in `sentry-opentelemetry`).
Our internal notion doc for TwP suggests to make the `propagation_context` field private.
However reusing `SentryTrace` until we have DSC, as you suggest, sounds like a better option. Its fields are `pub(crate)` so even if we change them it will not be a breaking change.

sentry-core/src/performance.rs
lcian: Yeah I agree, I also thought it was kinda ugly to have it as a tuple

sentry-core/src/scope/real.rs
lcian: Nice, I've used `TraceHeadersIter` here just because that's what the other functions on `Scope`/`Transaction` use but `impl Iterator<Item = TraceHeader>` sounds better.

szokeasaurusrex: Yeah, I mean long term (as in, before a 1.x release) I'd recommend updating those as well. For now, you can keep them as is to avoid a breaking change

sentry-core/src/performance.rs
```diff
@@ -774,13 +774,15 @@ impl Transaction {
     }
 
     /// Returns the headers needed for distributed tracing.
+    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active
+    /// span's/transaction's distributed tracing headers.
```
lcian, line 778: The idea is that this could also return the ids from PropagationContext if there is no ongoing span, not sure if this comment correctly reflects that. 
But on the other hand we should also hide PropagationContext as that's an "implementation detail".

szokeasaurusrex: In that case, I would suggest rephrasing this comment as follows:

```suggestion
    /// Use [`crate::Scope::iter_trace_propagation_headers`] to obtain the active
    /// trace's distributed tracing headers.
```

This is technically correct, and you avoid going into detail about span/transaction/propagation context/etc.

You may want to update the other comments as well

sentry-core/src/performance.rs
```diff
@@ -1133,6 +1140,15 @@ pub struct TraceHeadersIter {
     sentry_trace: Option<String>,
 }
 
+impl TraceHeadersIter {
+    #[cfg_attr(not(feature = "client"), allow(dead_code))]
```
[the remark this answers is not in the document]
szokeasaurusrex: Instead of allowing dead code when the `client` feature is not set, it would likely be better to only compile this method when the `client` feature is enabled.

```suggestion
    #[cfg(feature = "client")]
```

This is more readable, and has the added benefit of making the compiled library code slightly smaller when not using the `client` feature (since this code does not need to be compiled at all).

sentry-core/src/performance.rs
```diff
@@ -1143,8 +1159,12 @@ impl Iterator for TraceHeadersIter {
 
 /// A container for distributed tracing metadata that can be extracted from e.g. the `sentry-trace`
 /// HTTP header.
-#[derive(Debug, PartialEq)]
-pub struct SentryTrace(protocol::TraceId, protocol::SpanId, Option<bool>);
+#[derive(Debug, PartialEq, Clone, Copy, Default)]
+pub struct SentryTrace {
+    pub(crate) trace_id: protocol::TraceId,
+    pub(crate) span_id: protocol::SpanId,
+    pub(crate) sampled: Option<bool>,
```
szokeasaurusrex, line 1166: Looks like `sampled` is not being used outside this module, so you should be able to make it private I think.

```suggestion
    pub(crate) trace_id: protocol::TraceId,
    pub(crate) span_id: protocol::SpanId,
    sampled: Option<bool>,
```

Also [nit] I would consider just adding accessors for the `trace_id` and `span_id` since it doesn't look like you are mutating them from outside (but I suppose that is just my personal preference that I like to keep struct fields private and access through methods; since they are already only `pub(crate)` it doesn't matter that much so feel free to ignore this suggestion)

szokeasaurusrex: Also I think this change to named fields makes the code way more readable 🎉

sentry-core/src/scope/real.rs
```diff
@@ -357,4 +363,31 @@ impl Scope {
...
+    pub fn iter_trace_propagation_headers(&self) -> impl Iterator<Item = TraceHeader> {
+        if let Some(span) = self.get_span() {
+            span.iter_headers()
+        } else {
+            let data = SentryTrace::new(
+                self.propagation_context.trace_id,
+                self.propagation_context.span_id,
+                None,
+            );
+            TraceHeadersIter::new(data.to_string())
```
szokeasaurusrex, line 390: [optional suggestion] you could consider adding this as an `iter_headers` method on the `SentryTrace`

szokeasaurusrex (approved):

szokeasaurusrex (commented):
Looks good, please address comments before merging (I marked some as nit or optional, you can ignore those if you want)

PR #813, opened 2025-05-23
ci: adopt Danger with Sentry's shared config

lcian:
Enables Danger with our shared config.
I prefer to be forced to write the changelog directly with each PR.

#skip-changelog


lcian:
Closing and reopening this because either something is wrong or Danger detected "#skip-changelog" in my original description and it's not running, not sure.

PR #815, opened 2025-05-23
ci: adopt Danger with Sentry's shared config

lcian:
Enables Danger with our shared config.
I prefer to be forced to write the changelog directly with each PR.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,0 +1,9 @@
+name: Danger
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited, ready_for_review]
+
+jobs:
+  danger:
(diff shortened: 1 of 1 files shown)


lcian:
Okay it works, it looked like it wasn't working because a commit with `ci` scope doesn't require a changelog entry

Swatinem (approved):
well, if you must, why not :-D

I personally hate this, but I think it makes sense to actually enforce for a SDK repo (as opposed to symbolic/ator which had this set up, just to be a huge pain in the rear.

lcian:
Yeah, no one is forcing me to enable it but personally I find it easier to do the changelog immediately rather than having to describe what I did later and requiring an additional PR.

Issue #797, opened 2025-05-08
[Structured logging] Types


lcian:
### Description

Implement types required for structured logging.

AbhiPrasad:
I pushed up a branch of for this in `abhi-logs-protocol`: https://github.com/getsentry/sentry-rust/tree/abhi-logs-protocol

Was hoping to open a PR, but couldn't get the time to finish it up. Hopefully this helps as a starting point when looking at this!

lcian:
Thank you @AbhiPrasad !

PR #821, opened 2025-05-26
feat(logs): add log protocol types

lcian:
Closes https://github.com/getsentry/sentry-rust/issues/797

To do: full envelope de/serialization tests
Also need to confirm if it's a good idea to have this `ItemsContainer`

### DIFF
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -74,15 +74,10 @@ enum EnvelopeItemType {
 struct EnvelopeItemHeader {
     r#type: EnvelopeItemType,
     length: Option<usize>,
-
-    content_type: Option<String>, // Applies (only) to both Attachment and ItemsContainer Item type
-
     // Fields below apply only to Attachment Item type
     filename: Option<String>,
     attachment_type: Option<AttachmentType>,
-
-    // Field below applies only to ItemsContainer Item type
-    item_count: Option<usize>,
+    content_type: Option<String>,
 }
 
 /// An Envelope Item.
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -4,7 +4,7 @@ use serde::Deserialize;
 use thiserror::Error;
 use uuid::Uuid;
 
-use super::v7 as protocol;
+use super::v7::{self as protocol, ItemsContainer};
 
 use protocol::{
     Attachment, AttachmentType, Event, MonitorCheckIn, SessionAggregates, SessionUpdate,
@@ -61,9 +61,12 @@ enum EnvelopeItemType {
     /// An Attachment Item type.
     #[serde(rename = "attachment")]
     Attachment,
-    /// A Monitor Check In Item Type
+    /// A Monitor Check In Item Type.
     #[serde(rename = "check_in")]
     MonitorCheckIn,
+    /// A Log Items Type.
+    #[serde(rename = "log")]
+    Log,
 }
 
 /// An Envelope Item Header.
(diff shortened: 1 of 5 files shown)


sentry-types/src/protocol/envelope.rs
```diff
@@ -61,9 +61,12 @@ enum EnvelopeItemType {
     /// An Attachment Item type.
     #[serde(rename = "attachment")]
     Attachment,
-    /// A Monitor Check In Item Type
+    /// A Monitor Check In Item Type.
     #[serde(rename = "check_in")]
     MonitorCheckIn,
+    /// A Log Items Type.
+    #[serde(rename = "log")]
+    Log,
```
lcian, line 69: Here we break the "abstraction" of the `ItemsContainer`, however this enum is not public so it should be fine.

lcian:
This needs to be redone because `#[serde(rename = "items")]` is completely wrong.

sentry-types/src/protocol/v7.rs
```diff
@@ -2118,3 +2118,213 @@ impl fmt::Display for Transaction<'_> {
...
+
+/// An attribute that can be attached to a log.
+#[derive(Clone, Debug, PartialEq)]
+pub struct LogAttribute(pub Value);
+
+impl From<Value> for LogAttribute {
+    fn from(value: Value) -> Self {
+        Self(value)
+    }
+}
```
lcian, line 2196: It would be nice to inherit all the `From<T> for Value` impls that `Value` has and obtain `From<T> for LogAttribute`.
Otherwise right now it's always necessary to construct a `Value` and then convert it to a `LogAttribute`.
Is this somehow possible or would I need to copy-paste all of those from `Value`?

sentry-types/src/protocol/envelope.rs
```diff
@@ -112,12 +116,63 @@ pub enum EnvelopeItem {
...
     Raw,
     // TODO:
     // etc…
 }
 
+/// A container for a list of multiple items.
+/// It's considered a single envelope item, with its `type` corresponding to the contained items'
+/// `type`.
+#[derive(Clone, Debug, PartialEq)]
+pub enum ItemContainer {
```
Swatinem, line 132: marking this as `#[non_exhaustive]` means we can add more variants in the future without breaking semver

sentry-types/src/protocol/v7.rs
```diff
@@ -2118,3 +2118,213 @@ impl fmt::Display for Transaction<'_> {
...
+    pub timestamp: SystemTime,
+    /// The severity number of the log (required).
+    pub severity_number: LogSeverityNumber,
+    /// Additional arbitrary attributes attached to the log.
+    #[serde(default, skip_serializing_if = "Map::is_empty")]
+    pub attributes: Map<String, LogAttribute>,
+}
+
+/// Indicates the severity of a log, according to the
+/// OpenTelemetry [`SeverityText`](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitytext) spec.
```
Swatinem, line 2143: its a bit weird that you have both a text and a number representation of this. why are both required? is there a canonical mapping from one to the other?

sentry-types/src/protocol/v7.rs
Swatinem: would `impl<T> From<T> for LogAttribute where T: Into<Value>` or `where Value: From<T>` work?

Swatinem (approved):

sentry-types/src/protocol/v7.rs
lcian: Ah right, `severity_number` is optional in our spec and should be inferred from `severity_text` if missing.
So I'm thinking I'll make the `severity_number` an `Option<SeverityNumber>`.
And then we can do the inference with a custom serialization function.

lcian: Actually the inference is done server side so we don't need the custom serialization.

sentry-types/src/protocol/v7.rs
lcian: Ah yeah this works, nice!!

Swatinem (approved):

Issue #798, opened 2025-05-08
[Structured logging] Ability to send logs


lcian:
### Description

Implement ability to send logs on the Client.

PR #823, opened 2025-05-28
feat(logs): add ability to send logs and basic API

lcian:
Adds the ability to capture `Log`s and a basic API.
Link to the [spec](https://develop.sentry.dev/sdk/telemetry/logs/#public-api)

#### Sending logs
- The features are gated behind a "logs" flag. It could be a bit confusing because we have a "log" flag that pulls in "sentry-log" already, but at least we removed "debug-logs".
- We need it to be a feature flag for the same reason as `release-health`: we will use a thread to batch and flush the logs, which would not be compatible with e.g. WASM. So, we need to make it optional.
- TODO: disable the flag by default on `sentry` and mark it as `experimental`/`unstable` until we are good with the API.
- According to the spec, an additional init option `enable_logs` was added to enable capturing them.

#### API
- In a followup PR I plan to create macros equivalent to those of `tracing` that will construct the `Log` and capture it. 
- Usage would look like: `sentry::logger::warn!("warning", attribute = 42);`
- `sentry::capture_log` and `Hub::capture_log` will be removed in that PR in favor of the macros.

Part of https://github.com/getsentry/sentry-rust/issues/798
Comes after https://github.com/getsentry/sentry-rust/pull/821

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -5,9 +5,11 @@ use std::panic::RefUnwindSafe;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
-use rand::random;
+#[cfg(feature = "logs")]
+use crate::protocol::EnvelopeItem;
 #[cfg(feature = "release-health")]
-use sentry_types::protocol::v7::SessionUpdate;
+use crate::protocol::SessionUpdate;
+use rand::random;
 use sentry_types::random_uuid;
 
 use crate::constants::SDK_INFO;
@@ -18,6 +20,8 @@ use crate::types::{Dsn, Uuid};
 #[cfg(feature = "release-health")]
 use crate::SessionMode;
 use crate::{ClientOptions, Envelope, Hub, Integration, Scope, Transport};
+#[cfg(feature = "logs")]
+use sentry_types::protocol::v7::{Log, LogAttribute};
 
 impl<T: Into<ClientOptions>> From<T> for Client {
     fn from(o: T) -> Client {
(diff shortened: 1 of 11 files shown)


sentry-core/src/client.rs
```diff
@@ -363,6 +367,87 @@ impl Client {
...
+        }
+
+        if !log.attributes.contains_key("sentry.sdk.version") {
+            log.attributes.insert(
+                "sentry.sdk.version".to_owned(),
+                LogAttribute(self.sdk_info.version.to_owned().into()),
+            );
+        }
+
+        // TODO: set OS (and Rust?) context
```
lcian, line 440: We cannot do this at the moment because `sentry-contexts` adds those through an event processor, which logs don't go through.

sentry-types/src/protocol/v7.rs
```diff
@@ -2127,17 +2127,38 @@ pub struct Log {
...
-    pub trace_id: TraceId,
+    pub trace_id: Option<TraceId>,
     /// The timestamp of the log (required).
-    #[serde(with = "ts_seconds_float")]
-    pub timestamp: SystemTime,
+    #[serde(with = "ts_rfc3339_opt")]
+    pub timestamp: Option<SystemTime>,
     /// The severity number of the log (required).
-    pub severity_number: LogSeverityNumber,
+    pub severity_number: Option<LogSeverityNumber>,
```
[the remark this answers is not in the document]
lcian: The constructor too.

sentry-types/src/protocol/v7.rs
```diff
@@ -2127,7 +2127,8 @@ pub struct Log {
     /// The log body/message (required).
     pub body: String,
     /// The ID of the Trace in which this log happened (required).
-    pub trace_id: TraceId,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub trace_id: Option<TraceId>,
```
lcian, line 2131: Making this optional because conceptually the `TraceId` should be filled when applying scope to the log, which is done after the top-level API is called.

sentry-core/src/client.rs
```diff
@@ -363,6 +367,87 @@ impl Client {
...
+        scope.apply_to_log(&mut log, self.options.send_default_pii);
+
+        self.set_log_default_attributes(&mut log);
+
+        if let Some(ref func) = self.options.before_send_log {
+            if let Some(res) = func(log) {
+                log = res
+            } else {
+                return None;
+            }
```
Swatinem:
```suggestion
            log = func(log)?;
```

sentry-core/src/client.rs
```diff
@@ -363,6 +367,87 @@ impl Client {
...
+        }
+
+        Some(log)
+    }
+
+    #[cfg(feature = "UNSTABLE_logs")]
+    fn set_log_default_attributes(&self, log: &mut Log) {
```
Swatinem, line 403: So this is what all other SDKs are doing? IMO it might be a bit wasteful to copy the same fields into the log event for every single log event.

In the likely case that users are not logging any attributes, this will allocate a hashmap, and clone at least `4 * 2` Strings into it. That is quite some overhead.

sentry-core/src/scope/real.rs
```diff
@@ -346,6 +348,57 @@ impl Scope {
...
+                log.attributes.insert(
+                    "parent_span_id".to_owned(),
+                    LogAttribute(span_id.to_string().into()),
+                );
+            }
+        }
+
+        if send_default_pii {
+            if !log.attributes.contains_key("user.id") {
+                if let Some(id) = self.user().and_then(|user| user.id.as_ref()) {
```
Swatinem: maybe move the `if let Some(user) = self.user()` one level up, and then check for the `contains_key` within that block.
I think that might be a bit more efficient. Though I’m unsure if `self.user()` returns a reference or a clone.

Swatinem (approved):

sentry-core/src/client.rs
lcian: Yeah that's true. Unfortunately this is what the spec mandates, every single log should have these default attributes.

lcian: Do you see any way of optimizing it? @Swatinem 
Maybe we could pre populate a Map with some of the default attributes and store it somewhere to reuse it, but we would still need to clone it for every log.

Swatinem: I was thinking about using a more optimized string type (like my own https://docs.rs/smol_buf/latest/smol_buf/) that can avoid clones in favor of refcounting or small string optimization.
That would cut down on memory usage and allocations a bit. Though the hashmap might still be heavyweight, and hard to avoid.

Not sure these things are worth the effort though? Considering that those would appear everywhere in public types, so it would likely cause some pain for users as well.

lcian: Yeah, I don't know.
I was also thinking of optimizations using `Cow`s, providing special fields to store the default attributes that would take in a `Cow`. This way we wouldn't clone.
But the special fields would need to be `pub` anyway, because the struct lives in `sentry-types`.

lcian: I think we can merge it as is for now and we can always think about optimizing in the future.
However I agree it kinda sucks we need to attach this to every single log.

PR #827, opened 2025-06-04
feat(logs): add macro-based API

lcian:
Adds logging macros.

As described in the doc strings, there are technically 4 variants:
- logging a simple message
- logging a message with format string and args, forwards to `format!`
- the two cases above but with any number of attributes coming before the message and args. The syntax is similar to the one used in `tracing`, but we only support the values that can be converted to a `serde_json::Value`.
Full example:
```rust
warn!(
    error_code = 500,
    user.id = "12345",
    user.email = "test@test.com",
    success = false,
    "Error occurred: {}",
    "bad input"
);
```
Attributes containing a `.` will be nested in the UI.

The biggest difference with this and the `tracing` API is that we don't currently support an arbitrary struct as attribute. The way it works there is that it will work with any type implementing `Display`, calling it to get a string representation of the value, or with the special syntax `?` it will call the `Debug` impl.

Initially I split this into various internal macros but then they would be exported too, so I've refactored it to only export the needed APIs. This leads to some repetition in the various branches.

The spec says the logging API should be available under a `logger namespace`.
However it doesn't look there's a way to re-export the macros only under a specific `mod`, it seems that with `macro_export` they will always be exported at the crate level too (e.g. `sentry::error!`).
So for now I'm just exporting them at top level.


sentry-core/src/logs.rs
```diff
@@ -0,0 +1,396 @@
...
+    (@internal $attrs:ident, $level:expr, $key:ident = $value:expr, $($rest:tt)+) => {{
+        $attrs.insert(
+            stringify!($key).to_owned(),
+            $crate::protocol::LogAttribute($crate::protocol::Value::from($value))
+        );
+        $crate::log!(@internal $attrs, $level, $($rest)+)
+    }};
+
+    // Attributes recursive case: key with 1 dot
+    (@internal $attrs:ident, $level:expr, $key1:ident . $key2:ident = $value:expr, $($rest:tt)+) => {{
```
lcian: Here we need to explicitly specify all the variants. The UI currently supports up to 4 dots, I have added up to 8 just in case.
I have tried to do it recursively using `$key:tt` but that lead to ambiguity with the other remaining arguments

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,396 @@
...
+/// # Examples
+///
+/// ```
+/// use sentry::{log, protocol::LogLevel};
+///
+/// // Simple message
+/// log!(LogLevel::Info, "Hello world");
+///
+/// // Message with format args
+/// log!(LogLevel::Debug, "Value is {}", 42);
```
ghost: The main `log!` macro implementation is robust but could be improved with additional error handling. Consider adding defensive measures for handling potential failures when capturing logs, especially in the case where the Hub is not properly initialized or doesn't have a client bound to it.

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,396 @@
...
+///
+/// // Message with format args and attributes
+/// log!(LogLevel::Warn,
+///     error_code = 500,
+///     user.id = "12345",
```
ghost: In all the documentation examples, you're using `use sentry::*` imports, but it would be better to show the proper import path for each specific macro. This provides clearer guidance for users on how to import exactly what they need.
```suggestion
/// # Examples
///
/// ```
/// use sentry::protocol::LogLevel;
/// use sentry::log;
///
/// // Simple message
/// log!(LogLevel::Info, "Hello world");
///
/// // Message with format args
/// log!(LogLevel::Debug, "Value is {}", 42);
///
/// // Message with format args and attributes
/// log!(LogLevel::Warn,
///     error_code = 500,
///     user.id = "12345",
///     user.email = "test@test.com",
///     success = false,
///     "Error occurred: {}",
///     "bad input"
/// );
/// ```
```

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,396 @@
...
+
+    // Attributes recursive case: key with 8 dots
+    (@internal $attrs:ident, $level:expr, $key1:ident . $key2:ident . $key3:ident . $key4:ident . $key5:ident . $key6:ident . $key7:ident . $key8:ident . $key9:ident = $value:expr, $($rest:tt)+) => {{
+        $attrs.insert(
+            concat!(stringify!($key1), ".", stringify!($key2), ".", stringify!($key3), ".", stringify!($key4), ".", stringify!($key5), ".", stringify!($key6), ".", stringify!($key7), ".", stringify!($key8), ".", stringify!($key9)).to_owned(),
+            $crate::protocol::LogAttribute($crate::protocol::Value::from($value))
+        );
+        $crate::log!(@internal $attrs, $level, $($rest)+)
+    }};
+}
```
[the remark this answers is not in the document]
lcian: It's documented and the code will fail to compile if more than 8 dots are used.

sentry/tests/test_basic.rs
```diff
@@ -313,3 +313,268 @@ fn test_basic_capture_log() {
...
+            _ => panic!("expected logs"),
+        },
+        _ => panic!("expected item container"),
+    }
+}
+
+#[cfg(feature = "UNSTABLE_logs")]
+#[test]
+fn test_basic_capture_log_macro_message_formatted() {
+    use sentry::protocol::LogAttribute;
```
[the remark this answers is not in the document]
lcian: There's no need to test that

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,396 @@
...
+///     error_code = 500,
+///     user.id = "12345",
+///     user.email = "test@test.com",
+///     success = false,
+///     "Error occurred: {}",
+///     "bad input"
+/// );
+/// ```
+#[macro_export]
+macro_rules! trace {
```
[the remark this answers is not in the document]
lcian: This way we can have different documentation for each of those.

sentry-core/src/logs.rs
lcian: We're already using those

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,392 @@
+/// Captures a log at the given level, with the given message and attributes.
+/// The attributes are passed as `key = value` arguments before the message, which can be a simple string or a format string with its arguments.
+///
+/// The supported attribute keys are all valid Rust identifiers with up to 8 dots.
+/// Using dots will nest multiple attributes under their common prefix in the UI.
+///
+/// The supported attribute values are simple types, such as string, numbers, and boolean.
+///
+/// See also the [`crate::trace!`], [`crate::debug!`], [`crate::info!`], [`crate::warn!`], [`crate::error!`], and [`crate::fatal!`] macros, which call `log!` with the corresponding level.
```
Swatinem: as these will get rendered to rustdoc, using ``[`sentry::trace!`](crate::trace!)`` should probably do the trick of generating a nice looking (and correct) link.

sentry-core/src/logs.rs
Swatinem: curious how `tracing` is solving this?

Swatinem (approved):

sentry-core/src/lib.rs
```diff
@@ -132,6 +132,8 @@ pub use crate::intodsn::IntoDsn;
 pub use crate::performance::*;
 pub use crate::scope::{Scope, ScopeGuard};
 pub use crate::transport::{Transport, TransportFactory};
+#[cfg(feature = "UNSTABLE_logs")]
+mod logs; // logs macros exported with `#[macro_export]`
```
szokeasaurusrex: Let's rename this module to `logger` to remain at [least somewhat consistent with the specs](https://develop.sentry.dev/sdk/telemetry/logs/#logger-module).
 
```suggestion
mod logger; // logs macros exported with `#[macro_export]`
```

sentry-core/src/logs.rs
lcian: I think they might use a better way to solve it, I'll spend some more time looking at their code today to see if we can improve this.

sentry-core/src/logs.rs
szokeasaurusrex, line 1: As I mentioned offline, I think we should make some effort to come up with a solution that would allow these macros to be nested under something other than the top-level `sentry` crate's API. Having them under the top-level API may confuse users, as for example, the `sentry::error!` macro could get confused with the `sentry::capture_error` function; it is unclear from the name which of those two is related to Sentry logs, and which one would capture a Sentry error.

As long as these macros remain guarded under an unstable feature, I think it is okay to keep the macros under the top-level API; however, I would still add a `logger_` prefix to each of the macro's names.

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,392 @@
...
+///     error_code = 500,
+///     user.id = "12345",
+///     user.email = "test@test.com",
+///     success = false,
+///     "Error occurred: {}",
+///     "bad input"
+/// );
+/// ```
+#[macro_export]
+macro_rules! log {
```
szokeasaurusrex:
```suggestion
macro_rules! logger_log {
```

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,392 @@
...
+///     error_code = 500,
+///     user.id = "12345",
+///     user.email = "test@test.com",
+///     success = false,
+///     "Error occurred: {}",
+///     "bad input"
+/// );
+/// ```
+#[macro_export]
+macro_rules! trace {
```
szokeasaurusrex:
```suggestion
macro_rules! logger_trace {
```

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,392 @@
...
+///     error_code = 500,
+///     user.id = "12345",
+///     user.email = "test@test.com",
+///     success = false,
+///     "Error occurred: {}",
+///     "bad input"
+/// );
+/// ```
+#[macro_export]
+macro_rules! debug {
```
szokeasaurusrex:
```suggestion
macro_rules! logger_debug {
```

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,392 @@
...
+///     error_code = 500,
+///     user.id = "12345",
+///     user.email = "test@test.com",
+///     success = false,
+///     "Error occurred: {}",
+///     "bad input"
+/// );
+/// ```
+#[macro_export]
+macro_rules! info {
```
szokeasaurusrex:
```suggestion
macro_rules! logger_info {
```

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,392 @@
...
+///     error_code = 500,
+///     user.id = "12345",
+///     user.email = "test@test.com",
+///     success = false,
+///     "Error occurred: {}",
+///     "bad input"
+/// );
+/// ```
+#[macro_export]
+macro_rules! warn {
```
szokeasaurusrex:
```suggestion
macro_rules! logger_warn {
```

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,392 @@
...
+///     error_code = 500,
+///     user.id = "12345",
+///     user.email = "test@test.com",
+///     success = false,
+///     "Error occurred: {}",
+///     "bad input"
+/// );
+/// ```
+#[macro_export]
+macro_rules! error {
```
szokeasaurusrex:
```suggestion
macro_rules! logger_error {
```

sentry-core/src/logs.rs
```diff
@@ -0,0 +1,392 @@
...
+///     error_code = 500,
+///     user.id = "12345",
+///     user.email = "test@test.com",
+///     success = false,
+///     "Error occurred: {}",
+///     "bad input"
+/// );
+/// ```
+#[macro_export]
+macro_rules! fatal {
```
szokeasaurusrex:
```suggestion
macro_rules! logger_fatal {
```

sentry/tests/test_basic.rs
szokeasaurusrex, line 1: will these new tests get run in CI, even though they are guarded by the feature flag?

CHANGELOG.md
```diff
@@ -17,11 +17,11 @@
...
 - feat(logs): add log protocol types (#821) by @lcian
-  - Basic types for [Sentry structured logs](https://docs.sentry.io/product/explore/logs/) have been added.
 - feat(logs): add ability to capture and send logs (#823) by @lcian
-  - A method `capture_log` has been added to the `Hub` to enable sending logs.
-  - This is gated behind the `UNSTABLE_logs` feature flag (disabled by default).
-  - Additionally, the new client option `enable_logs` needs to be enabled for logs to be sent to Sentry.
+- feat(logs): add macro-based API (#827) by @lcian
+  - Support for [Sentry structured logs](https://docs.sentry.io/product/explore/logs/) has been added.
+  - To enable logs, enable the `UNSTABLE_logs` feature of the `sentry` crate and set `enable_logs` to `true` in your client options.
+  - Then, use the `trace!`, `debug!`, `info!`, `warn!`, `error!` and `fatal!` macros to capture logs.
```
szokeasaurusrex: If you take my suggestion regarding the `logger_` prefix, also update the changelog

```suggestion
  - Then, use the `logger_trace!`, `logger_debug!`, `logger_info!`, `logger_warn!`, `logger_error!` and `logger_fatal!` macros to capture logs.
```

lcian: Thanks! 
I think I would prefer the solution where we create a separate crate and export the macros from there with `#[macro_export]`, then reexport those under a module in the top-level `sentry` crate.
I think that should allow us to get the macros exported only under the module.
Let me give it a try.

sentry/tests/test_basic.rs
lcian: Yeah, we are running them with
```
cargo test --workspace --all-features --all-targets
```
see https://github.com/getsentry/sentry-rust/actions/runs/15448954690/workflow?pr=827#L61

sentry-core/src/logs.rs
Swatinem: A while ago, there was talk about "macros 2.0" which would act more similar to normal items in the sense that they follow the same export/import rules, etc.
Though I believe that initiative has stalled since then and I haven’t seen any progress there in quite a while.

CHANGELOG.md
lcian: This doesn't work because the macros need to access `$crate::protocol::Log`, `$crate::Hub`, etc which are defined in `sentry-core`.
We cannot add `sentry-core` as a dependency of the macros crate, because that would be a circular dependency.
If we hardcode the paths in the macro, e.g. `sentry_core::Hub`, then this won't work when a user only imports the top-level `sentry` crate and doesn't depend on `sentry_core` directly...

sentry-core/src/logs.rs
lcian: Yeah, this is the tracking issue https://github.com/rust-lang/rust/issues/39412
There doesn't seem to be much progress unfortunately

sentry-core/src/logs.rs
lcian: Yeah actually I was wrong, fixed it.
We can use
```
$($key:ident).+ = $value:expr,
```
This way we ensure that we are matching on an attribute and there is no ambiguity between this and the following `$rest:tt`.
And `$($key:ident).+` will match any number of identifier separated by dots, which is what we want.

PR #830, opened 2025-06-06
feat(logs): send logs in batches

lcian:
Implements a batcher for logs.
- The implementation is basically the same as the one for the `SessionFlusher`, with different thresholds.
- As an optimization, we don't even create the batcher if `enable_logs` is false.
- The flushing is implemented as per the [spec](https://develop.sentry.dev/sdk/telemetry/logs/#buffering) initial suggestion of flushing every 100 items or 5 seconds. We can also use different thresholds though.
- In the future we are going to change this implementation to use the [batch processor](https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/) approach, which will allow us to batch logs, spans (for span streaming) and possibly other telemetry data.

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -416,11 +416,7 @@ impl Client {
         self.set_log_default_attributes(&mut log);
 
         if let Some(ref func) = self.options.before_send_log {
-            if let Some(res) = func(log) {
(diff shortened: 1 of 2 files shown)


lcian:
Closing this and making a clean one

PR #831, opened 2025-06-06
feat(logs): send logs in batches

lcian:
Implements a batcher for logs.
- The implementation is basically the same as the one for the `SessionFlusher`, with different thresholds.
- As an optimization, we don't even create the batcher if `enable_logs` is false.
- The flushing is implemented as per the [spec](https://develop.sentry.dev/sdk/telemetry/logs/#buffering) initial suggestion of flushing every 100 items or 5 seconds. We could also use different thresholds though.
- In the future we are going to change this implementation to use the [batch processor](https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/) approach, which will allow us to batch logs, spans (for span streaming) and possibly other telemetry data.

#skip-changelog

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -5,14 +5,14 @@ use std::panic::RefUnwindSafe;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
-#[cfg(feature = "UNSTABLE_logs")]
-use crate::protocol::EnvelopeItem;
 #[cfg(feature = "release-health")]
 use crate::protocol::SessionUpdate;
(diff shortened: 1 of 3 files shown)


lcian:
> nice, looks good.
> 
> > we are going to change this implementation to use the [batch processor](https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/) approach
> 
> Looking forward to this. would this mean that we end up with only one background thread (or maybe two if we count the transport thread)?

Yeah, exactly. At least for logs and spans, but I don't see why we wouldn't use it for sessions as well.

PR #835, opened 2025-06-06
Update ureq to 3.x (#834)

### LINKED ISSUE #834: Update to ureq 3.x
### Description

DISCLAIMER: I'm the creator of ureq.

ureq 3.x has been out for about 4 months. The 2.x branch is end-of-the-line (We will do important security updates). The new version is much better structured and I consider a more solid foundation going forward.

I suggest we update from 2.x to 3.x (PR incoming).

algesten:
Update from ureq 2.x to 3.x. In ureq 3.x, rustls API is in itself not
exposed. That simplifies some of the TLS config. Furthermore the
webpki-roots are bundled and updated using patch-versions.

Close #834

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -437,6 +437,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "base64ct"
+version = "1.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
+
+[[package]]
 name = "bindgen"
 version = "0.71.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -4158,6 +4204,24 @@ dependencies = [
  "wasm-bindgen",
 ]
 
+[[package]]
+name = "webpki-root-certs"
```
Swatinem, line 4208: just curious, what is the difference to `webpki-roots`? or is this complementary?

Maybe its a good idea to run a `cargo update` as well, to harmonize the versions, and give cargo an opportunity to maybe remove some duplicates.

Swatinem (approved):
nice, the code looks a lot simpler and there is less manual steps involved 🎉 

IMO this should be good to go, but a `cargo update` can’t hurt.

Swatinem:
Looks like `base64ct` has already updated to rust 2024. Maybe you can use the MSRV-aware solver to make sure we downgrade to a compatible version?

Cargo.lock
algesten: `webpki-roots` is optimized for rustls, while `webpki-root-certs` is more generic so it can be used with other TLS backends.

algesten:
Looks good to me!

PR #837, opened 2025-06-10
fix(logs): send environment in `sentry.environment` default attribute

lcian:
Writes the `environment` in the right place.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3137,7 +3137,7 @@ dependencies = [
 
 [[package]]
 name = "sentry"
-version = "0.38.1"
+version = "0.39.0"
 dependencies = [
  "actix-web",
  "anyhow",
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -3137,7 +3137,7 @@ dependencies = [
 
 [[package]]
 name = "sentry"
-version = "0.38.1"
+version = "0.39.0"
```
Swatinem, line 3140: you can run a `cargo metadata` command in our "bump version" script to avoid this: https://github.com/getsentry/ophio/blob/c14d6497bdd816bdceb9217c0dbb14ca659982d3/scripts/bump-version.sh#L25

Swatinem (approved):

Cargo.lock
lcian: Ah, cool

PR #839, opened 2025-06-12
refactor(tracing): refactor internal code and improve docs

lcian:
- Refactor some code in preparation for the logs integration
- No breaking changes
- Improve the docs a little
- Test that a breadcrumb with errors carries the stringified error chain, behavior we added in https://github.com/getsentry/sentry-rust/pull/768
- It's also a behavioral change because we were previously attaching all error fields in an `errors` field in the JSON data of the breadcrumb, which doesn't make much sense. Now we'll store them in separate fields, retaining the original name

example:
```rust
            tracing::info!(
                database_error = &db_error as &dyn Error,
                parse_error = &parse_error as &dyn Error,
                retry_count = 3,
                connection_timeout = "30s",
                "Multiple errors occurred during database operation"
            );
```

Before:
Screenshot 2025-06-12 at 13 27 15

After:
Screenshot 2025-06-12 at 13 27 41

### DIFF
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -11,16 +11,17 @@ use tracing_subscriber::registry::LookupSpan;
 use super::layer::SentrySpanData;
 use crate::TAGS_PREFIX;
 
-/// Converts a [`tracing_core::Level`] to a Sentry [`Level`]
-fn convert_tracing_level(level: &tracing_core::Level) -> Level {
-    match level {
-        &tracing_core::Level::TRACE | &tracing_core::Level::DEBUG => Level::Debug,
-        &tracing_core::Level::INFO => Level::Info,
-        &tracing_core::Level::WARN => Level::Warning,
-        &tracing_core::Level::ERROR => Level::Error,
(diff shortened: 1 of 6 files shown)


sentry-tracing/src/converters.rs
```diff
@@ -98,11 +107,14 @@ where
...
+/// Records the fields of a [`tracing_core::Event`].
 #[derive(Default)]
 pub(crate) struct FieldVisitor {
-    pub json_values: BTreeMap<String, Value>,
-    pub exceptions: Vec<Exception>,
+    pub(crate) json_values: BTreeMap<String, Value>,
+    pub(crate) exceptions: Vec<Exception>,
+    /// If `true`, stringify and store errors in `self.json_values` under the original field name
+    /// else (default), convert to `Exception`s and store in `self.exceptions`.
+    store_errors_in_values: bool,
```
lcian, line 117: As of this PR, `store_errors_in_values = true` will be used for breadcrumbs.
We will use the same behavior for `logs`, so instead of duplicating code we make it a flag here on the visitor

Swatinem (approved):

PR #840, opened 2025-06-12
feat(tracing): add support for logs

lcian:
Logs support for `sentry-tracing`.

Open question for the future:
Someone might want to capture multiple things (e.g. if you get an error event, capture both a Sentry error and an error level log).
To do this they can provide a custom `event_mapper`, but perhaps we could consider providing an option to map to multiple things at the same time - it looks like that would require 2**n enum variants, or a breaking change.

### DIFF
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -2,7 +2,11 @@ use std::collections::BTreeMap;
 use std::error::Error;
 
 use sentry_core::protocol::{Event, Exception, Mechanism, Value};
+#[cfg(feature = "logs")]
+use sentry_core::protocol::{Log, LogAttribute, LogLevel};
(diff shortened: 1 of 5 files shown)


sentry-tracing/src/converters.rs
```diff
@@ -308,3 +324,43 @@ where
...
+    }
+    if let Some(file) = event_meta.file() {
+        attributes.insert("tracing.file".to_owned(), file.into());
+    }
+    if let Some(line) = event_meta.line() {
+        attributes.insert("tracing.line".to_owned(), line.into());
+    }
+
+    attributes.insert("sentry.origin".to_owned(), "auto.tracing".into());
+
```
lcian, line 357: Unfortunately we cannot get the `sentry.message.template` with tracing.

PR #842, opened 2025-06-13
refactor(logs): cache default attributes and add OS attributes

lcian:
- Cache default attributes and store them on the client
- They are always the same anyway. This way we clone once (from the cached attributes map to the actual log attributes) instead of doing multiple smaller clones.
- Add `os.name` and `os.version` attributes to the log, which are mandatory for native SDKs according to the spec, and personally I think they could be useful.
- The best way I've found to retrieve the OS context in `sentry-core` without making breaking changes or exposing additional API surface is the one I'm proposing, which still sucks.

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -1,5 +1,7 @@
 use std::any::TypeId;
 use std::borrow::Cow;
+#[cfg(feature = "logs")]
+use std::collections::BTreeMap;
 use std::fmt;
 use std::panic::RefUnwindSafe;
 use std::sync::{Arc, RwLock};
@@ -20,6 +22,8 @@ use crate::types::{Dsn, Uuid};
 #[cfg(feature = "release-health")]
 use crate::SessionMode;
 use crate::{ClientOptions, Envelope, Hub, Integration, Scope, Transport};
+#[cfg(feature = "logs")]
+use sentry_types::protocol::v7::Context;
 #[cfg(feature = "logs")]
 use sentry_types::protocol::v7::{Log, LogAttribute};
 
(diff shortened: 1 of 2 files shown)


sentry-core/src/client.rs
```diff
@@ -413,60 +492,18 @@ impl Client {
     fn prepare_log(&self, mut log: Log, scope: &Scope) -> Option<Log> {
         scope.apply_to_log(&mut log, self.options.send_default_pii);
 
-        self.set_log_default_attributes(&mut log);
+        if let Some(default_attributes) = self.default_log_attributes.clone() {
+            for (key, val) in default_attributes.into_iter() {
+                log.attributes.entry(key).or_insert(val);
```
Swatinem: I would iterate over `default_log_attributes` and then clone each key/val separately. This should be cheaper as it avoids having to clone/deallocate the container itself.

lcian: Okay, I thought this would be better because it's just a single call to clone but maybe I'm wrong (and the compiler does whatever it wants anyways). I'll do it like you said.

lcian: > Still reminds me that splitting the SDK up into multiple crates was a really bad decision :-D

I think the `sentry-core` and `sentry` split is useful, maybe it would be better for all integrations to live in `sentry`.

lcian: But I guess this is a topic for v1 😄

lcian: Fwiw this problem of fetching the OS context applies to other SDKs such as Java, there we will add a new method to the `Integration` interface to process logs.
I like this approach we're using here more as it seems like a one-off thing and doesn't require API changes, especially as they wouldn't be aligned across SDKs. 
We can always revisit in the future.

PR #847, opened 2025-06-18
feat(tracing): support combined EventFilters and EventMappings

lcian:
It seems common enough that someone would want to want to map  a `tracing` event to multiple item types, and right now that's not possible, you have to choose between Ignore/Event/Breadcrumb/Log.
For example, in Relay we want to send everything at or above INFO level to logs, while still sending ERROR events to Sentry events at the same time.

This is a proposal for how to implement that.
This particular solution requires 2 breaking changes, even though both could be avoided.
I would argue that they are not actually breaking for most users though, i.e. those that just set up a custom filter, as this is still backwards compatible in terms of syntax, even though the underlying type changes.

### DIFF
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -77,7 +77,7 @@ fn extract_event_data(
 /// Extracts the message and metadata from an event, including the data in the current span.
 fn extract_event_data_with_context<S>(
     event: &tracing_core::Event,
-    ctx: Option<Context<S>>,
+    ctx: Option<&Context<S>>,
     store_errors_in_values: bool,
 ) -> (Option<String>, FieldVisitor)
 where
@@ -182,7 +182,7 @@ impl Visit for FieldVisitor {
 /// Creates a [`Breadcrumb`] from a given [`tracing_core::Event`].
 pub fn breadcrumb_from_event<'context, S>(
     event: &tracing_core::Event,
-    ctx: impl Into<Option<Context<'context, S>>>,
+    ctx: impl Into<Option<&'context Context<'context, S>>>,
 ) -> Breadcrumb
 where
     S: Subscriber + for<'a> LookupSpan<'a>,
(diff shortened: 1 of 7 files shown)


sentry-tracing/src/converters.rs
```diff
@@ -182,7 +182,7 @@ impl Visit for FieldVisitor {
 /// Creates a [`Breadcrumb`] from a given [`tracing_core::Event`].
 pub fn breadcrumb_from_event<'context, S>(
     event: &tracing_core::Event,
-    ctx: impl Into<Option<Context<'context, S>>>,
+    ctx: impl Into<Option<&'context Context<'context, S>>>,
```
lcian, line 185: Breaking change here and in the 2 functions below.
This could be avoided by cloning the `Context` in `on_event`.

sentry-tracing/src/layer.rs
```diff
@@ -13,21 +14,22 @@ use tracing_subscriber::registry::LookupSpan;
...
-    /// Ignore the [`Event`]
-    Ignore,
-    /// Create a [`Breadcrumb`] from this [`Event`]
-    Breadcrumb,
-    /// Create a [`sentry_core::protocol::Event`] from this [`Event`]
-    Event,
-    /// Create a [`sentry_core::protocol::Log`] from this [`Event`]
-    #[cfg(feature = "logs")]
-    Log,
+bitflags! {
```
lcian, line 17: Breaking change because we switch from enum to struct.
This could be avoided by adding a `Combined` enum variant and possibly implementing `BitOr` on `EventFilter` if we want the same syntax that bitflags offers.

sentry-tracing/src/layer.rs
```diff
@@ -211,30 +234,44 @@ where
...
+                    items.push(EventMapping::Log(log_from_event(event, span_ctx.as_ref())));
+                }
+                EventMapping::Combined(CombinedEventMapping(items))
             }
         };
-
-        match item {
-            EventMapping::Event(event) => {
-                sentry_core::capture_event(event);
+        let items = CombinedEventMapping::from(items);
```
lcian, line 263: I convert to `CombinedEventMapping` here so that below I can just handle it with the for loop instead of handling single and combined separately.

PR #849, opened 2025-06-23
feat(core): emit debug log when calling capture_log but logs are disabled

lcian:
Maybe this can be useful in case someone forgets to set `enable_logs`.


adinauer (approved):
LGTM

PR #864, opened 2025-07-18
feat: filter username and password in URLs

lcian:
Close https://github.com/getsentry/sentry-rust/issues/516

### DIFF
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -32,6 +32,7 @@ rand = { version = "0.9.0", optional = true }
 sentry-types = { version = "0.41.0", path = "../sentry-types" }
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = { version = "1.0.46" }
+url = { version = "2.1.1", features = ["serde"] }
 uuid = { version = "1.0.0", features = ["v4", "serde"], optional = true }
 
 [dev-dependencies]
--- a/sentry-core/src/utils.rs
+++ b/sentry-core/src/utils.rs
@@ -1,5 +1,7 @@
 //! Utilities reused across dependant crates and integrations.
 
+use url::Url;
+
 const SENSITIVE_HEADERS_UPPERCASE: &[&str] = &[
     "AUTHORIZATION",
     "PROXY_AUTHORIZATION",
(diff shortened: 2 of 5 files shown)


sentry-core/src/utils.rs
```diff
@@ -10,8 +10,21 @@ const SENSITIVE_HEADERS_UPPERCASE: &[&str] = &[
...
     SENSITIVE_HEADERS_UPPERCASE.contains(&name.to_ascii_uppercase().replace("-", "_").as_str())
 }
+
+/// Scrub PII (username and password) from the given URL.
+pub fn scrub_pii_from_url(mut url: url::Url) -> url::Url {
+    if !url.username().is_empty() {
+        let _ = url.set_username(PII_REPLACEMENT);
+    }
+    if url.password().is_some() {
+        let _ = url.set_password(Some(PII_REPLACEMENT));
```
lcian, line 29: The cases in which these can fail are documented as follows:
```
If this URL is cannot-be-a-base or does not have a host, do nothing and return `Err`.
```

This happens for instance if the URL is relative. In these cases we just ignore the errors.

szokeasaurusrex: Please add this information as a code comment, as it could be helpful for future developers

sentry-core/Cargo.toml
```diff
@@ -32,6 +32,7 @@ rand = { version = "0.9.0", optional = true }
 sentry-types = { version = "0.41.0", path = "../sentry-types" }
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = { version = "1.0.46" }
+url = { version = "2.1.1" }
```
szokeasaurusrex, line 35: [nit] Why not specify the latest version here? We already appear to be using 2.5.4 in Cargo.lock.

```suggestion
url = { version = "2.5.4" }
```

szokeasaurusrex (approved):
Mostly looks good, please address my inline comments before merging.

Also, we discussed this offline: as a further improvement (so not blocking this PR), I would suggest enforcing the URL stripping via the type system, by creating a new type called something like `SafeUrl` or `StrippedUrl`, which would wrap the existing `Url` type, and requiring the `SafeUrl` to be used when sending URLs to Sentry. This `SafeUrl` would ensure via the constructor that any URLs it is wrapping are stripped. This avoids the possibility of a future SDK dev forgetting to strip the URL

sentry-core/Cargo.toml
lcian: I copy pasted this from `sentry-types` where the dependency is on `2.1.1` in the `Cargo.toml`.

Issue #866, opened 2025-07-21
Add more envelope headers


lcian:
Right now we just send `event_id`, there's some recommended headers we should add<br>[https://develop.sentry.dev/sdk/data-model/envelopes/#envelope-headers](https://develop.sentry.dev/sdk/data-model/envelopes/#envelope-headers)

Issue #865, opened 2025-07-21
Send some DSC information in envelope header


lcian:
For correct extrapolation of span metrics we need some dynamic sampling information to be included in (transaction) envelope headers (`trace` envelope header). 

Specifically:

* `trace_id`
* `public_key`
* `sample_rate`
* `sampled`

PR #867, opened 2025-07-22
feat(types): add all the missing supported envelope headers

lcian:
Adds the [documented envelope headers](https://develop.sentry.dev/sdk/data-model/envelopes/#envelope-headers) to the types.
In a separate PR we can start to populate the new values.

Closes https://github.com/getsentry/sentry-rust/issues/866
Required for https://github.com/getsentry/sentry-rust/issues/865

#skip-changelog

### DIFF
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -1,8 +1,4 @@
-use std::{
-    io::{BufRead, Write},
-    path::Path,
-    time::SystemTime,
-};
+use std::{io::Write, path::Path, time::SystemTime};
 
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -35,7 +31,7 @@ pub enum EnvelopeError {
     MissingNewline,
     /// Invalid envelope header
     #[error("invalid envelope header")]
-    InvalidHeader(#[source] Box<dyn std::error::Error>),
+    InvalidHeader(#[source] serde_json::Error),
     /// Invalid item header
     #[error("invalid item header")]
     InvalidItemHeader(#[source] serde_json::Error),
@@ -523,16 +519,18 @@ impl Envelope {
     }
 
     fn parse_headers(slice: &[u8]) -> Result<(EnvelopeHeaders, usize), EnvelopeError> {
-        let mut lines = slice.lines();
-        let first_line = lines
+        let first_line = slice
+            .split(|b| *b == b'\n')
             .next()
-            .ok_or(EnvelopeError::MissingHeader)?
-            .map_err(|err| EnvelopeError::InvalidHeader(Box::new(err)))?;
+            .ok_or(EnvelopeError::MissingHeader)?;
 
-        let headers: EnvelopeHeaders = serde_json::from_str(first_line.as_str())
-            .map_err(|err| EnvelopeError::InvalidHeader(Box::new(err)))?;
+        let headers: EnvelopeHeaders =
+            serde_json::from_slice(first_line).map_err(EnvelopeError::InvalidHeader)?;
 
-        Ok((headers, first_line.len() + 1))
+        let offset = first_line.len();
+        Self::require_termination(slice, offset)?;
+
+        Ok((headers, offset + 1))
     }
 
     fn parse_items(slice: &[u8], mut offset: usize) -> Result<Vec<EnvelopeItem>, EnvelopeError> {
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -1,4 +1,8 @@
-use std::{io::Write, path::Path, time::SystemTime};
+use std::{
+    io::{BufRead, Write},
+    path::Path,
+    time::SystemTime,
+};
 
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
@@ -519,13 +523,17 @@ impl Envelope {
     }
 
     fn parse_headers(slice: &[u8]) -> Result<(EnvelopeHeaders, usize), EnvelopeError> {
-        let first_line = slice
-            .splitn(2, |b| *b == b'\n')
+        let mut lines = slice.lines();
+        let first_line = lines
             .next()
-            .ok_or(EnvelopeError::MissingNewline)?;
+            .ok_or(EnvelopeError::MissingHeader)?
+            .map_err(|_| EnvelopeError::MissingHeader)?;
+        if lines.next().is_none() {
+            return Err(EnvelopeError::MissingNewline);
+        }
 
         let headers: EnvelopeHeaders =
-            serde_json::from_slice(first_line).map_err(EnvelopeError::InvalidHeader)?;
+            serde_json::from_str(first_line.as_str()).map_err(EnvelopeError::InvalidHeader)?;
 
         Ok((headers, first_line.len() + 1))
     }
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -7,7 +7,7 @@ use uuid::Uuid;
 use crate::utils::ts_rfc3339_opt;
 use crate::Dsn;
 
-use super::v7::{self as protocol};
+use super::v7 as protocol;
 
 use protocol::{
     Attachment, AttachmentType, ClientSdkInfo, DynamicSamplingContext, Event, Log, MonitorCheckIn,
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -1,10 +1,13 @@
-use std::{io::Write, path::Path};
+use std::{io::Write, path::Path, time::SystemTime};
 
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use uuid::Uuid;
 
-use super::v7 as protocol;
+use crate::utils::ts_rfc3339_opt;
+use crate::Dsn;
+
+use super::v7::{self as protocol, ClientSdkInfo, DynamicSamplingContext};
 
 use protocol::{
     Attachment, AttachmentType, Event, Log, MonitorCheckIn, SessionAggregates, SessionUpdate,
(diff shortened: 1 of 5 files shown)


lcian:
@sentry review

sentry-types/src/protocol/envelope.rs
```diff
@@ -37,9 +40,23 @@ pub enum EnvelopeError {
     InvalidItemPayload(#[source] serde_json::Error),
 }
 
-#[derive(Deserialize)]
-struct EnvelopeHeader {
+/// The supported [Sentry Envelope Headers](https://develop.sentry.dev/sdk/data-model/envelopes/#headers).
+#[derive(Clone, Default, Debug, Serialize, Deserialize, PartialEq)]
+struct EnvelopeHeaders {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     event_id: Option<Uuid>,
```
szokeasaurusrex, line 47: isn't `event_id` supposed to be required?

In any case, I think it could be potentially risky to add the `skip_serializing_if` here

sentry-types/src/protocol/envelope.rs
```diff
@@ -391,11 +408,8 @@ impl Envelope {
         };
 
         // write the headers:
-        let event_id = self.uuid();
-        match event_id {
-            Some(uuid) => writeln!(writer, r#"{{"event_id":"{uuid}"}}"#)?,
-            _ => writeln!(writer, "{{}}")?,
-        }
+        serde_json::to_writer(&mut writer, &self.headers)?;
+        writer.write_all(b"\n")?;
```
szokeasaurusrex: Why not use `write!` macro (I think that is what it is called)?

sentry-types/src/protocol/envelope.rs
```diff
@@ -37,9 +40,23 @@ pub enum EnvelopeError {
     InvalidItemPayload(#[source] serde_json::Error),
 }
 
-#[derive(Deserialize)]
-struct EnvelopeHeader {
+/// The supported [Sentry Envelope Headers](https://develop.sentry.dev/sdk/data-model/envelopes/#headers).
+#[derive(Clone, Default, Debug, Serialize, Deserialize, PartialEq)]
```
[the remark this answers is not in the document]
lcian: We'll keep it private until we need to make it pub

sentry-types/src/protocol/envelope.rs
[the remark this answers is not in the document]
lcian: The tests already cover this.

sentry-types/src/protocol/envelope.rs
```diff
@@ -504,19 +518,16 @@ impl Envelope {
         Self::from_bytes_raw(bytes)
     }
 
-    fn parse_header(slice: &[u8]) -> Result<(EnvelopeHeader, usize), EnvelopeError> {
-        let mut stream = serde_json::Deserializer::from_slice(slice).into_iter();
+    fn parse_headers(slice: &[u8]) -> Result<(EnvelopeHeaders, usize), EnvelopeError> {
+        let first_line = slice
+            .splitn(2, |b| *b == b'\n')
```
szokeasaurusrex: You should be able to use `.lines` here. You need to import the `BufRead` trait

```suggestion
            .lines()
```

Imo `lines` is a lot easier to understand than this `splitn` call (in any case, you could also just do a normal `split`. Iterators are evaluated lazily, so there should be no performance difference)

sentry-types/src/protocol/v7.rs
```diff
@@ -2336,3 +2336,67 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    // Strictly required fields
+    // Still typed as optional, as when deserializing an envelope created by an older SDK they might still be missing
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    trace_id: Option<TraceId>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    public_key: Option<String>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "display_from_str_opt"
```
szokeasaurusrex, line 2406: Why use `display_from_str_opt` here (or on any of the other `Option<f32>` and `Option<bool>` items), rather than the defualt serializer?

sentry-types/src/utils.rs
```diff
@@ -189,6 +189,38 @@ pub mod ts_rfc3339_opt {
     }
 }
 
+// Serialize and deserialize the inner value into/from a string using the `ToString`/`FromStr` implementation.
+pub mod display_from_str_opt {
```
szokeasaurusrex: Does this need to be `pub` or would `pub(crate)` be enough?


```suggestion
pub(crate) mod display_from_str_opt {
```

sentry/src/transports/thread.rs
```diff
@@ -7,6 +7,7 @@ use std::time::Duration;
 use super::ratelimit::{RateLimiter, RateLimitingCategory};
 use crate::{sentry_debug, Envelope};
 
+#[expect(clippy::large_enum_variant)]
```
szokeasaurusrex, line 10: I'm not sure we should ignore this lint, seems to be a good reasoning behind why it exists: https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant

I am fine with keeping this if we expect most tasks to contain `SendEnvelope`. But if not, I believe we can get around this warniong by boxing the `Envelope`

sentry/src/transports/thread.rs
```diff
@@ -7,6 +7,7 @@ use std::time::Duration;
...
 enum Task {
     SendEnvelope(Envelope),
```
szokeasaurusrex, line 12: See comment above

```suggestion
    SendEnvelope(Box<Envelope>),
```

sentry/src/transports/tokio_thread.rs
```diff
@@ -7,6 +7,7 @@ use std::time::Duration;
 use super::ratelimit::{RateLimiter, RateLimitingCategory};
 use crate::{sentry_debug, Envelope};
 
+#[expect(clippy::large_enum_variant)]
 enum Task {
     SendEnvelope(Envelope),
```
szokeasaurusrex, line 12: See my other comment on `thread.rs`

```suggestion
enum Task {
    SendEnvelope(Box<Envelope>),
```

sentry-types/src/protocol/v7.rs
lcian: According to the spec, these are supposed to be sent as strings in the JSON: https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#dsc-specification

By looking more closely at the code in Relay, it seems that we could send them as their proper type (bool, number) too
https://github.com/getsentry/relay/blob/8e929cc64c3ce3ae518b20e938b47c04fb0ddda3/relay-sampling/src/dsc.rs#L395

However I've noticed that other SDKs (Java, Python are the ones I've checked) just send these as strings. As this is also the approach documented in the dev docs, I would align with it.

sentry-types/src/utils.rs
lcian: We don't reexport these so they're practically `pub(crate)` at the moment, but might as well make it explicitly `pub(crate)`.

sentry/src/transports/thread.rs
lcian: We only call flush on panic, and shutdown on client close, so in most cases this is going to be `SendEnvelope`, keeping it like this.

sentry-types/src/protocol/v7.rs
lcian: Changed the `sample_rand` serialization to match the spec (exactly 6 decimal digits) after our discussion.

sentry-types/src/protocol/envelope.rs
lcian: For envelopes containing only certain item types it's optional, such as for sessions (see the test `test_session` in this file).
With `skip_serializing_if = "Option::is_none"` we'll skip the field if it's none, matching the previous behavior where in `Envelope::to_writer` we had the following:
```rust
        let event_id = self.uuid();
        match event_id {
            Some(uuid) => writeln!(writer, r#"{{"event_id":"{uuid}"}}"#)?,
            _ => writeln!(writer, "{{}}")?,
        }
```
This is also covered by the tests that were already in place.

sentry-types/src/protocol/envelope.rs
lcian: Updated. Looks like this is slightly worse because `split` returns `Result<String>` meaning it allocates, `splitn` just returns subslices, so I would prefer to keep `splitn`.

sentry-types/src/protocol/envelope.rs
```diff
@@ -1,14 +1,21 @@
...
 
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use uuid::Uuid;
 
-use super::v7 as protocol;
+use crate::utils::ts_rfc3339_opt;
+use crate::Dsn;
+
+use super::v7::{self as protocol};
```
szokeasaurusrex: Not sure why this change is needed (surprising though if this wasn't caught by rustfmt or clippy)

```suggestion
use super::v7 as protocol;
```

sentry-types/src/protocol/envelope.rs
```diff
@@ -28,7 +35,7 @@ pub enum EnvelopeError {
     MissingNewline,
     /// Invalid envelope header
     #[error("invalid envelope header")]
-    InvalidHeader(#[source] serde_json::Error),
+    InvalidHeader(#[source] Box<dyn std::error::Error>),
```
szokeasaurusrex: This seems like a (potentially disruptive) breaking change. Why is it needed to change the type here?

sentry-types/src/protocol/envelope.rs
szokeasaurusrex: Okay, was lacking this context

sentry-types/src/protocol/envelope.rs
```diff
@@ -391,11 +412,8 @@ impl Envelope {
         };
 
         // write the headers:
-        let event_id = self.uuid();
-        match event_id {
-            Some(uuid) => writeln!(writer, r#"{{"event_id":"{uuid}"}}"#)?,
-            _ => writeln!(writer, "{{}}")?,
-        }
+        serde_json::to_writer(&mut writer, &self.headers)?;
```
szokeasaurusrex, line 411: yeah this is cleaner

sentry-types/src/protocol/envelope.rs
```diff
@@ -504,19 +522,17 @@ impl Envelope {
...
-            None => return Err(EnvelopeError::MissingHeader),
-            Some(Err(error)) => return Err(EnvelopeError::InvalidHeader(error)),
-            Some(Ok(header)) => header,
-        };
+    fn parse_headers(slice: &[u8]) -> Result<(EnvelopeHeaders, usize), EnvelopeError> {
+        let mut lines = slice.lines();
+        let first_line = lines
+            .next()
+            .ok_or(EnvelopeError::MissingHeader)?
+            .map_err(|err| EnvelopeError::InvalidHeader(Box::new(err)))?;
```
szokeasaurusrex: Let's go back to `split` rather than `lines`. I did not realize that `lines` performs allocations and that it returns IO errors. 

But, please use `split` rather than `splitn` – performance-wise, the two should be identical, and the `split` is likely easier to reason about.

sentry-types/src/protocol/v7.rs
```diff
@@ -2336,3 +2336,99 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+/// Context](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/).
+///
+/// Sentry supports sampling at the server level through [Dynamic Sampling](https://docs.sentry.io/organization/dynamic-sampling/).
+/// This feature allows users to specify target sample rates for each project via the frontend instead of requiring an application redeployment.
+/// The backend needs additional information from the SDK to support these features, contained in
+/// the Dynamic Sampling Context.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub(crate) struct DynamicSamplingContext {
+    // Strictly required fields
+    // Still typed as optional, as when deserializing an envelope created by an older SDK they might still be missing
```
szokeasaurusrex, line 2398: I wonder whether maybe we should have a separate struct for serializing and deserializing. For deserializing, the fields would be optional, but when serializing, they would be required. 

Or alternatively, we can check that the strictly required fields are all `Some` when serializing

sentry-types/src/protocol/v7.rs
```diff
@@ -2336,3 +2336,99 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    trace_id: Option<TraceId>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    public_key: Option<String>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "display_from_str_opt"
+    )]
+    sample_rate: Option<f32>,
+    // Required fields
```
szokeasaurusrex, line 2409: What is the difference between "Required" and "Strictly required"? 😅

sentry-types/src/utils.rs
```diff
@@ -189,6 +189,38 @@ pub mod ts_rfc3339_opt {
     }
 }
 
+// Serialize and deserialize the inner value into/from a string using the `ToString`/`FromStr` implementation.
```
szokeasaurusrex: I'd make this a doc comment.

```suggestion
/// Serialize and deserialize the inner value into/from a string using the `ToString`/`FromStr` implementation.
```

Also, adding an example (like stating that the int `1` would get serialized as `"1"` rather than `1`) would help a ton, bonus points if you can add this as a `doctest`

sentry-types/src/protocol/envelope.rs
lcian: I added new imports from `v7` which I later removed because we import `v7` itself, however `rustfmt` will try to put all imports in a single line, resulting in the `use ...{self as x}` syntax.

sentry-types/src/protocol/v7.rs
lcian: Still you could be serializing an envelope that doesn't have those fields (because it was generated by deserializing one that doesn't have them), so we cannot make them non-optional when serializing.

sentry-types/src/protocol/v7.rs
lcian: This is how they're defined in the develop docs.

sentry-types/src/utils.rs
lcian: Done

sentry-types/src/protocol/envelope.rs
lcian: Done

sentry-types/src/utils.rs
lcian: Added the doctest but it needs to be ignored because doctests cannot access something with `pub(crate)` visibility. See e.g. https://stackoverflow.com/questions/70111757/how-can-i-write-documentation-tests-for-private-modules

sentry-types/src/utils.rs
```diff
@@ -189,6 +189,65 @@ pub mod ts_rfc3339_opt {
...
+///     host: Some("localhost".to_string()),
+///     port: Some(8080),
+///     enabled: Some(true),
+/// };
+/// let json = serde_json::to_string(&config).unwrap();
+/// assert_eq!(json, r#"{"host":"localhost","port":"8080","enabled":"true"}"#);
+///
+/// let deserialized: Config = serde_json::from_str(&json).unwrap();
+/// assert_eq!(deserialized, config);
+/// ```
```
szokeasaurusrex, line 219: I forgot you cannot make true doctests for private functions, but thanks for the example, I still think it is helpful

szokeasaurusrex (approved):

PR #868, opened 2025-07-23
feat(types): add setters for envelope headers

lcian:
Modifies visibility where needed, and setters for envelope headers and the DSC (trace header).
Even though these are part of the public API, they are mostly going to be used internally, so I think this level of documentation is fine, especially considering we already have good docs on the structs themselves.

Stacked on: https://github.com/getsentry/sentry-rust/pull/867
Required for: https://github.com/getsentry/sentry-rust/issues/865


sentry-types/src/protocol/envelope.rs
```diff
@@ -59,6 +59,64 @@ struct EnvelopeHeaders {
...
+        Default::default()
+    }
+
+    /// Returns the Event ID.
+    pub fn event_id(&self) -> Option<&Uuid> {
+        self.event_id.as_ref()
+    }
+
+    /// Sets the Event ID.
+    pub fn set_event_id(&mut self, event_id: Option<Uuid>) {
```
szokeasaurusrex: As discussed offline, you could consider using a builder pattern here:


```suggestion
    pub fn set_event_id(mut self, event_id: Option<Uuid>) -> Self {
```

sentry-types/src/protocol/envelope.rs
```diff
@@ -59,6 +59,64 @@ struct EnvelopeHeaders {
...
 }
 
+impl EnvelopeHeaders {
+    /// Creates empty Envelope headers.
+    pub fn new() -> EnvelopeHeaders {
+        Default::default()
+    }
+
+    /// Returns the Event ID.
+    pub fn event_id(&self) -> Option<&Uuid> {
```
szokeasaurusrex: As discussed offline, if we don't expect that the getters will be needed, let's not include them in the public API

sentry-types/src/protocol/envelope.rs
```diff
@@ -37,9 +40,66 @@ pub enum EnvelopeError {
...
+    pub fn new() -> EnvelopeHeaders {
+        Default::default()
+    }
+
+    /// Sets the Event ID.
+    #[must_use]
+    pub fn with_event_id(mut self, event_id: Option<Uuid>) -> Self {
+        self.event_id = event_id;
+        self
+    }
```
szokeasaurusrex: [optional suggestion] If you expect that users never will need to clear the `event_id` field, you can potentially achieve a cleaner API by taking `Uuid` rather than `Option<Uuid>` (so you only have a `None` value when you skip calling this method). Same applies to the other `with_` functions here. I would recommend staying consistent for all of the `with_` functions, regardless of what you decide here

```suggestion
    pub fn with_event_id(mut self, event_id: Uuid) -> Self {
        self.event_id = Some(event_id);
        self
    }
```

sentry-types/src/protocol/envelope.rs
```diff
@@ -317,9 +377,19 @@ impl Envelope {
...
 
+    /// Returns the Envelope headers.
+    pub fn headers(&self) -> &EnvelopeHeaders {
+        &self.headers
+    }
+
+    /// Sets the Envelope headers.
+    pub fn set_headers(&mut self, headers: EnvelopeHeaders) {
+        self.headers = headers
+    }
```
szokeasaurusrex: are these being used anywhere yet?

sentry-types/src/protocol/v7.rs
```diff
@@ -2336,3 +2336,169 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    type Err = std::num::ParseFloatError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        s.parse().map(Self)
+    }
+}
+
+impl std::fmt::Display for SampleRand {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:.6}", self.0)
```
szokeasaurusrex: [change request] `sample_rand`, after applying this formatting, must still satisfy `0 <= sample_rand < 1`. However, I believe this rounding logic could round a value close to 1 (such as `0.9999999`, which has seven 9s after the decimal point) to be rounded up to `1.000000`.

Please check whether this is the case, and add a unit test to verify correct behavior if possible. The correct behavior (to maintain equal probabilities of every six-decimal-digit value) would be to always round down – we should check whether there is a formatting option that allows us to do this

sentry-types/src/protocol/v7.rs
```diff
@@ -2336,3 +2336,169 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+
+impl DynamicSamplingContext {
+    /// Creates an empty Dynamic Sampling Context.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Sets the trace ID.
+    #[must_use]
+    pub fn with_trace_id(mut self, trace_id: Option<TraceId>) -> Self {
```
szokeasaurusrex: [optional suggestion] Same note about having `Option<TraceId>` or `TraceId` in the function signature also applies here.

sentry/src/transports/thread.rs
```diff
@@ -7,6 +7,7 @@ use std::time::Duration;
 use super::ratelimit::{RateLimiter, RateLimitingCategory};
 use crate::{sentry_debug, Envelope};
 
+#[expect(clippy::large_enum_variant)]
 enum Task {
     SendEnvelope(Envelope),
```
szokeasaurusrex: I thought I had mentioned this before, but cannot find the comment.

I believe boxing the Envelope should allow us to avoid this [`large_enum_variant`](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant) warning. 

```suggestion
enum Task {
    SendEnvelope(Box<Envelope>),
```

sentry/src/transports/tokio_thread.rs
```diff
@@ -7,6 +7,7 @@ use std::time::Duration;
 use super::ratelimit::{RateLimiter, RateLimitingCategory};
 use crate::{sentry_debug, Envelope};
 
+#[expect(clippy::large_enum_variant)]
 enum Task {
     SendEnvelope(Envelope),
```
szokeasaurusrex: Also here, let's try to remove the `large_enum_variant` warning if possible.

```suggestion
enum Task {
    SendEnvelope(Box<Envelope>),
```

szokeasaurusrex (changes requested):
Requesting changes, I noticed the `sample_rand` serialization logic could result in rounding up. Also left some lower-prio and optional comments

sentry/src/transports/thread.rs
lcian: As I replied in the other PR, most of the time this will be the `SendEnvelope` variant, we get something else only on shutdown and that shouldn't happen frequently, therefore I would keep the `expect`.

szokeasaurusrex: can you please write a code comment to explain this for future maintainers? Or, I think you can add a `reason` to the `expect`

sentry-types/src/protocol/envelope.rs
lcian: In the PR on top of this one https://github.com/getsentry/sentry-rust/pull/869 and in the serialization to grab the event ID

lcian: Actually I'll change this to be `with_headers` as it's going to be more natural to use and matches the other APIs we're adding here

sentry-types/src/protocol/envelope.rs
lcian: Addressed

sentry-types/src/protocol/v7.rs
lcian: There is no way to choose the rounding, "rounding half-to-even" is always used, meaning that anything greater or equal than 0.9999995 will be serialized as 1.0.
Added a special case and also improved the `SampleRand` type itself, now we only implement fallible conversions and return proper errors if the conversion fails (string is not valid float, or float is outside the specification range for a sample_rand).

sentry-types/src/protocol/v7.rs
lcian: Addressed

sentry-types/src/protocol/v7.rs
```diff
@@ -2336,3 +2336,195 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    /// half-open interval [0.0, 1.0).
+    #[error("sample rand value out of admissible interval [0.0, 1.0)")]
+    OutOfRange,
+}
+
+impl TryFrom<f64> for SampleRand {
+    type Error = InvalidSampleRandError;
+
+    fn try_from(value: f64) -> Result<Self, Self::Error> {
+        if !(0.0..1.0).contains(&value) {
```
lcian, line 2385: Clippy forces me to write it like this

sentry-types/src/protocol/envelope.rs
```diff
@@ -1011,6 +1080,36 @@ some content
...
+
+        let envelope = Envelope::from_slice(bytes);
+        assert!(envelope.is_ok());
+        let envelope = envelope.unwrap();
+        let serialized = to_str(envelope);
+        assert_eq!(bytes, serialized.as_bytes());
+    }
+
+    #[test]
+    fn test_sample_rand_rounding() {
```
lcian, line 1098: The new test

sentry-types/src/protocol/envelope.rs
```diff
@@ -1039,6 +1094,21 @@ some content
         assert_eq!(bytes, serialized.as_bytes());
     }
 
+    #[test]
+    fn test_sample_rand_rounding() {
+        let envelope = Envelope::new().with_headers(
+            EnvelopeHeaders::new().with_trace(
+                DynamicSamplingContext::new()
+                    .with_sample_rand(SampleRand::try_from(0.9999999).unwrap()), // 7 nines
```
szokeasaurusrex: You could add some underscores (they don't change the value) to make it easier to tell that this is seven nines 

```suggestion
                    .with_sample_rand(SampleRand::try_from(0.999_999_9).unwrap()),
```

sentry-types/src/protocol/v7.rs
szokeasaurusrex: that is fine

sentry-types/src/protocol/v7.rs
```diff
@@ -2363,25 +2363,51 @@ impl std::fmt::Display for OrganizationId {
...
 
 impl std::fmt::Display for SampleRand {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{:.6}", self.0)
+        // Special case: "{:.6}" would round values greater than or equal to 0.9999995 to 1.0,
+        // as Rust uses [rounding half-to-even](https://doc.rust-lang.org/std/fmt/#precision).
+        // Round to 0.999999 instead to comply with spec.
+        if self.0 >= 0.9999995 {
+            write!(f, "0.999999")
+        } else {
```
szokeasaurusrex, line 2408: Technically, this makes a value of `"0.999999"` more likely than other values (1.5 in a million, versus 1 in a million for all other values, besides of course `"0.000000"` which has probability 0.5 in a million for the same reason) 😅

The most reliable way would be to convert this to an int and make it zero-padded to six digts and prepend with `0.` but I guess how you have it is probs good enough

sentry/src/transports/thread.rs
```diff
@@ -7,7 +7,10 @@ use std::time::Duration;
 use super::ratelimit::{RateLimiter, RateLimitingCategory};
 use crate::{sentry_debug, Envelope};
 
-#[expect(clippy::large_enum_variant)]
+#[expect(
+    clippy::large_enum_variant,
+    reason = "In normal usage this is usually SendEnvelope, the other variants are only used when the user manually calls transport.flush() or when the transport is shut down."
```
szokeasaurusrex: You should try to wrap this string, maybe otherwise it will cause that issue that prevents rustfmt from formattting the whole file (and it would also probably be more readable)


```suggestion
    reason = "In normal usage this is usually SendEnvelope, \
        the other variants are only used when the user manually calls \
        transport.flush() or when the transport is shut down."
```

sentry/src/transports/tokio_thread.rs
```diff
@@ -7,7 +7,10 @@ use std::time::Duration;
 use super::ratelimit::{RateLimiter, RateLimitingCategory};
 use crate::{sentry_debug, Envelope};
 
-#[expect(clippy::large_enum_variant)]
+#[expect(
+    clippy::large_enum_variant,
+    reason = "In normal usage this is usually SendEnvelope, the other variants are only used when the user manually calls transport.flush() or when the transport is shut down."
```
szokeasaurusrex:
```suggestion
    reason = "In normal usage this is usually SendEnvelope, \  
        the other variants are only used when the user manually calls \  
        transport.flush() or when the transport is shut down." 
```

szokeasaurusrex (approved):
lgtm, thanks for addressing comments! Left a couple more small ones to address before merging

sentry-types/src/protocol/envelope.rs
lcian: ah, nice

PR #869, opened 2025-07-24
feat(core): add some DSC fields to transaction envelope headers

lcian:
Adds the necessary DSC fields to solve the metrics extrapolation issue.
Close https://github.com/getsentry/sentry-rust/issues/865


sentry-core/src/performance.rs
```diff
@@ -601,19 +601,50 @@ fn transaction_sample_rate(
...
+    fn determine_sampling_decision(&self, ctx: &TransactionContext) -> SamplingDecision {
         let client_options = self.options();
-        self.sample_should_send(transaction_sample_rate(
+        let sample_rate = transaction_sample_rate(
             client_options.traces_sampler.as_deref(),
             ctx,
             client_options.traces_sample_rate,
-        ))
+        );
+        let sampled = self.sample_should_send(sample_rate);
```
lcian, line 611: `sample_should_send` is part of public API so can't rename it

sentry-core/src/performance.rs
```diff
@@ -622,6 +653,7 @@ impl Client {
 #[derive(Clone, Debug)]
 pub struct Transaction {
     pub(crate) inner: TransactionArc,
+    metadata: TransactionMetadata,
```
lcian, line 633: I know this is mixed visibility but there's really no reason to make this public or pub(crate)

sentry-core/src/performance.rs
```diff
@@ -682,30 +714,39 @@ impl Transaction {
...
                 transaction: None,
             })),
-        }
+            metadata: TransactionMetadata {
+                sample_rate: sampling_decision.sample_rate,
+            },
+        };
+        // use the field on cfg(not(feature = "client"))
+        let _ = slf.metadata.sample_rate;
+        slf
```
lcian: Alternative is to allow the lint (on that cfg if possible) but I prefer to just do this

sentry-core/src/performance.rs
szokeasaurusrex: I mean, both fields are internal to the crate so I suppose this is okay. Tbh I did not even know that mixed visibility was considered bad

sentry-core/src/performance.rs
```diff
@@ -601,19 +601,50 @@ fn transaction_sample_rate(
...
+        }
     }
 }
 
+/// Some metadata associated with a transaction.
+#[derive(Clone, Debug)]
+struct TransactionMetadata {
+    /// The sample rate used when making the sampling decision for the associated transaction.
+    sample_rate: f32,
+}
```
szokeasaurusrex: What's the purpose of this struct? Why not just have the `sample_rate` as `f32`, especially given that this is private?

sentry-core/src/performance.rs
```diff
@@ -660,15 +692,15 @@ impl<'a> TransactionData<'a> {
...
             Some(client) => (
-                client.is_transaction_sampled(&ctx),
+                client.determine_sampling_decision(&ctx),
                 Some(protocol::Transaction {
                     name: Some(ctx.name),
                     ..Default::default()
                 }),
             ),
-            None => (ctx.sampled.unwrap_or(false), None),
+            None => (SamplingDecision::from(&ctx), None),
```
szokeasaurusrex: I find it confusing that there are two ways here to conver the `TransactionContext` into a  `SamplingDecision`: either via `client.determine_sampling_decision` or `SamplingDecision::from`. I would prefer we unify these two – otherwise, it is super confusing to understand which method you would use when.

sentry-core/src/performance.rs
szokeasaurusrex: I would suggest removing this line, and instead of allowing the lint, you should only compile the `metadata` field when the `client` feature is enabled by guarding it with a `cfg(feature = "client")`

```suggestion
        slf
```

sentry-core/src/performance.rs
```diff
@@ -820,9 +861,16 @@ impl Transaction {
                     transaction.sdk = Some(std::borrow::Cow::Owned(client.sdk_info.clone()));
                     transaction.server_name.clone_from(&opts.server_name);
 
+                    let dsc = protocol::DynamicSamplingContext::new().with_trace_id(Some(inner.context.trace_id))
+                    .with_sample_rate(Some(self.metadata.sample_rate))
+                    .with_public_key(opts.dsn.as_ref().map(|dsn| dsn.public_key().to_owned()))
+                    .with_sampled(Some(inner.sampled));
```
szokeasaurusrex: `rustfmt` is not formatting this file, possibly because of [this issue](https://github.com/rust-lang/rustfmt/issues/3863). I am unsure which line is the culprit here, but you should probably investigate (please fix in a separate PR, though)

sentry-core/src/performance.rs
lcian: Because maybe in the future we need to add more stuff (e.g. for Dynamic Sampling we would probably want to add stuff here)

sentry-core/src/performance.rs
lcian: It is, at least in our [Rust development guidelines](https://develop.sentry.dev/engineering-practices/rust/#field-visibility) :)

sentry-core/src/performance.rs
lcian: I'll just remove the struct and make this into a pair

sentry-core/src/performance.rs
```diff
@@ -604,16 +604,26 @@ fn transaction_sample_rate(
 /// Determine whether the new transaction should be sampled.
 #[cfg(feature = "client")]
 impl Client {
-    fn is_transaction_sampled(&self, ctx: &TransactionContext) -> bool {
+    fn determine_sampling_decision(&self, ctx: &TransactionContext) -> (bool, f32) {
```
szokeasaurusrex, line 604: [nit] Would add this doc comment so folks can see what is in the return value by just hovering over the function name in the IDE

```suggestion
    /// Returns sampling decision and sample rate based on the transaction context
    fn determine_sampling_decision(&self, ctx: &TransactionContext) -> (bool, f32) {
```

sentry-core/src/performance.rs
```diff
@@ -660,15 +672,22 @@ impl<'a> TransactionData<'a> {
...
                     name: Some(ctx.name),
                     ..Default::default()
                 }),
             ),
-            None => (ctx.sampled.unwrap_or(false), None),
+            None => (
+                (
+                    ctx.sampled.unwrap_or(false),
+                    ctx.sampled
+                        .map_or(0.0, |sampled| if sampled { 1.0 } else { 0.0 }),
```
szokeasaurusrex: [`f32::from`](https://doc.rust-lang.org/std/primitive.f32.html#method.from-6) does this conversion already

```suggestion
                        .map_or(0.0, f32::from),
```

szokeasaurusrex (approved):
looks good, see two comments before merge (both optional)

Issue #873, opened 2025-07-30
Report Response Context


lcian:
We should add the [Response context](https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#response-context) to the types, then add it to events generated by the actix and tower integrations.  
This will allow users to filter based on status code in `before_send`.

This came up with client errors previously being reported to Sentry. Now that we don't do that, this issue is less pressing, but it still would allow users to filter on `before_send`. This is needed in Rust SDK, since we don't have hints.

PR #874, opened 2025-07-30
feat(types): add Response context

lcian:
Adds the `Reponse` context, which can be attached to events.

The implementation is adapted directly from the one in [Relay](https://github.com/getsentry/relay/blob/1bdc463896fb6f1425fac10915fa49799ecc5d85/relay-event-schema/src/protocol/contexts/response.rs#L11), minus the `inferred_content_type`, which as documented [here](https://github.com/getsentry/relay/blob/1bdc463896fb6f1425fac10915fa49799ecc5d85/relay-event-normalization/src/normalize/contexts.rs#L326) can technically be sent by SDKs but will always be overridden by Relay.

Part of #873

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@
 
 - feat(core): add Response context ([#874](https://github.com/getsentry/sentry-rust/pull/874)) by @lcian
   - The `Response` context can now be attached to events, to include information about HTTP responses such as headers, cookies and status code.
+  - Example:
+    ```rust
+    let mut event = Event::new();
+    let response = ResponseContext {
+        cookies: Some(r#""csrftoken": "1234567""#.to_owned()),
+        headers: Some(headers_map),
+        status_code: Some(500),
+        body_size: Some(15),
+        data: Some("Invalid request"),
+    };
+    event
+        .contexts
+        .insert("response".to_owned(), response.into());
+    ```
 
 ## 0.42.0
 


sentry-types/src/protocol/v7.rs
```diff
@@ -1351,6 +1354,29 @@ pub struct OtelContext {
...
+    pub headers: Map<String, String>,
+    /// The HTTP response status code.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub status_code: Option<u64>,
+    /// The response body size in bytes.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub body_size: Option<u64>,
+    /// Response data in any format that makes sense.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub data: Option<Value>,
```
szokeasaurusrex, line 1377: Have you considered making the string types here `Cow<str>` instead of `String`? That way, folks could pass borrowed values instead of owned values, which could avoid clones in some cases

sentry-types/tests/test_protocol_v7.rs
```diff
@@ -1313,6 +1313,43 @@ mod test_contexts {
...
+        let event = v7::Event {
+            event_id: event_id(),
+            timestamp: event_time(),
+            contexts: {
+                let mut m = v7::Map::new();
+                m.insert(
+                    "response".into(),
+                    v7::ResponseContext {
+                        status_code: Some(400),
+                        cookies: Some("sessionId=abc123; Path=/; HttpOnly,authToken=xyz789; Secure; SameSite=Strict".into()),
```
szokeasaurusrex, line 1327: e.g. here, if we replaced the `String` type with `Cow<str>`, we could pass `Cow::Borrowed("...")`, avoiding a clone here.

sentry-types/src/protocol/v7.rs
szokeasaurusrex: Also, having all of the fields be `pub` means that adding even just one additional field would be a breaking change. It seems like many of the types in this file have all `pub` fields (and so this suggestion is non-blocking), but if we want to make the API more flexible in the future, we may want to start having private fields and adding constructors, getters, and setters to the types.

sentry-types/src/protocol/v7.rs
```diff
@@ -1501,9 +1527,10 @@ into_context!(Browser, BrowserContext);
 into_context!(Trace, TraceContext);
 into_context!(Gpu, GpuContext);
 into_context!(Otel, OtelContext);
+into_context!(Response, ResponseContext);
 
 const INFERABLE_CONTEXTS: &[&str] = &[
-    "device", "os", "runtime", "app", "browser", "trace", "gpu", "otel",
+    "device", "os", "runtime", "app", "browser", "trace", "gpu", "otel", "response",
 ];
```
szokeasaurusrex, line 1534: I am having trouble understanding what this code does (probably just due to lack of familiarity with the codebase), could you explain it?

lcian: `into_context` just generates the code to implement `From<ResponseContext>` for `Context`, which is the type you end up de/serializing. That just boxes the `ResponseContext` and constructs the appropriate variant of `Context` (in this case `Context::Response`).
As this is the same for all contexts, we generate it using a macro.

The `INFERABLE_CONTEXTS` is used to support a similar transformation but when deserializing.
e.g. if you get a JSON with `"response": { ... }`, you want to deserialize that to the `Context::Response` variant.
If it's not in `INFERABLE_CONTEXTS`, then we deserialize it into `Context::Unknown` which is practically an arbitrary map.
It was implemented like this for forwards compatibility. Right now I cannot think of any super good reasons to be forwards compatible, but for the time being I think we can keep using the existing patterns.

sentry-types/src/protocol/v7.rs
lcian: The reason for not using `Cow`s in this PR is also to adapt to the existing types, as mostly we don't use `Cow`s in them (and honestly I wonder why, I'll check tomorrow).
Same for the `pub` fields. Eventually we're going to want to have constructors, getters and setters for everything, otherwise it's too easy to make breaking changes. My idea would be to change all of the types at the same time though, instead of having a mix of visibility between them.

lcian:
> * How is this new type supposed to be used? Does not look like we are using it in the SDK, is it something that users are just supposed to know how to set? Or, does the `into_context!` stuff take care of setting it?

I'll make a PR on top of this to use this in the Actix Web integration.
We already capture most of this data using the [Request Interface](https://develop.sentry.dev/sdk/data-model/event-payloads/request/) but we're missing the status code.
This way, not only we will report the status code to Sentry, but most importantly a user will be able to filter an error in `before_send` based on the status code.

sentry-types/src/protocol/v7.rs
szokeasaurusrex: Yeah, I mean I would recommend doing this work in separate PRs, but it makes sense to release them at the same time (good usecase for a PR stack)

sentry-types/src/protocol/v7.rs
szokeasaurusrex: Got it. I think this might make more sense as an enum then, but I am not sure. Would be a future improvement though, definitely out of scope to swap this for an enum in this PR

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+## Features
+
+- feat(core): add Response context ([#874](https://github.com/getsentry/sentry-rust/pull/874)) by @lcian
+  - The `Response` context can now be attached to events, to include information about HTTP responses such as headers, cookies and status code.
```
szokeasaurusrex, line 8: I'd suggest adding an example of how to do this to the changelog. Otherwise, it is not super useful information for users

Issue #876, opened 2025-07-30
Don't capture client errors in Actix Web integration


lcian:
Due to [this code path](https://github.com/getsentry/sentry-rust/blob/4c4fc9d35317c2bfec48cc9c92eed961130a2b72/sentry-actix/src/lib.rs#L354), when polling the future returns an error, we capture it only based on `capture_server_errors`.<br><br>This can happen, for example, when a middleware processes the request after ours, and returns a `actix_web::error::Error`.<br>The problem is that the middleware could be, for instance, an authentication middleware, that returns 401 (a client error).<br>We would capture this error because we don't check if it's a client or server error.

PR #877, opened 2025-07-31
fix(actix): capture only server errors

lcian:
Before this change, if a middleware processing the request after the Sentry one returned an error, we would capture it unconditionally.
With this change, we capture it only if it's a server error.
(note that the behavior was correct for responses returned by services (i.e. request handler, not middleware), as at line 373 we check if it's a server error).

This is one of the possible ways to tackle this, and I think this is the one that makes most sense for the following reasons:
- Client errors could be spammy and don't usually provide much value to the user of Sentry.
- Capturing errors can be expensive when attaching stack traces.
- Better experience for new users that won't be surprised by this unintended behavior.

It must be said that there is a small possibility that someone misses this in the changelog who was actually relying on these errors to show up in Sentry.

Close https://github.com/getsentry/sentry-rust/issues/876

### DIFF
--- a/sentry-actix/src/lib.rs
+++ b/sentry-actix/src/lib.rs
@@ -351,7 +351,9 @@ where
             let mut res: Self::Response = match fut.await {
                 Ok(res) => res,
                 Err(e) => {
-                    if inner.capture_server_errors {
+                    // Errors returned by middleware, and possibly other lower level errors
+                    if inner.capture_server_errors && e.error_response().status().is_server_error()
+                    {
                         hub.capture_error(&e);
                     }
 
@@ -474,6 +476,7 @@ fn process_event(mut event: Event<'static>, request: &Request) -> Event<'static>
 mod tests {
     use std::io;
 
+    use actix_web::body::BoxBody;
     use actix_web::test::{call_service, init_service, TestRequest};
     use actix_web::{get, web, App, HttpRequest, HttpResponse};
     use futures::executor::block_on;
@@ -622,9 +625,9 @@ mod tests {
         }
     }
 
-    /// Ensures client errors (4xx) are not captured.
+    /// Ensures client errors (4xx) returned by service are not captured.
     #[actix_web::test]
-    async fn test_client_errors_discarded() {
+    async fn test_service_client_errors_discarded() {
         let events = sentry::test::with_captured_events(|| {
             block_on(async {
                 let service = HttpResponse::NotFound;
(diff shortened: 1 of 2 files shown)


Swatinem (approved):

sentry-actix/src/lib.rs
```diff
@@ -351,7 +351,9 @@ where
             let mut res: Self::Response = match fut.await {
                 Ok(res) => res,
                 Err(e) => {
-                    if inner.capture_server_errors {
+                    // Errors returned by middleware, and possibly other lower level errors
+                    if inner.capture_server_errors && e.error_response().status().is_server_error()
```
szokeasaurusrex, line 355: [future improvement] Does the Rust SDK offer a way to configure the exact status codes which are sent to Sentry? We have such a feature in several of the Python SDK's server integrations; maybe it would be worth it to consider adding this feature to the Rust SDK, as well.

sentry-actix/src/lib.rs
```diff
@@ -622,9 +625,9 @@ mod tests {
         }
     }
 
-    /// Ensures client errors (4xx) are not captured.
+    /// Ensures client errors (4xx) returned by service are not captured.
```
szokeasaurusrex, line 628: What is a "service" in this context? Is it a middleware?

[edit] actually if I understand correctly, the service is the endpoint itself, not the middleware. Unless this is common terminology in actix-web, I would clarify the wording in the comment

sentry-actix/src/lib.rs
```diff
@@ -645,6 +648,66 @@ mod tests {
...
+                        })
+                        .wrap(Sentry::builder().with_hub(Hub::current()).finish())
+                        .service(web::resource("/test").to(hello_world)),
+                )
+                .await;
+
+                let req = TestRequest::get().uri("/test").to_request();
+                let res = app.call(req).await;
+                assert!(res.is_err());
+                assert!(res.unwrap_err().error_response().status().is_client_error());
```
szokeasaurusrex, line 674: [nit] I would check the exact status code, not just that it is a client error here

sentry-actix/src/lib.rs
```diff
@@ -645,6 +648,66 @@ mod tests {
...
+                        })
+                        .wrap(Sentry::builder().with_hub(Hub::current()).finish())
+                        .service(web::resource("/test").to(hello_world)),
+                )
+                .await;
+
+                let req = TestRequest::get().uri("/test").to_request();
+                let res = app.call(req).await;
+                assert!(res.is_err());
+                assert!(res.unwrap_err().error_response().status().is_server_error());
```
szokeasaurusrex, line 704: [nit] also would check exact status code here

szokeasaurusrex (approved):

szokeasaurusrex:
looks good, couple of small (mostly optional) suggestions

sentry-actix/src/lib.rs
lcian: Ah, I didn't know we had it, then yes it could make sense to offer it.

sentry-actix/src/lib.rs
lcian: Yeah it refers to the handler itself for one or more routes + possibly other components such as configuration.
This is a standard term in Actix so I think it's fine.

PR #886, opened 2025-08-25
Update README.md

PJUllrich:
The Readme has a typo AFAIK. The `logs` feature is called `log` but the code says it should be `logs`. I updated it to `logs` now :)

### DIFF
--- a/sentry/README.md
+++ b/sentry/README.md
@@ -75,7 +75,7 @@ extra setup to function properly.
 | `anyhow`          |         | 🔌             |            |                                                                                          |
 | `test`            |         |                |            |                                                                                          |
 | `debug-images`    | ✅      | 🔌             |            |                                                                                          |
-| `log`             |         | 🔌             |            | Requires extra setup; See [`sentry-log`]'s documentation.                                |
+| `logs`            |         | 🔌             |            | Requires extra setup; See [`sentry-log`]'s documentation.                                |
 | `slog`            |         | 🔌             |            | Requires extra setup; See [`sentry-slog`]'s documentation.                               |
 | `reqwest`         | ✅      |                |            |                                                                                          |
 | `native-tls`      | ✅      |                |            | `reqwest` must be enabled.                                                               |
@@ -107,7 +107,7 @@ extra setup to function properly.
 - `test`: Enables testing support.
 
 ### Logging
-- `log`: Enables support for the `log` crate.
+- `logs`: Enables support for the `log` crate.
 - `slog`: Enables support for the `slog` crate.
 
 ### Transports


PJUllrich:
Wait, there's a `log` **and** a `logs` feature?

https://docs.sentry.io/platforms/rust/logs/#log-integration

PJUllrich:
Thank you for the context @lcian. I ended up manually logging messages to Sentry because I'm building a Tauri app and they already have a log plugin and require me to provide a custom `fern::Dispatch` logger: https://docs.rs/tauri-plugin-log/latest/tauri_plugin_log/enum.TargetKind.html

But regarding `log` vs `logs`: I'm no Rust expert Rust at all, but as someone coming from the outside I'd find it easier to only use e.g. the `log` dependency with e.g. `env_logger` instead of using `sentry::logger_info`. So, you could  consider removing the `log` feature and move entirely towards `logs` with either `tracing` or `log + env_logger`. For me that feels like the most "correct" way of tapping into the Rust logs.

I'll close this for now since it's not a typo anymore :)

lcian:
Cool! I haven't used `fern` yet but I assume the setup would require setting up a `fern::Dispatch` that calls the `sentry::logger_` macros, right?
If that's the case, I might as well make the `sentry::logger_log` "internal" macro (which takes the log level as first param) public as well, as it seems it could reduce code duplication for use cases like this.

PJUllrich:
I'm okay with how it is right now :) I'm able to log my errors to disk, stdout, and Sentry at the same time with this Tauri plugin setup:

```rust
tauri::Builder::default()
.plugin(
            tauri_plugin_log::Builder::new()
                .rotation_strategy(tauri_plugin_log::RotationStrategy::KeepAll)
                // Max log file size: 5 MB
                .max_file_size(5_000_000)
                .format(|out, message, record| {
                    out.finish(format_args!(
                        "[{}][{}] {}",
                        Local::now().format("%Y-%m-%d %H:%M:%S"),
                        record.level(),
                        message
                    ))
                })
                // Add a custom Log Target for sending Errors to Sentry.
                .target(tauri_plugin_log::Target::new(
                    tauri_plugin_log::TargetKind::Dispatch(
                        fern::Dispatch::new().level(log::LevelFilter::Error).chain(
                            fern::Output::call(|record| {
                                let message = format!("{}", record.args());
                                sentry::capture_message(&message, sentry::Level::Error);
                            }),
                        ),
                    ),
                ))
                .level(if cfg!(any(dev, test)) {
                    log::LevelFilter::Debug
                } else {
                    log::LevelFilter::Info
                })
                .build(),
        )
```

Issue #880, opened 2025-08-04
tracing::span "message" field should NOT be empty during span creation


georgeboot:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

0.42.0

### Steps to reproduce

This issue relates to #879

```rust
#[tracing::instrument(name = "http.client", fields(message = tracing::field::Empty, http.response.status_code = tracing::field::Empty))]
async fn do_expensive_operation(webhook_request: Request) {
    let url = webhook_request.url();
    let http_endpoint = &url[..url::Position::AfterPath];
    let method = request.method().as_str();
    let message = format!("{method} {http_endpoint}");
    
    let current_span = Span::current();
    
    // set the message
    current_span.record("message", &message);
    
    let response = my_http_client.execute(request).await;
   
    // record status_code
    current_span.record("http.response.status_code", response.status().as_u16());
}
```

### Expected result

A sentry span gets created with these fields:
```json
{
  "op": "http.client",
  "description": "GET https://google.com",
  "data": {
    "http.response.status_code": 200
  }
}
```

### Actual result

A span was created with these fields:
```json
{
  "op": "http.client",
  "description": "my_module::path::http.client", // this is unexpected!
  "data": {
    "message": "GET https://google.com",
    "http.response.status_code": 200
  }
}
```

Issue #879, opened 2025-08-04
Sentry span.description requires setting "message" field


georgeboot:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

0.42.0

### Steps to reproduce

As a Rust developer, I expect to be able to do something like:
```rust
let span = tracing::info_span!(
    "http.client",
    description = "GET https://google.com",
);
my_future.instrument(span).await;
```

### Expected result

A span should created in sentry with these fields:
```json
{
  "op": "http.client",
  "description": "GET https://google.com"
}
```

### Actual result

A span is created in sentry with these fields:
```json
{
  "op": "http.client",
  "description": "my_module::path::http.client", // this is unexpected!
  "data": {
    "description": "GET https://google.com"
  }
}
```

PR #887, opened 2025-08-28
ref(tracing): rework `tracing` to Sentry span name/op conversion

lcian:
### Description

This changes how the `tracing` integration works with regards to span names and operations.
(In the analysis below, I assume the spans are all created with the `tracing::instrument` macro)

### Previous behavior
- Sentry span name = `<module>::<function>`
   - unless a `message` field is set, in which case: Sentry span name = tracing span `message`
   - this `message` field is arbitrary, not documented, and doesn't align with our conventions
   - note that it's different from the tracing span **name**, this requires an attribute called `message`
- Sentry op = function name

Screenshot 2025-08-28 at 16 16 01

### Behavior after this PR
- Sentry span name = `tracing` span name (= function name, when using the `tracing::instrument` macro)
- Sentry op = `default`, as we have no way to infer an `op` that makes sense - it will be hidden in the UI
- Users can set 3 additional fields that have special meaning for our `tracing` Subscriber and override values on the corresponding Sentry span:
   - `sentry.name`
   - `sentry.op`
   - `sentry.trace`: while not related to this issue, I noticed that there is confusion on how one would achieve distributed tracing. This makes it possible without having to use the Sentry tracing API and understand the interplay between Sentry and the `tracing` crate.

Screenshot 2025-08-28 at 16 15 51

### Advantages of the change
- The new conversion is more aligned with Sentry conventions, documented, and gives additional capabilities to users
- It's also similar to the approach taken by `tracing-opentelemetry`, which offers a `otel.name` special property

### Drawbacks of the change
- We break existing customers that might be using span metrics/dashboards/queries based on the old span name - they will have to change them
- If you have 2 functions with the same name in different modules, they will now have the same span name. We could mitigate this by writing the info from `span.metadata.target` into span attributes, like we do for logs to surface function and line

### Closes
Close https://github.com/getsentry/sentry-rust/issues/880
Close https://github.com/getsentry/sentry-rust/issues/879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@
         // ...
     }
     ```
+  - Additional attributes are sent along with each span by default:
+    - `sentry.tracing.target`: corresponds to the `tracing` span's `metadata.target()`
+    - `code.module.name`, `code.file.path`, `code.line.number`
 
 - feat(core): add Response context ([#874](https://github.com/getsentry/sentry-rust/pull/874)) by @lcian
   - The `Response` context can now be attached to events, to include information about HTTP responses such as headers, cookies and status code.


Swatinem (approved):
nice, this looks a lot cleaner indeed.

I think its fair to use just the function name, from the users perspective, I think that makes sense. Although I also might want to see the fully qualified name sometimes, in particular for method names.
its a bit unfortunate that `tracing` does not differentiate between an impl method and a free function here.
Ideally, it should qualify methods with the impl (such as `Vec::extend_from_slice`).
I think not qualifying anything could lead to confusion for example if you have `str::from_utf8` vs `String::from_utf8`, which are two different methods.

But okay, I guess that is a minor annoyance.

lcian:
Thanks @Swatinem, that's a fair point.
We could still report the fully qualified name as the op, so the information in the UI would be the same as before but swapped. Maybe that's the way to go, as `default` doesn't bring any value to the user anyway, might as well "abuse" the field to report the full name.

lcian:
> One thing to try would be whether it actually is the "fully qualified" name, because just concatenating the `module` to the `name`, does it also include the type name for impl methods?

It seems it's just the module path, which doesn't include any information about trait impls.
Seems something it could make sense to propose upstream though.

lcian:
Found some discussions around this topic
https://github.com/tokio-rs/tracing/issues/2116
https://github.com/tokio-rs/tracing/issues/1939
https://github.com/tokio-rs/tracing/issues/1123

sentry-core/src/performance.rs
```diff
@@ -781,6 +797,20 @@ impl Transaction {
         inner.context.status = Some(status);
     }
 
+    /// Set the operation of the Transaction.
+    pub fn set_op(&self, op: &str) {
+        let mut inner = self.inner.lock().unwrap();
```
szokeasaurusrex, line 802: Perhaps of scope for this PR, as it seems we are unwrapping mutexes elsewhere in this file, but this `.unwrap()` seems quite a bit risky to me, as we would be causing a panic in the user's application code. I think it would be safer to have this function return a result.

sentry-tracing/src/layer.rs
```diff
@@ -410,7 +438,14 @@ where
...
 /// Extracts the message and attributes from a span
-fn extract_span_data(attrs: &span::Attributes) -> (Option<String>, BTreeMap<&'static str, Value>) {
+fn extract_span_data(
+    attrs: &span::Attributes,
+) -> (
+    BTreeMap<&'static str, Value>,
+    Option<String>,
+    Option<String>,
+    Option<String>,
+) {
```
szokeasaurusrex, line 473: Would consider creating a struct, with named fields, for this. Otherwise, it is extremely difficult to understand what each of these fields represents.

sentry-tracing/src/layer.rs
```diff
@@ -397,6 +399,32 @@ where
...
+        let sentry_op = data
+            .json_values
+            .remove(SENTRY_OP_FIELD)
+            .and_then(|v| match v {
+                Value::String(s) => Some(s),
+                _ => None,
+            });
+
+        // `sentry.trace` cannot be applied retroactively
+        data.json_values.remove(SENTRY_TRACE_FIELD);
```
szokeasaurusrex, line 427: Just wondering, in what case do we hit this code path, rather than the `on_new_span` path?

sentry-tracing/src/lib.rs
```diff
@@ -198,6 +198,42 @@
...
+//!
+//! By default, the `op` of the span sent to Sentry is `default`.
+//!
+//! ## Special Span Fields
+//!
+//! Some fields on spans are treated specially by the Sentry tracing integration:
+//! - `sentry.name`: overrides the span name sent to Sentry.
+//!   This is useful to customize the span name when using `#[tracing::instrument]`, or to update
+//!   it retroactively (using `span.record`) after the span has been created, as `tracing` doesn't allow doing it.
+//! - `sentry.op`: overrides the span `op` sent to Sentry.
```
szokeasaurusrex: > as `tracing` doesn't allow doing it

This part of the doc comment is confusing and vague; I am not sure what the word "it" is referring to here.

Please replace "it" with what you actually meant to refer to. In general, I would recommend avoiding the word "it" (also "this") in technical documentation, as they are prone to being vague and misinterpreted 🙏

sentry-tracing/src/lib.rs
```diff
@@ -198,6 +198,42 @@
...
+//! - `sentry.trace`: in Sentry, the `sentry-trace` header is sent with HTTP requests to achieve distributed tracing.
+//!   If the value of this field is set to the value of a valid `sentry-trace` header, which
+//!   frontend SDKs send automatically with outgoing requests, then the SDK will continue the trace using the given distributed tracing information.
+//!   This is useful to achieve distributed tracing at service boundaries by using only the
+//!   `tracing` API.
+//!   Note that this will only be effective on span creation (cannot be applied retroactively) and
```
szokeasaurusrex:
```suggestion
//!   Note that `sentry.trace` will only be effective on span creation (cannot be applied retroactively) and
```

sentry-tracing/src/lib.rs
```diff
@@ -198,6 +198,42 @@
...
+//! ## Special Span Fields
+//!
+//! Some fields on spans are treated specially by the Sentry tracing integration:
+//! - `sentry.name`: overrides the span name sent to Sentry.
+//!   This is useful to customize the span name when using `#[tracing::instrument]`, or to update
+//!   it retroactively (using `span.record`) after the span has been created, as `tracing` doesn't allow doing it.
+//! - `sentry.op`: overrides the span `op` sent to Sentry.
+//! - `sentry.trace`: in Sentry, the `sentry-trace` header is sent with HTTP requests to achieve distributed tracing.
+//!   If the value of this field is set to the value of a valid `sentry-trace` header, which
+//!   frontend SDKs send automatically with outgoing requests, then the SDK will continue the trace using the given distributed tracing information.
```
szokeasaurusrex: Technically not only frontend SDKs, but all SDKs should set these headers on outgoing requests.

sentry-tracing/tests/name_op_updates.rs
```diff
@@ -0,0 +1,45 @@
...
+        assert_eq!(transaction.name.as_deref().unwrap(), "updated name");
+        let ctx = transaction.contexts.get("trace");
+        match ctx {
+            Some(sentry::protocol::Context::Trace(trace_ctx)) => {
+                assert_eq!(trace_ctx.op, Some("updated op".to_owned()))
+            }
+            _ => panic!("expected trace context"),
+        }
+    }
+}
```
szokeasaurusrex, line 45: I would also consider adding a test to valiate the new default behavior, when the fields are not specified.

CHANGELOG.md
```diff
@@ -4,6 +4,28 @@
...
+    ```rust
+    #[tracing::instrument(skip_all, fields(
+        sentry.op = "http.server",
+        sentry.name = "GET /payments",
+        sentry.trace = headers.get("sentry-trace").unwrap_or(&"".to_owned()),
+    ))]
+    async fn handle_request(headers: std::collections::HashMap<String, String>) {
+        // ...
+    }
+    ```
```
szokeasaurusrex: As only some of these changes are breaking (unless I am missing something), I would indicate more clearly which specific changes are breaking and which ones are new features, which do not require migration.

sentry-tracing/src/layer.rs
lcian: Here we're in `on_record`, this is hit when you do something like
```
    tracing::Span::current().record("sentry.name", "updated name");
```

szokeasaurusrex: So this is something you would do if there is already a span running, and you want to update the name of it?

lcian: Yep

sentry-tracing/src/lib.rs
lcian: It seems pretty clear to me, `tracing` doesn't allow what is said in the previous part of the phrase, i.e. `customize the span name [...], or [to] update it retroactively [...] after the span has been created`
I don't see how to make this clearer without excessive repetition

szokeasaurusrex: Ok, I see what you mean, but it was difficult for me to get that meaning even from re-reading now.

I would honestly just suggest deleting the whole "as `tracing` doesn't allow doing it" part, because it is potentially confusing. The important thing for folks to take away is that they can retroactively update the span.

sentry-tracing/tests/name_op_updates.rs
lcian: op was already covered by the smoke test below, added an assertion on the name too

sentry-core/src/performance.rs
lcian: Fair point

sentry-tracing/src/layer.rs
lcian: It's internal, will leave it like this for now

szokeasaurusrex: For sure, feel free to implement in a separate PR. But I think it is still worth making named fields, even if the function remains internal. Otherwise, anyone else who works on the Rust SDK will struggle to understand what these fields each do.

szokeasaurusrex: At a minimum, adding something to the doc comment to briefly explain each value would help.

sentry-tracing/src/layer.rs
```diff
@@ -435,6 +437,22 @@ where
     }
 }
 
+fn set_default_attributes(span: &mut TransactionOrSpan, metadata: &'static Metadata<'static>) {
```
szokeasaurusrex: The only thing I would question here is whether we need the `'static` restriction, as it makes the API less flexible. But, if it is necessary for these changes, it is fine to keep – just wanted to suggest you make sure that it is actually needed, if you have not already done so.

PR #899, opened 2025-09-17
ref(tracing): use standard code attributes

lcian:
### Description

Use the `code` attributes standardized in [OTEL](https://opentelemetry.io/docs/specs/semconv/registry/attributes/code/) to surface module, file and line info when creating logs out of `tracing` events, instead of using "custom" `tracing.<x>` ones.


lcian:
Even though logs recently launched and I don't expect people to already have used these attributes anywhere in the product, this is technically a breaking change, so I'm documenting it as such in the changelog.

szokeasaurusrex (approved):
lgtm

PR #902, opened 2025-09-17
meta(vscode): Run `rust-analyzer` with all features

szokeasaurusrex:
So that `rust-analyzer` also analyzes feature-flag-guarded portions of code when using VSCode/Cursor.

#skip-changelog

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,5 @@
+[profile.ci]
+fail-fast = false
+
+[profile.ci.junit]
+path = "junit.xml"
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,14 @@
+### Description
+<!-- What changed and why? -->
+
+#### Issues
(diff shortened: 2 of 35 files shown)


szokeasaurusrex (commented):
Feel free to close if you don't want this – I noticed the potential improvement while reviewing #899

lcian:
Nice thanks @szokeasaurusrex, should we enable all features directly? Any downside of doing so?

lcian:
I guess `rust-analyzer` will be a bit slower, but I think it makes sense to enable all features

.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
+    "rust-analyzer.cargo.features": [
+        "logs"
+    ]
```
lcian:
```suggestion
    "rust-analyzer.cargo.features": "all"
```
this should do

Issue #895, opened 2025-09-11
Fix tracing span attributes in logs


lcian:
### Description

When `enable_span_attributes` is used, the span attributes merge with the log attribute, which is wrong.

<img width="338" height="56" alt="Image" src="https://github.com/user-attachments/assets/657e5cf6-d58c-411b-bf06-25cd8c50d28a" />

Fix it.

lcian:
2 points:
- In `log_from_event` we need to use `extract_event_data`, not `extract_event_data_with_context`
- The changes in https://github.com/getsentry/sentry-rust/pull/887 should also be revisited because if you use `enable_span_attributes` you would attach all the default attributes every time you use `extract_event_data_with_context` which doesn't make too much sense. We need to exclude the default attributes

lcian:
Actually it's fine to propagate the span attributes as log attributes (i.e. use `extract_event_data_with_context`) but we should ignore the default ones.
Also spaces are not supported in attribute names so we should make sure the name is normalized either here or in ingestion.

PR #904, opened 2025-09-19
fix(tracing): skip default span attributes when propagating to event

lcian:
### Description

In https://github.com/getsentry/sentry-rust/pull/887 we added some attributes that are attached by default on all spans to surface more metadata.
We have an option on the `tracing` layer called `enable_span_attributes` that, when emitting an error/log out of a `tracing` event, attaches all the attributes of each span in the currently active span hierarchy to the error/log.

This results in a lot of noise, as you will see those attributes for every span that is currently active, which doesn't provide much value when looking at an error/log.
See this example with a log:
Screenshot 2025-09-22 at 12 56 37

With this change, we'll only propagate user-provided attributes to the error/log, skipping those we set by default on the spans.
When looking at the trace in Sentry, you'll still be able to see those on the corresponding span, as it provides valuable information.

#### Issues

Close https://github.com/getsentry/sentry-rust/issues/895
Close RUST-104

#skip-changelog because this fixes a bug for something that wasn't released

### DIFF
--- a/sentry-tracing/src/converters.rs
+++ b/sentry-tracing/src/converters.rs
@@ -101,6 +101,9 @@ where
                 match &span_data.sentry_span {
                     TransactionOrSpan::Span(span) => {
                         for (key, value) in span.data().iter() {
+                            if is_sentry_span_attribute(key) {
+                                continue;
+                            }
                             if key != "message" {
                                 let key = format!("{name}:{key}");
                                 visitor.json_values.insert(key, value.clone());
@@ -109,6 +112,9 @@ where
                     }
                     TransactionOrSpan::Transaction(transaction) => {
                         for (key, value) in transaction.data().iter() {
+                            if is_sentry_span_attribute(key) {
+                                continue;
+                            }
                             if key != "message" {
                                 let key = format!("{name}:{key}");
                                 visitor.json_values.insert(key, value.clone());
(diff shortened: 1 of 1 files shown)


sentry-tracing/src/converters.rs
```diff
@@ -101,6 +101,9 @@ where
                 match &span_data.sentry_span {
                     TransactionOrSpan::Span(span) => {
                         for (key, value) in span.data().iter() {
+                            if is_sentry_span_attribute(key) {
```
szokeasaurusrex, line 104: Is it possible that a user would set a Sentry span attribute manually, and would we want to keep that attribute in that case?

szokeasaurusrex: [`l`](https://develop.sentry.dev/engineering-practices/code-review/#logaf-scale): also I would suggest using a `.filter` on the iterator, rather than the `if` block with a `continue`. Imo iterator chains tend to be more readable than branching via `if` statements

lcian: Unlikely, but in theory it's possible that you could want to override them yeah.

lcian:
This behavior (`enable_span_attributes`) is unique to the `tracing` layer, it doesn't apply to any other integration or SDK, so there's no standard/spec for this.
If we were to propose it, I don't think it would be implemented.
But in this SDK it's already here because someone found it useful, and honestly I can see the value in it.
So I would just keep it there and with this patch we ensure those fields don't become too noisy for no benefit.

Issue #908, opened 2025-09-23
Support multiple mappings in `log` integration


lcian:
### Description

As we did for the `tracing` integration, it should be easy to map a single `log` log into multiple Sentry items.

Issue #909, opened 2025-09-23
Enable logs by default in `log` integration


lcian:
### Description

If the user already enabled the `logs` feature flag, add a logs mapping in the default event mapper.
Requires https://github.com/getsentry/sentry-rust/issues/908

PR #914, opened 2025-09-30
feat(log): support combined LogFilters and RecordMappings

lcian:
### Description

Adds support for combining multiple `LogFilters` when using our integration for the `log` crate.
This way, it's possible to map a `log` record to multiple items in Sentry, for example to both a `log` and an `event`.

This is required for https://github.com/getsentry/sentry-rust/issues/909.
There, we want to enable logs for records at or above INFO by default, in order to remove the double opt-in, while still sending breadcrumbs for them at the same time.

The changes are analogous with the ones we did for the `tracing` integration here: https://github.com/getsentry/sentry-rust/pull/847

#### Issues

Close https://github.com/getsentry/sentry-rust/issues/908
Close RUST-111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - feat(log): support combined LogFilters and RecordMappings ([#914](https://github.com/getsentry/sentry-rust/pull/914)) by @lcian
   - `sentry::integrations::log::LogFilter` has been changed to a `bitflags` struct.
   - It's now possible to map a `log` record to multiple items in Sentry by combining multiple log filters in the filter, e.g. `log::Level::ERROR => LogFilter::Event | LogFilter::Log`.
-  - It's also possible to use `sentry::integrations::log::RecordMapping::Combined` to map a `log` record to multiple items in Sentry.
+  - If using a custom `mapper` instead, it's possible to return a `Vec<sentry::integrations::log::RecordMapping>` to map a `log` record to multiple items in Sentry. 
 
 ## 0.43.0
 


sentry-log/src/logger.rs
```diff
@@ -1,24 +1,27 @@
...
+        /// Ignore the [`Record`].
+        const Ignore = 0b0000;
+        /// Create a [`Breadcrumb`] from this [`Record`].
+        const Breadcrumb = 0b0001;
+        /// Create a message [`Event`] from this [`Record`].
+        const Event = 0b0010;
+        /// Create an exception [`Event`] from this [`Record`].
+        const Exception = 0b0100;
+        /// Create a [`sentry_core::protocol::Log`] from this [`Record`].
+        const Log = 0b1000;
```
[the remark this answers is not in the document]
lcian: I guess we can gate it, doesn't really matter.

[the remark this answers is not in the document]
szokeasaurusrex: Gating makes sense I think

sentry-log/src/logger.rs
```diff
@@ -1,24 +1,28 @@
...
-    /// Create a [`Breadcrumb`] from this [`Record`].
-    Breadcrumb,
-    /// Create a message [`Event`] from this [`Record`].
-    Event,
-    /// Create an exception [`Event`] from this [`Record`].
-    Exception,
-    /// Create a [`sentry_core::protocol::Log`] from this [`Record`].
-    #[cfg(feature = "logs")]
-    Log,
+bitflags! {
```
szokeasaurusrex, line 10: Cool, TIL about `bitflags!`

It does seem a _bit_ low-level, but probably also makes sense here – pun intended 😅

sentry-log/src/logger.rs
```diff
@@ -1,24 +1,28 @@
...
+    /// The action that Sentry should perform for a [`log::Metadata`].
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub struct LogFilter: u32 {
```
szokeasaurusrex, line 13: Why use `u32` here? I guess `u8` would be sufficient, unless we plan to add many more filter types in the future. 

Does not really matter, I am just wondering if there is a reason.

sentry-log/src/logger.rs
```diff
@@ -132,30 +159,50 @@ impl<L: log::Log> SentryLogger<L> {
 
 impl<L: log::Log> log::Log for SentryLogger<L> {
     fn enabled(&self, metadata: &log::Metadata<'_>) -> bool {
-        self.dest.enabled(metadata) || !matches!((self.filter)(metadata), LogFilter::Ignore)
+        self.dest.enabled(metadata) || !((self.filter)(metadata) == LogFilter::Ignore)
```
szokeasaurusrex, line 147: Why not use `contains` instead of `==` here?

sentry-log/src/logger.rs
```diff
@@ -34,6 +38,29 @@ pub enum RecordMapping {
...
     Log(sentry_core::protocol::Log),
+    /// Captures multiple items to Sentry.
+    /// Nesting multiple `RecordMapping::Combined` is not supported and will cause the mappings to
+    /// be ignored.
+    Combined(CombinedRecordMapping),
+}
+
+/// A list of record mappings.
+#[derive(Debug)]
+pub struct CombinedRecordMapping(Vec<RecordMapping>);
```
szokeasaurusrex: m: I think this whole `CombinedRecordMapping` stuff is adding unnecessary complexity. In my opinion, it would be simpler just to use a `Vec<RecordMapping>` directly. I also question the need to have a `Combined` item in the `RecordMapping` enum; we can instead just use the `Vec` wherever we might need multiple `RecordMapping`s (as `Vec` can also contain a single item).

sentry-log/Cargo.toml
```diff
@@ -19,6 +19,7 @@ logs = ["sentry-core/logs"]
 [dependencies]
 sentry-core = { version = "0.43.0", path = "../sentry-core" }
 log = { version = "0.4.8", features = ["std", "kv"] }
+bitflags = "2.0.0"
```
szokeasaurusrex: Latest version is `2.9.4`, why not specify that directly?

szokeasaurusrex (commented):
Looks good in principle, but I have a suggestion for simplifying the code

sentry-log/src/logger.rs
lcian: In theory (even though it doesn't make too much sense) one could have `LogFilter::Combined` with `Ignore` and something else inside.

sentry-log/src/logger.rs
lcian: Yeah no reason really, u8 would most definitely work here as well, as you mentioned we most likely won't ever saturate a whole u8 even.

sentry-log/src/logger.rs
lcian: I think there was a reason I did it like this in the first place but right now I don't remember, let me look into it a bit more and see if we can simplify it.

lcian: You were right @szokeasaurusrex :), please take a look at the new version.

sentry-log/Cargo.toml
lcian: Specified the latest here and in `sentry-tracing` where it was also used.

Swatinem (approved):
One weird thing might be that with `bitflags`, you can combine `Ignore` with any of the others just fine, and the compiler won’t complain, even though it does not make sense.
But thats a minor papercut, not worth wasting time over.

PR #915, opened 2025-09-30
ref(log): send logs by default when logs feature flag is enabled

lcian:
### Description

Removes the double opt-in for the `log` crate integration, as we'll now map to both an `exception`/`breadcrumb` and a `log` when the `logs` feature flag is active and the level is at or above INFO.

If the user has set `enable_logs: false` in their client options, the log will be created but dropped when it reaches the `client`.
This creates the log uselessly to later discard it. Still, in 99% of cases, if I use the `logs` feature flag, it means I want to get logs, so most probably `enable_logs == true` and the log is actually used.

This change is analogous to the one we already did for the `tracing` integration.

#### Issues

Closes https://github.com/getsentry/sentry-rust/issues/909
Closes RUST-112


szokeasaurusrex (approved):
Looks good! Long term, I would consider completely removing the `enable_logs` flag, and have setting/unsetting the feature flag be the only way to enable/disable logs

szokeasaurusrex:
@lcian Ah I was more referring to that we maybe should consider completely removing the flag, so that you cannot disable logs if the logs feature flag is turned on

PR #937, opened 2025-11-12
feat: put sentry_debug! logs behind feature flag

N3M0dropserver:
### Description
feature/chore to allow for enabling and disabling [sentry] debug logs. I was getting my container and other log systems filled with logs like this:
```sh
[sentry] [LogsBatcher] Flushing 31 logs
[sentry] Get response: `{"id":"f42674c3591b41dfa855d216760a9ea8"}`
```
Totally get having the debug logs there and I understand their use but I wanted the ability to disable them.

Perhaps another approach could be to make them traces at the trace/debug level if tracing flag is enabled that way people could opt in with level filters?

For now I've just feature flagged the macro to a no-op when used if the debug_logs flag is not enabled.

Happy for input or thoughts but I would personally love this.

#### Issues
N/A

### DIFF
--- a/sentry-core/Cargo.toml
+++ b/sentry-core/Cargo.toml
@@ -20,7 +20,7 @@ name = "scope_benchmark"
 harness = false
 
 [features]
-default = ["debug_logs"]
+default = []
 client = ["rand"]
 test = ["client", "release-health"]
 debug_logs = []
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -282,7 +282,6 @@ impl Client {
             event = match integration.process_event(event, &self.options) {
                 Some(event) => event,
                 None => {
-                    #[cfg(feature = "debug_logs")]
                     sentry_debug!("integration dropped event {:?}", id);
                     return None;
                 }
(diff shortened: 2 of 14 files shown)


lcian:
This is basically how it worked before I changed this, when we had the `debug-logs` feature flag.

I really like the fact that this allows us to have a no-op implementation in case you don't want to log, and I think this is something I overlooked when removing the feature flag.
This way of implementing it would allow us to be much more verbose with logging. Admittedly there've been cases where I wanted to add debug logs for some part of the code but decided not to do it because it was in a "hot path" - without the approach proposed here, you would always need to grab the current Hub -> client -> check if `debug` is true regardless.

This approach however has 2 downsides:
- more confusing to users: we would have 3 very similar feature flags: `log` (the `log` crate integration), `logs` (enables capturing logs through the batcher logger which runs on its own thread), `debug-logs` (this feature)
  - if we decide to do this, we need to document it, and decide what to do with the existing `debug` field of `ClientOptions`. Either we remove it in favor of this flag or we make it conditional on this flag.
- if you want to emit debug logs from an integration crate (which ideally you would, considering it would also be zero-cost in general) then you would also need to define the `debug-logs` feature there and enable it if it's enabled in `sentry-core` (and vice versa maybe)

PR #945, opened 2025-12-04
feat: Add data option in tracing integration to force sampling of transaction

Ten0:
### Description
Similarly to `sentry.trace`, this allows overriding whether the transaction is sampled while still going through the `tracing` integration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- ref(tracing): Add data option in tracing integration to force sampling of transaction ([#945](https://github.com/getsentry/sentry-rust/pull/945)) by @Ten0
+  - Additional special fields have been added that allow overriding certain data on the Sentry span:
+    - `sentry.sample`: If the span initiates a transaction, this can be used to force whether the transaction will be sampled.
+
 ## 0.46.0
 
 ### Breaking changes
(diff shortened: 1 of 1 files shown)


lcian:
Hi @Ten0, thank you for the PR. 
Do you need tail-based sampling based on some attributes, or would using a `span_filter` work for you? https://github.com/getsentry/sentry-rust/blob/196501d8fa067bdf35b7c95261f6b7fe7bb84bb8/sentry-tracing/src/layer.rs#L147

Generally, all of our other SDKs offer `before_send_span`/`before_send_transaction` hooks to centrally manage span filtering and mutation. So we should consider whether we want to introduce this, the hooks, or both.
There's also `traces_sampler` which is already available in this SDK, albeit less powerful (very similar to `span_filter` but it acts on whole traces as opposed to spans).

Ten0:
Hello, yes I'm looking for specifying sampling not just based on the span itself, specifically I'm using another header than sentry trace to specify just "please sample this", without specifying a trace ID. That's useful e.g. when request is initialized via Insomnia.
I guess traces_sampler might work as well if I could efficiently read an attribute that I would dedicate to this...

lcian:
@Ten0 thanks for the context.
I think `TracesSampler` wouldn't work, because you only have access to some limited information about the trace there like name and op but not attributes.

`before_send_span/transaction` however would work because that would happen, clearly, before sending, so all attributes would be available in that callback.
If we implemented `before_send_span`, you could even recreate the feature you're PRing here on your own, so that could be another point in favor of implementing one of these hooks.

PR #957, opened 2026-01-14
fix!(core): Make HubSwitchGuard !Send to prevent thread corruption

### LINKED ISSUE #943: HubSwitchGuard should not be Send
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

0.46.0

### Steps to reproduce

1. `let sg = HubSwitchGuard::new(Hub::current())`
2. `std::thread::spawn(move|| { let _ = sg; }`

### Expected result

Compilation error

### Actual result

New thread gets its hub replaced by that of the parent thread, and parent thread never recovers their own hub.

Because `HubSwitchGuard` is about managing the thread-local that represents the current hub for this thread, it really doesn't make sense to move it to another thread, and as such that should be prevented. (`HubSwitchGuard` should be `!Send`)

This can be achieved with e.g. a field like this `_make_type_not_send: std::marker::PhantomData<std::sync::MutexGuard<'static, ()>>,`

### LINKED ISSUE #946: Entering the same span several times causes esoteric behavior
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

0.46.0

### Steps to reproduce

Create a tracing span.
Enter it in a thread/task n°1 (to get the hub/subspans).
Enter it in another thread/task n°2 (to get the same hub/subspans).
Exit it in n°2
Exit it in n°1

### Expected result

They don't conflict with one another

### Actual result

Due to:
https://github.com/getsentry/sentry-rust/blob/196501d8fa067bdf35b7c95261f6b7fe7bb84bb8/sentry-tracing/src/layer.rs#L355

entering the span on thread 2 will lead to the hub switch guard of thread 1 being dropped in thread 2, which means that thread 2 will actually not enter the hub, and then when exiting the span in thread 1 it won't actually exit the hub (because hub_switch_guard will have been left at None by thread 2).

szokeasaurusrex:
## Description

This PR does three important things, which are all interdependent, so I think it makes sense to do it all in a single PR.

### (1) Making `HubSwitchGuard` `!Send`

This PR makes `HubSwitchGuard` `!Send` by adding `PhantomData<MutexGuard<'static, ()>>` while keeping it `Sync`. The type system now prevents the guard from being moved across threads at compile time.

This change is important because `HubSwitchGuard` manages thread-local hub state, but was previously `Send`, allowing it to be moved to another thread. When dropped on the wrong thread, it could corrupt that thread's hub state instead of restoring the original thread. This change resolves #943.

**Tests related to this change:** `!Send` is enforced by the compiler, so there are no additional tests here.

### (2) A new stack for spans to manage `HubSwitchGuards`

As `HubSwitchGuard` is now `!Send`, we needed a new way to manage the `HubSwitchGuard` associated with a given span, in the `tracing` integration. Previously, we just had the guards live directly on the span via the `SentrySpanData` type, but this no longer works, since spans need to be `Send`. So, we now declare a thread-local mapping from span IDs to `HubSwitchGuard`s. The guards are stored in a stack, as each span will have one guard per span entry (spans can be entered multiple times, even on the same thread, and without the stack, we would restore the original hub too early). We drop the guards on span exit in LIFO order.

**Tests related to this change:** [`span_reentrancy.rs`](https://github.com/getsentry/sentry-rust/pull/957/changes#diff-203b34b6e9f8f4bad9c8c43e05c781e119eea7b27648f257bfe7b35e912ba2b1) contains tests to validate correct span tree structure when re-entering the same span multiple times. A previous iteration of this PR only allowed a single guard per span; the test failed against this implementation because it produces an incorrect span structure (two transactions instead of one). The test only passes with the guard stack.

### (3) Forking the `Hub` on each span (re-)entry

Change (2) is insufficient to make the [`span_reentrancy.rs`](https://github.com/getsentry/sentry-rust/pull/957/changes#diff-203b34b6e9f8f4bad9c8c43e05c781e119eea7b27648f257bfe7b35e912ba2b1) test pass because with only that change, there is still the fundamental problem that each span does not get its own `Hub` to manage state. Thus, in that test, I believe we were actually only ever using one hub, because there was no place we were forking the hub. So, on span exit, we prematurely [set the parent span on the hub](https://github.com/getsentry/sentry-rust/pull/957/changes#diff-5acb70e20dc764b608e1acf81b57fea59308624b7c2bc87906b310ff8b1f0eb2L372).

Forking the hub ensures proper isolation, so the span gets set back at the right time (I also [suspect](https://github.com/getsentry/sentry-rust/pull/957/changes#r2773227449) we don't need to manually set it back, but I am unsure).

This change resolves #946.

**Tests related to this change:** [`span_cross_thread.rs`](https://github.com/getsentry/sentry-rust/pull/957/changes#diff-aa629e96442a0995ed2fd39dd68a18dd4be293732d2435b9aa5e03c848e12c38) ensures that entering the same span in multiple threads produces the correct span tree. Basically, it is a reproduction of #946. [`span_reentrancy.rs`](https://github.com/getsentry/sentry-rust/pull/957/changes#diff-203b34b6e9f8f4bad9c8c43e05c781e119eea7b27648f257bfe7b35e912ba2b1) also only passes with this change.

## Issues

- Fixes #943
- Fixes [RUST-130](https://linear.app/getsentry/issue/RUST-130/hubswitchguard-should-not-be-send)
- Fixes #946 
- Fixes [RUST-132](https://linear.app/getsentry/issue/RUST-132/entering-the-same-span-several-times-causes-esoteric-behavior)

### DIFF
--- a/sentry-core/src/macros.rs
+++ b/sentry-core/src/macros.rs
@@ -113,14 +113,22 @@ mod tests {
     }
 
     #[test]
+    #[cfg(debug_assertions)]
     #[should_panic(expected = "assertion failed: 1 == 2")]
     fn debug_assert_or_log_panics_with_default_message_when_condition_fails() {
         crate::debug_assert_or_log!(1 == 2);
     }
 
     #[test]
+    #[cfg(debug_assertions)]
     #[should_panic(expected = "custom invariant message")]
     fn debug_assert_or_log_panics_with_custom_message_when_condition_fails() {
         crate::debug_assert_or_log!(false, "custom invariant message");
+    }
+
+    #[test]
+    #[cfg(not(debug_assertions))]
+    fn no_panic_without_debug_assertions() {
+        crate::debug_assert_or_log!(false, "should not panic");
     }
 }


sentry-tracing/src/layer/mod.rs
```diff
@@ -236,7 +236,6 @@ pub(super) struct SentrySpanData {
     pub(super) sentry_span: TransactionOrSpan,
     parent_sentry_span: Option<TransactionOrSpan>,
     hub: Arc<sentry_core::Hub>,
-    hub_switch_guard: Option<sentry_core::HubSwitchGuard>,
```
szokeasaurusrex, line 239: @lcian I've been using Claude to work on this PR. It removed this field to keep SentrySpanData `Send` even after `HubSwitchGuard` has been made `!Send`. However, to me, it does not make that much sense that a span should move between threads, so I am not sure it should be `Send`. Claude says maybe in async contexts it makes sense. 

What do you think?

lcian: The span data needs to be `Send` due to this trait bound
https://github.com/tokio-rs/tracing/blob/efc690fa6bd1d9c3a57528b9bc8ac80504a7a6ed/tracing-subscriber/src/registry/extensions.rs#L87

Logically, that makes complete sense. 
Say I have:
```rust
#[tracing::instrument]
async fn f() {
  // some async code
}
``` 
Then the behavior I want is for the span tracking the execution of `f` to be entered and exited every time the future representing `f` is polled. And if I'm using tokio, in general, the future can be polled on any thread, it can freely move.
Therefore a `Span` and its extensions need to be `Send`.

Technically, the extensions are not stored on the `Span` itself but rather in the `Registry`, however the intuition above still stands.

I will have a better look at this PR tomorrow.

lcian: This is the same reason why `Hub` itself needs to be send by the way.
Anyways I think this PR might be beneficial for the case that the user reported, and I would like a test which we can run with and without the change to see the difference.

It is possible that to do this test you will need a way to distinguish between hubs which right now is not directly possible. For a while I wanted to introduce a UUID on the Hub just to identify which would be helpful for debugging and tests like this, so that should be a way.

szokeasaurusrex: Ok, I will try this out. It is likely I will only get around to it during the week of January 26, since I have some higher prio stuff the rest of this week and vacation next week

sentry-tracing/src/layer.rs
```diff
@@ -366,18 +367,24 @@ where
...
+        let extensions = span.extensions();
+        if let Some(data) = extensions.get::<SentrySpanData>() {
             data.hub.configure_scope(|scope| {
                 scope.set_span(data.parent_sentry_span.clone());
             });
-            data.hub_switch_guard.take();
         }
+        SPAN_GUARDS.with(|guards| {
+            guards.borrow_mut().remove(id);
+        });
```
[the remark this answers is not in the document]
szokeasaurusrex: The scenario described requires closing a span while an entered guard is still active, which `tracing` appears to disallow.

That said, we can still take a defensive approach and remove the guard from thread-local storage before consulting the registry so it won’t leak even if someone does this.

lcian:
@szokeasaurusrex it rewrites the original guard with the same one I think, so what's the issue there?

sentry-tracing/tests/span_reentrancy.rs
```diff
@@ -0,0 +1,70 @@
...
+            let _reenter_a = span_a.enter();
+        }
+
+        // Create another span while the original span is still entered to ensure
+        // it is recorded on the same transaction rather than starting a new one.
+        let span_b = tracing::info_span!("b");
+        {
+            let _enter_b = span_b.enter();
+        }
+    }
```
szokeasaurusrex, line 26: @lcian This reproduces the bug I was talking about. With the original implementation, `span_a` and `span_b` ended up as two separate transactions because dropping `_reenter_a` restored the original hub. There were two underlying issues here that were causing this behavior; I fixed both of them here.

The first fix was that I changed the `SPAN_GUARDS` so that it would not just have a single guard per span, but rather, to have a stack of guards that are pushed on each entry and popped on each exit. This way, we only restored the original hub on the final exit from the span.

That first fix was only part of the problem, though. To get the span structure correct here, I also needed to fix #946. For that, we needed to be forking the hub, as you suggested last week. Otherwise, if we just use the hub on the span directly, we end up not actually getting proper scope isolation.

sentry-tracing/src/layer/mod.rs
```diff
@@ -366,18 +382,29 @@ where
             None => return,
         };
 
-        let mut extensions = span.extensions_mut();
-        if let Some(data) = extensions.get_mut::<SentrySpanData>() {
+        let extensions = span.extensions();
+        if let Some(data) = extensions.get::<SentrySpanData>() {
             data.hub.configure_scope(|scope| {
                 scope.set_span(data.parent_sentry_span.clone());
             });
```
szokeasaurusrex: Now that we are forking the hub, I actually don't think we need to be setting the parent span on the hub here; the hub we restore should already have the parent span set (unless some other code intentionally changed the span on that hub, which we probably should respect).

```suggestion
```

What do you think @lcian? Am I missing something?

lcian: Yeah, I agree. And with that we won't even need to store `parent_sentry_span` in the extensions anymore.

sentry-tracing/src/layer/mod.rs
```diff
@@ -334,47 +335,66 @@ where
...
     }
 
-    /// Sets entered span as *current* sentry span. A tracing span can be
-    /// entered and existed multiple times, for example, when using a `tracing::Instrumented` future.
+    /// Sets the entered span as *current* sentry span.
+    ///
+    /// A tracing span can be entered and exited multiple times, for example,
+    /// when using a `tracing::Instrumented` future.
+    ///
+    /// Spans must be exited on the same span that they are entered. The
```
lcian:
```suggestion
    /// Spans must be exited on the same thread that they are entered. The
```

sentry-tracing/tests/await_cross_thread_guard.rs
```diff
@@ -0,0 +1,187 @@
...
+use std::task::{Context, Poll, Waker};
+use std::thread;
+
+use sentry::protocol::{EnvelopeItem, Transaction};
+use sentry::{Envelope, Hub, HubSwitchGuard};
+use tracing::Span;
+
+/// Test that the [`sentry_tracing::SentryLayer`]'s `on_exit` implementation panics
+/// (only when `debug_assertions` are enabled) if a span, captured by Sentry, is exited
+/// on a span that was entered on a different thread than where it was exited.
```
lcian:
```suggestion
/// (only when `debug_assertions` are enabled) if a span, captured by Sentry, is exited
/// on a different thread than where it was exited.
```

lcian: entered*

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@
 - Added a `Envelope::into_items` method, which returns an iterator over owned [`EnvelopeItem`s](https://docs.rs/sentry/0.46.2/sentry/protocol/enum.EnvelopeItem.html) in the [`Envelope`](https://docs.rs/sentry/0.46.2/sentry/struct.Envelope.html) ([#983](https://github.com/getsentry/sentry-rust/pull/983)).
 - Expose transport utilities ([#949](https://github.com/getsentry/sentry-rust/pull/949))
 
+### Fixes
+
+- Fixed thread corruption bug where `HubSwitchGuard` could be dropped on wrong thread ([#957](https://github.com/getsentry/sentry-rust/pull/957)).
+  - **Breaking change**: `sentry_core::HubSwitchGuard` is now `!Send`, preventing it from being moved across threads. Code that previously sent the guard to another thread will no longer compile.
+- We now fork the `Hub` every time a span is entered. This prevents data from leaking across spans ([#957](https://github.com/getsentry/sentry-rust/pull/957)).
```
szokeasaurusrex, line 14: @lcian does this changelog entry look good?

lcian: Yep looks good

PR #958, opened 2026-01-14
build(deps): update `reqwest` from v0.12.25 to v0.13.1

AlexTMjugador:
### Description

This PR updates `reqwest` from v0.12.25 to v0.13.1. From a user perspective, the most noteworthy change is that [it switches the default `rustls` provider from `ring` to `aws-lc-rs`](https://github.com/seanmonstar/reqwest/blob/master/CHANGELOG.md#v0130). This brings `reqwest` in line with the broader ecosystem's move toward `aws-lc-rs` as the default and increases the likelihood that users will no longer need to include multiple TLS implementations in their binaries.

Stacked on #968 

#### Issues

This resolves a personal goal of mine, as a user, to remove `ring` from my transitive dependency tree. However, it does not address an issue that was previously documented in this repository.


Dav1dde (approved):
Yep, looks good 🙏

szokeasaurusrex:
@AlexTMjugador @Iskander508 I have opened #998, which supersedes this PR

PR #959, opened 2026-01-19
chore(repo): Add Claude Code settings with basic permissions

philipphofmann:
Adds `.claude/settings.json` to enable Claude Code with:
- Basic bash commands (find, ls, git, grep, mv)
- WebFetch access to GitHub and Sentry documentation sites

This follows the same pattern used in other Sentry SDK repositories like sentry-javascript, sentry-dart, and sentry-react-native.

Note: This adds .claude/settings.json with -f flag since .claude is in .gitignore for local settings.


.claude/settings.json
```diff
@@ -0,0 +1,34 @@
...
+      "Bash(gh pr list:*)",
+      "Bash(gh pr checks:*)",
+      "Bash(gh pr diff:*)",
+      "Bash(gh issue view:*)",
+      "Bash(gh issue list:*)",
+      "Bash(gh run view:*)",
+      "Bash(gh run list:*)",
+      "Bash(gh run logs:*)",
+      "Bash(gh repo view:*)",
+      "WebFetch(domain:github.com)",
```
szokeasaurusrex: l: I want my Claude agent to access GitHub URLs directly via `gh`, as that avoids fetching the full page source, and would recommend others also do the same.

```suggestion
```

.claude/settings.json
```diff
@@ -0,0 +1,34 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(find:*)",
```
szokeasaurusrex: m: Per @lcian's point that this can allow arbitrary code execution, I would remove this.

```suggestion
```

.claude/settings.json
```diff
@@ -0,0 +1,34 @@
...
+      "Bash(ls:*)",
+      "Bash(git:*)",
```
szokeasaurusrex: question: Would this allow `git push` to be run without prompting? I have talked with some folks who don't like giving push permissions to agents, so if it does, let's remove it (it also seems redundant with all the git subcommands being listed below).


```suggestion
```

.claude/settings.json
```diff
@@ -0,0 +1,34 @@
...
+      "Bash(gh issue list:*)",
+      "Bash(gh run view:*)",
+      "Bash(gh run list:*)",
+      "Bash(gh run logs:*)",
+      "Bash(gh repo view:*)",
+      "WebFetch(domain:github.com)",
+      "WebFetch(domain:docs.sentry.io)",
+      "WebFetch(domain:develop.sentry.dev)",
+      "Bash(grep:*)",
+      "Bash(mv:*)"
```
szokeasaurusrex: m: Likewise, I am hesitant to allow agents to modify the filesystem arbitrarily as a default.

```suggestion
```

.claude/settings.json
```diff
@@ -0,0 +1,30 @@
...
+      "Bash(gh pr diff:*)",
+      "Bash(gh issue view:*)",
+      "Bash(gh issue list:*)",
+      "Bash(gh run view:*)",
+      "Bash(gh run list:*)",
+      "Bash(gh run logs:*)",
+      "Bash(gh repo view:*)",
+      "WebFetch(domain:docs.sentry.io)",
+      "WebFetch(domain:develop.sentry.dev)",
+      "Bash(grep:*)",
```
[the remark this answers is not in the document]
fbernier: @szokeasaurusrex  Why was this ignored? claude code does error out when launching because of this trailing comma...

PR #968, opened 2026-02-02
build: `cargo update`

szokeasaurusrex:
Run `cargo update` to bump all dependencies to the latest versions.

Stacked on #970.


Dav1dde (approved):
Makes sense, but looks like it breaks MSRV as crates may not consider MSRV for breaking changes.

PR #969, opened 2026-02-03
ref(cargo): Extract common items to workspace

szokeasaurusrex:
Extract all common package items, [which can be extracted](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table), to the workspace `Cargo.toml`, except for the `version` field.

## Note on `version` & `license`

We cannot extract `version` because it causes a parse error when [generating the crate `README.md` files](https://github.com/getsentry/sentry-rust/blob/c203754655ea2cbb62632ad43ceac9547b1da2e9/scripts/update-readme.sh#L9) while [bumping the version](https://github.com/getsentry/sentry-rust/blob/c203754655ea2cbb62632ad43ceac9547b1da2e9/scripts/bump-version.sh#L16) prior to release, as the [`cargo-readme`](https://crates.io/crates/cargo-readme) tool we use to generate these files apparently cannot parse `version = { workspace = true }`, as it expects version to be set to a string. Similarly, `cargo readme` also fails if `license` is set with `{ workspace = true }`

@Dav1dde, do you have any suggestions for how to get around this? I think the way forward may be to simply stop using `cargo readme`, as our README files are quite verbose. Instead, I would propose having a short static README for each crate, directing users interested in full docs over to our docs.rs pages. What do you think?

In any case, this should happen in a separate PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,3 +16,10 @@ members = [
     "sentry-tracing",
     "sentry-types",
 ]
+
+[workspace.package]
+authors = ["Sentry <hello@sentry.io>"]
+repository = "https://github.com/getsentry/sentry-rust"
+homepage = "https://sentry.io/welcome/"
+edition = "2021"
+rust-version = "1.81"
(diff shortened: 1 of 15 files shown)


lcian (approved):

szokeasaurusrex:
> I assume it will also not accept the shorthand: `<dep>.workspace = true`.

I will double-check this

PR #970, opened 2026-02-03
chore(msrv): Increase MSRV to 1.88

szokeasaurusrex:
Bump the MSRV to 1.88 to allow dependency updates (see #968). I do not think this change requires a major version bump, as our [MSRV policy](https://github.com/getsentry/sentry-rust?tab=readme-ov-file#requirements) states we try to support the last six months of Rust versions, and 1.88 is more than six months old.

Stacked on #969.


lcian:
If you don't want to bump to 1.88, I believe it's also possible to address this by bumping to 1.85 and switching to `edition = "2024"` and therefore `resolver = "3"`.
That way, `cargo update` should only pick versions of the deps that are compatible with that MSRV.

Issue #973, opened 2026-02-03
Delete the Makefile and Remove References


szokeasaurusrex:
Remove the Makefile and replace any `make` usage in CI with explicit commands.

PR #987, opened 2026-02-11
ci: Replace make targets and remove Makefile

szokeasaurusrex:
Replace CI make invocations with explicit cargo commands to make workflow behavior easier to read directly in the CI config.

Remove the root Makefile to avoid maintaining a separate command layer in addition to the workflow definitions.

_Stacked on #986_

Closes GH-973
Closes [RUST-139](https://linear.app/getsentry/issue/RUST-139/delete-the-makefile-and-remove-references)


.github/workflows/ci.yml
```diff
@@ -40,7 +40,45 @@ jobs:
...
+        run: cargo check --no-default-features --features curl,panic
+        working-directory: sentry
+        env:
+          RUSTFLAGS: -Dwarnings
+
+      - name: Check sentry-actix
+        run: cargo check
+        working-directory: sentry-actix
+        env:
+          RUSTFLAGS: -Dwarnings
```
lcian, line 81: We could probably expand this by checking each of the subcrates individually, I think we just forgot to add them to the makefile as new crates were created

szokeasaurusrex: Is there any benefit to checking the sub crates individually? Shouldn't we just run check at the workspace level?

lcian: Yeah but you would also need to make sure you're selecting the features that enable a particular subcrate.
e.g. instead of `cargo check` in `sentry-actix` working directory you would cargo check `sentry` with the `actix` feature, etc.

szokeasaurusrex: Ok, got it. Let's do this in a later PR; for now I am trying to keep a narrow scope on the CI improvement effort

PR #989, opened 2026-02-12
ci: align stable and MSRV check/test coverage

### LINKED ISSUE #974: Align MSRV and Stable CI Coverage (Including Doc Tests)
Update MSRV to run the same checks and tests as stable, including doc tests, so both lanes are equivalent. To ensure the two don't drift apart, we probably will want to have the same matrix running both stable and MSRV. We should ensure doc tests are part of PR CI; everything currently being tested in either MSRV or stable currently should be tested in both after the changes.

Depends on #972 and #973.

szokeasaurusrex:
### Description
Run both check and test over a shared os×rust matrix for stable and 1.88.0, and remove the standalone MSRV job to reduce drift.

Keep check permutations identical across toolchains, add explicit workspace doc-test runs in test, and retain an explicit sentry default+test feature assertion with an inline rationale comment.

#### Issues
Closes #974
Closes [RUST-140](https://linear.app/getsentry/issue/RUST-140/align-msrv-and-stable-ci-coverage-including-doc-tests)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,13 +33,19 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+        rust: [stable, 1.88.0]
 
-    name: Check feature permutations using Rust stable on ${{ matrix.os }}
+    name: Check feature permutations using Rust ${{ matrix.rust }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
+
+      - name: Install rust ${{ matrix.rust }} toolchain
+        run: |
+          rustup toolchain install ${{ matrix.rust }} --profile minimal --no-self-update
+          rustup default ${{ matrix.rust }}
 
       - uses: Swatinem/rust-cache@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -119,20 +107,16 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - name: Check sentry-core without default features
-        run: cargo check --no-default-features
```
lcian, line 112: We removed this step, why is that?

lcian: Ah I think it might be because the other workflow already does this

szokeasaurusrex:
> LGTM but the downside is that you'll have to keep the version here in this file in sync with the MSRV

The previous approach [had the same limitation](https://github.com/getsentry/sentry-rust/pull/989/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL106). I agree, we can probably improve this in the future though!

.github/workflows/ci.yml
szokeasaurusrex: Indeed, [the check has moved here](https://github.com/getsentry/sentry-rust/pull/989/changes/BASE..c4ca361e9541387e5ead48839b77e87b0758261f#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL48)

PR #996, opened 2026-02-16
[wip] plan for supporting client reports

szokeasaurusrex:
Ref #996
Ref [RUST-98](https://linear.app/getsentry/issue/RUST-98/support-and-send-client-reports)


docs/plan.md
```diff
@@ -0,0 +1,207 @@
...
+4. HTTP handling:
+   - `2xx`: success
+   - `4xx/5xx` (except `429`): discard and record `send_error`
+   - `413`: discard and record `send_error`
+   - `429`: discard but **do not** record client report (to avoid double counting)
+5. Logs:
+   - track dropped log count as `log_item`
+   - track dropped log size as `log_byte` (approximate size is acceptable)
+6. Logs buffering spec also requires a hard cap of 1000 queued logs.
+   - **Scope note for #882:** keep implementation scope narrow to client reports; track the 1000-cap as a separate follow-up issue unless we discover a low-risk one-line fix while implementing.
```
lcian, line 26: Relay is optimized for up to 100 logs per envelope. SDKs MUST NOT send more than 100 logs per envelope.

To prevent out-of-memory issues in customer applications, SDKs MUST enforce a hard limit of 1000 queued log events. Any logs added beyond this limit are dropped. SDKs MAY use a lower limit but MUST NOT exceed 1000.

szokeasaurusrex: Bad clanker

Will update this!

lcian (approved):
Looks good generally! 🚀

docs/plan.md
szokeasaurusrex: On further reflection, I think adding this buffer limit is out of scope in this client report project, I have made a [follow up issue](https://github.com/getsentry/sentry-rust/issues/999) though!

szokeasaurusrex:
tidying up 🧹

Issue #938, opened 2025-11-16
Sentry Metrics for Rust


dingsdax:
[https://develop.sentry.dev/sdk/telemetry/metrics/](https://develop.sentry.dev/sdk/telemetry/metrics/)

Add support for [Sentry Metrics ](https://docs.sentry.io/product/explore/metrics/getting-started/)to the Rust SDK!

This will enable users to record [counters, gauges, and distributions](https://develop.sentry.dev/sdk/telemetry/metrics/#metric-types).

1. Define the [Metrics Protocol](https://develop.sentry.dev/sdk/telemetry/metrics/#metrics-protocol) and \`trace_metric envelope type in the SDK.
2. Add the Public API (SDK options and methods) as per the [docs](https://develop.sentry.dev/sdk/telemetry/metrics/#public-api).
3. Make sure the the SDK follow the documented [behavior](https://develop.sentry.dev/sdk/telemetry/metrics/#sdk-behavior).
4. Attach [default attributes](https://develop.sentry.dev/sdk/telemetry/metrics/#default-attributes) and optional [user attributes](https://develop.sentry.dev/sdk/telemetry/metrics/#user-attributes) to the SDK as per docs.
5. Follow requirements for rate limiting, buffering, batching & flushing and build in measures to prevent excessive memory usage.

For general feedback & questions about Sentry Metrics use this [GH discussion](https://github.com/getsentry/sentry/discussions/102275); for Rust SDK specific feedback & questions use the comment functionality on this GH issue.

lcian:
We should consider what are the most widely adopted crates for capturing metrics and consider building integrations for those, as that way it will be much more easier for this to see adoption. There's for example https://crates.io/crates/metrics and there's also https://crates.io/crates/cadence which we integrated with during the old metrics beta, even though I believe our new protocol won't be compatible with the latter.

thomaseizinger:
I am very interested in adopting this. Happy to help with any early testing you've got.

PR #997, opened 2026-02-17
feat: Add trace metrics support

jjbayer:
This PR introduces logic for emiting & batching trace metrics, largely copy-pasted from logs.

Usage example:

```rust
sentry::metric_count!("api.requests", 1, route = "/users", method = "GET");
sentry::metric_gauge!("memory.usage", 1024.0, unit = "byte");
sentry::metric_distribution!("response.time", 150.0, unit = "millisecond", route = "/users");
```

Resolves https://github.com/getsentry/sentry-rust/issues/938.

### DIFF
--- a/sentry-core/src/batcher.rs
+++ b/sentry-core/src/batcher.rs
@@ -268,14 +268,17 @@ mod tests {
         }
 
         #[test]
-        fn test_metrics_disabled_by_default() {
+        fn test_metrics_disabled_explicitly() {
             let envelopes = test::with_captured_envelopes_options(
                 || {
                     for i in 0..10 {
                         crate::Hub::current().capture_metric(test_metric(&format!("metric.{i}")));
                     }
                 },
-                crate::ClientOptions::default(),
+                crate::ClientOptions {
+                    enable_metrics: false,
+                    ..Default::default()
+                },
             );
 
             assert_eq!(0, envelopes.len());
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -579,7 +579,7 @@ impl Client {
     /// processing it through `before_send_metric`.
     #[cfg(feature = "metrics")]
     fn prepare_metric(&self, mut metric: TraceMetric, scope: &Scope) -> Option<TraceMetric> {
-        scope.apply_to_metric(&mut metric);
+        scope.apply_to_metric(&mut metric, self.options.send_default_pii);
 
         if let Some(default_attributes) = self.default_metric_attributes.as_ref() {
             for (key, val) in default_attributes.iter() {
--- a/sentry-core/src/clientoptions.rs
+++ b/sentry-core/src/clientoptions.rs
@@ -174,7 +174,7 @@ pub struct ClientOptions {
     /// Determines whether captured structured logs should be sent to Sentry (defaults to false).
     #[cfg(feature = "logs")]
     pub enable_logs: bool,
-    /// Determines whether captured trace metrics should be sent to Sentry (defaults to false).
+    /// Determines whether captured trace metrics should be sent to Sentry (defaults to true).
     #[cfg(feature = "metrics")]
     pub enable_metrics: bool,
     /// Callback that is executed for each TraceMetric being added.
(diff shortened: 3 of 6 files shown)


sentry-core/src/batcher.rs
```diff
@@ -0,0 +1,337 @@
+//! Generic batching for Sentry envelope items (logs, metrics, etc.).
```
jjbayer, line 1: This was moved & generalized from `logs.rs`.

sentry-core/src/logs.rs
jjbayer, line 1: This has moved to `batcher.rs`.

sentry-core/src/metrics.rs
```diff
@@ -0,0 +1,110 @@
...
+//!
+//! # Examples
+//!
+//! ```
+//! sentry::metrics::count("api.requests").emit(1);
+//!
+//! sentry::metrics::distribution("response.time")
+//!     .with_attribute("route", "my_route")
+//!     .with_unit("millisecond")
+//!     .emit(0.123);
```
jjbayer: This is maybe the most controversial piece of this PR: Should we use macros instead, like we do for logs?

sentry-core/src/clientoptions.rs
```diff
@@ -172,6 +174,15 @@ pub struct ClientOptions {
     /// Determines whether captured structured logs should be sent to Sentry (defaults to false).
     #[cfg(feature = "logs")]
     pub enable_logs: bool,
+    /// The sample rate for trace metrics. (0.0 - 1.0, defaults to 0.0)
+    ///
+    /// Metrics are disabled by default. Sampled metrics are annotated with a
+    /// `sentry.client_sample_rate` attribute so the server can extrapolate.
+    #[cfg(feature = "metrics")]
+    pub metrics_sample_rate: f32,
```
jjbayer: The develop docs state that this should be a boolean flag, and that it should go into `_experimental`: https://develop.sentry.dev/sdk/telemetry/metrics/#initialization-options

I'm willing to change this but I still think it makes sense to have a sample rate.

sentry-core/src/client.rs
```diff
@@ -247,6 +280,35 @@ impl Client {
...
+            "sentry.sdk.version".to_owned(),
+            self.sdk_info.version.to_owned().into(),
+        );
+
+        if let Some(server) = &self.options.server_name {
+            attributes.insert("server.address".to_owned(), server.clone().into());
+        }
+
+        self.default_metric_attributes = Some(attributes);
+    }
```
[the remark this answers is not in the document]
lcian: The OS attributes aren't supposed to go on there https://develop.sentry.dev/sdk/telemetry/metrics/#default-attributes

sentry-core/src/batcher.rs
```diff
@@ -0,0 +1,345 @@
...
+        #[test]
+        fn test_metrics_batching() {
+            let envelopes = test::with_captured_envelopes_options(
+                || {
+                    for i in 0..150 {
+                        crate::Hub::current().capture_metric(test_metric(&format!("metric.{i}")));
+                    }
+                },
+                crate::ClientOptions {
+                    metrics_sample_rate: 1.0,
```
lcian: Spec doesn't mention any `metrics_sample_rate`, I don't think we should have it.

jjbayer: Will revert to a boolean flag. I need sampling for the relay use case (I wouldn't dare to send the entire volume at once), but I can add the `sentry.client_sample_rate` attribute there manually.

What about what [the specs say about an `_experiments` section](https://develop.sentry.dev/sdk/telemetry/metrics/#initialization-options), should we move the flag there like the spec suggests?

lcian: I guess we could also keep the sample rate, if this is going to be used by internal services we're gonna need it at some point anyways.

Another approach which I prefer for Rust is to feature flag it and make sure we state that it's clear that the flag is experimental and could be removed.

jjbayer: Keeping the `enable_metrics` flag for now to keep it simple & spec-conformant. The spec now seems to allow [putting it in the stable section](https://develop.sentry.dev/sdk/telemetry/metrics/#initialization-options).

jjbayer:
Stale.

PR #1022, opened 2026-03-02
feat(sentry-types): Add metric protocol envelope support

### LINKED ISSUE #1008: Add trace metric protocol models and envelope item container support
This issue captures the protocol/envelope portion of [#997](<https://github.com/getsentry/sentry-rust/pull/997>), so later client work can build on a small, reviewable base.

### Scope from getsentry/sentry-rust#997

* [`sentry-types/src/protocol/v7.rs`](<https://github.com/getsentry/sentry-rust/blob/343a63d025276b0474882e2dbc2ff27bc5400743/sentry-types/src/protocol/v7.rs>)
* [`sentry-types/src/protocol/envelope.rs`](<https://github.com/getsentry/sentry-rust/blob/343a63d025276b0474882e2dbc2ff27bc5400743/sentry-types/src/protocol/envelope.rs>)

### Details

Extract only the trace metric model and envelope container implementation described in the [trace metrics spec](<https://develop.sentry.dev/sdk/telemetry/metrics/>): metric type enum, metric payload fields, `trace_metric` envelope item type, and encode/decode handling including correct content type and item count behavior. Keep this part free of client capture or macro API concerns, and include round-trip coverage in `sentry-types` to show metric items survive serialization/deserialization intact.

szokeasaurusrex:
Add `Metric` protocol types and `trace_metric` envelope container serialization/deserialization in `sentry-types`.

Stacked on #1072

Closes #1008
Closes [RUST-159](https://linear.app/getsentry/issue/RUST-159/add-trace-metric-protocol-models-and-envelope-item-container-support)

### DIFF
--- a/sentry-types/src/protocol/v7.rs
+++ b/sentry-types/src/protocol/v7.rs
@@ -2369,6 +2369,43 @@ impl<'de> Deserialize<'de> for LogAttribute {
     }
 }
 
+/// The type of a [metric](https://develop.sentry.dev/sdk/telemetry/metrics/).
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum MetricType {
+    /// A counter metric that only increments.
+    Counter,
+    /// A gauge metric that can go up and down.
+    Gauge,
+    /// A distribution metric for statistical spread measurements.
+    Distribution,
+}
+
+/// A single [metric](https://develop.sentry.dev/sdk/telemetry/metrics/).
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct Metric {
+    /// The metric type.
+    pub r#type: MetricType,
+    /// The metric name. Uses dot separators for hierarchy.
+    pub name: Cow<'static, str>,
+    /// The numeric value.
+    pub value: f64,
+    /// The timestamp when recorded.
+    #[serde(with = "ts_seconds_float")]
+    pub timestamp: SystemTime,
+    /// The trace ID this metric is associated with.
+    pub trace_id: TraceId,
+    /// The span ID of the active span, if any.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub span_id: Option<SpanId>,
+    /// The measurement unit.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub unit: Option<Unit>,
+    /// Additional key-value attributes.
+    #[serde(default, skip_serializing_if = "Map::is_empty")]
+    pub attributes: Map<Cow<'static, str>, LogAttribute>,
+}
+
 /// An ID that identifies an organization in the Sentry backend.
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
 pub struct OrganizationId(u64);
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -10,8 +10,8 @@ use crate::Dsn;
 use super::v7 as protocol;
 
 use protocol::{
-    Attachment, AttachmentType, ClientSdkInfo, DynamicSamplingContext, Event, Log, MonitorCheckIn,
-    SessionAggregates, SessionUpdate, Transaction,
+    Attachment, AttachmentType, ClientSdkInfo, DynamicSamplingContext, Event, Log, Metric,
+    MonitorCheckIn, SessionAggregates, SessionUpdate, Transaction,
 };
 
 /// Raised if a envelope cannot be parsed from a given input.
@@ -127,6 +127,10 @@ enum EnvelopeItemType {
     /// A container of Log items.
     #[serde(rename = "log")]
     LogsContainer,
+    /// A container of Metric items.
+    /// Serialized to a `trace_metric` envelope item.
+    #[serde(rename = "trace_metric")]
+    MetricsContainer,
 }
 
 /// An Envelope Item Header.
@@ -192,6 +196,8 @@ pub enum EnvelopeItem {
 pub enum ItemContainer {
     /// A list of logs.
     Logs(Vec<Log>),
+    /// A list of metrics.
+    Metrics(Vec<Metric>),
 }
 
 #[allow(clippy::len_without_is_empty, reason = "is_empty is not needed")]
@@ -200,6 +206,7 @@ impl ItemContainer {
     pub fn len(&self) -> usize {
         match self {
             Self::Logs(logs) => logs.len(),
+            Self::Metrics(metrics) => metrics.len(),
         }
     }
 
@@ -207,6 +214,7 @@ impl ItemContainer {
     pub fn ty(&self) -> &'static str {
         match self {
             Self::Logs(_) => "log",
+            Self::Metrics(_) => "trace_metric",
         }
     }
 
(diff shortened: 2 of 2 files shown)


sentry-types/src/protocol/v7.rs
```diff
@@ -2368,6 +2368,45 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    Distribution,
+}
+
+/// A single [trace metric](https://develop.sentry.dev/sdk/telemetry/metrics/).
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub struct TraceMetric {
+    /// The metric type.
+    pub r#type: TraceMetricType,
+    /// The metric name . Uses dot separators for hierarchy.
```
lcian:
```suggestion
    /// The metric name. Uses dot separators for hierarchy.
```

sentry-types/src/protocol/v7.rs
```diff
@@ -2368,6 +2368,45 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    pub name: String,
+    /// The numeric value.
+    pub value: f64,
```
lcian, line 2392: Is this fine?
It's technically not defined in the spec, but logically you should only be able to send ints for counters while you should be able to send floats for e.g. gauges.
I wonder what happens if you send a counter but serialize the value as 1.0, I guess it depends on how loosely this is defined in Relay/other systems.

sentry-types/src/protocol/v7.rs
```diff
@@ -2368,6 +2368,45 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    pub trace_id: TraceId,
+    /// The span ID of the active span, if any.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub span_id: Option<SpanId>,
+    /// The measurement unit.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub unit: Option<String>,
+    /// Additional key-value attributes.
+    #[serde(default, skip_serializing_if = "Map::is_empty")]
+    pub attributes: Map<String, LogAttribute>,
```
lcian: I think we could rename this to `Attribute` and then alias `type LogAttribute = Attribute` so that it's not a breaking change if anyone is using it.
wdyt?

sentry-types/src/protocol/v7.rs
szokeasaurusrex: Looks like Relay [converts all metric values to `f64`](https://github.com/getsentry/relay/blob/4e991d495646a70286714e772cd577cb6868237e/relay-server/src/processing/trace_metrics/store.rs#L101-L108). There's also an example test where a [counter is sent with a float value of `1.0`](https://github.com/getsentry/relay/blob/4e991d495646a70286714e772cd577cb6868237e/tests/integration/test_trace_metrics.py#L208-L209).

So, I think it makes sense to stick with `f64`.

sentry-types/src/protocol/v7.rs
szokeasaurusrex: Yeah, that could make sense; although, I am slightly concerned that the name `Attribute` is maybe a bit vague. What if we would want to use `Attribute` for something else one day? Thoughts?

In any case, let's make this change in a separate PR.

sentry-types/src/protocol/v7.rs
```diff
@@ -2368,6 +2368,45 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+}
```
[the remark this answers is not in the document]
szokeasaurusrex: We will fix this if and when it becomes a problem.

sentry-types/src/protocol/v7.rs
lcian: Well the attributes API we're gonna introduce will take exactly the same types so the concepts would match.
I'm fine with making the change in a separate PR, but should probably be before we ship this, otherwise it would seem confusing to me to have something of type LogAttribute inside a TraceMetric.

sentry-types/src/protocol/v7.rs
lcian: Cool. This is technically not defined as part of the spec so I think it's fine to match what Relay takes, as that's very unlikely to become more restrictive over time due to backwards compatibility guarantees.
As part of the API we should probably take an integer and convert to float internally for counters though.

lcian (approved):
Looks good to me.

sentry-types/src/protocol/v7.rs
```diff
@@ -2368,6 +2368,45 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    /// A gauge metric that can go up and down.
+    Gauge,
+    /// A distribution metric for statistical spread measurements.
+    Distribution,
+}
+
+/// A single [trace metric](https://develop.sentry.dev/sdk/telemetry/metrics/).
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub struct TraceMetric {
```
lcian: Let's call it `Metric`.

sentry-types/src/protocol/v7.rs
```diff
@@ -2368,6 +2368,45 @@ impl<'de> Deserialize<'de> for LogAttribute {
     }
 }
 
+/// The type of a [trace metric](https://develop.sentry.dev/sdk/telemetry/metrics/).
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, PartialEq)]
+#[serde(rename_all = "lowercase")]
+#[non_exhaustive]
+pub enum TraceMetricType {
```
lcian: Here too.

lcian (changes requested):

sentry-types/src/protocol/v7.rs
```diff
@@ -2368,6 +2368,45 @@ impl<'de> Deserialize<'de> for LogAttribute {
...
+    /// The span ID of the active span, if any.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub span_id: Option<SpanId>,
+    /// The measurement unit.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub unit: Option<String>,
+    /// Additional key-value attributes.
+    #[serde(default, skip_serializing_if = "Map::is_empty")]
+    pub attributes: Map<String, LogAttribute>,
+}
```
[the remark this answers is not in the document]
szokeasaurusrex: To be addressed in future PR

lcian (approved):
LGTM

szokeasaurusrex:
Hey @lcian, I restacked this PR on top of #1072, and changed the type of `unit` accordingly to `Option<Unit>`. What do you think?

Also tagging you @giortzisg in case you would like to review (but it is okay if you skip this one).

force-push-refs.zsh
```diff
@@ -0,0 +1,19 @@
...
+  refs/heads/szokeasaurusrex/metrics-rate-limit
+  refs/heads/szokeasaurusrex/metrics-trace-id
+  refs/heads/szokeasaurusrex/unit-enum
+  refs/heads/szokeasaurusrex/user-attributes-metrics
+)
+
+for ref in "${refs[@]}"; do
+  echo "Pushing $ref -> origin/$ref"
+  git push --force-with-lease origin "$ref:$ref" || break
+done
```
lcian: Remove?

lcian (approved):
Looks good!

giortzisg (approved):

PR #1040, opened 2026-03-27
feat: Make transport channel capacity configurable

### LINKED ISSUE #994: Make transport channel capacity configurable
In [https://github.com/getsentry/sentry-rust/pull/923](<https://github.com/getsentry/sentry-rust/pull/923>), @lorenzocian proposed an improvement to make the transport channel capacity configurable. This could make it less likely that we saturate the capacity of the channel in high-throughput scenarios.

As this change requires us to add a field to ClientOptions, it is breaking, so we decided not to proceed with that PR for now, but it is probably a good improvement to do at some point.

mvanhorn:
## Summary

Adds transport-level `with_channel_capacity` constructors so users can tune the bounded channel size used by the transport thread. The default remains 30, preserving existing behavior.

The original design added a `transport_channel_capacity` field to `ClientOptions`. Per @szokeasaurusrex's review, refactored to additive transport-level methods so `ClientOptions` stays minimal and the feature is opt-in at the transport boundary.

## Why this matters

In high-throughput scenarios (many transactions with single spans each), the hardcoded capacity of 30 can saturate quickly, leading to dropped envelopes. Identified in [#923](<https://github.com/getsentry/sentry-rust/issues/923>), tracked in getsentry/sentry-rust#994. Making it configurable lets users trade memory for reliability based on their workload.

## Changes

* `sentry/src/transports/thread.rs`: `TransportThread::new(send)` keeps its original signature. New `TransportThread::with_capacity(send, channel_capacity)` accepts a custom capacity, clamped to a minimum of 1 to avoid rendezvous channels.
* `sentry/src/transports/tokio_thread.rs`: Same pattern for the async transport variant.
* `sentry/src/transports/curl.rs`: Added `CurlHttpTransport::with_channel_capacity(options, channel_capacity)`. `new(options)` unchanged.
* `sentry/src/transports/reqwest.rs`: Added `ReqwestHttpTransport::with_channel_capacity(options, channel_capacity)`. `new(options)` / `with_client(options, client)` unchanged.
* `sentry/src/transports/ureq.rs`: Added `UreqHttpTransport::with_channel_capacity(options, channel_capacity)`. `new(options)` / `with_agent(options, agent)` unchanged.

## Usage

```rust
let opts = ClientOptions {
    transport: Some(Arc::new(move |opts| {
        Arc::new(ReqwestHttpTransport::with_channel_capacity(opts, 256))
    })),
    ..Default::default()
};
```

## Testing

* `cargo check --workspace --all-features` passes
* `cargo fmt --all` clean
* `cargo clippy --workspace --all-features` clean
* Default behavior unchanged (capacity stays at 30 unless explicitly configured)

Closes getsentry/sentry-rust#994
Closes [RUST-149](https://linear.app/getsentry/issue/RUST-149/make-transport-channel-capacity-configurable)

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/sentry/src/transports/curl.rs
+++ b/sentry/src/transports/curl.rs
@@ -265,10 +265,9 @@ impl CurlHttpTransportOptions {
     /// Set the capacity of the channel that queues envelopes for the background
     /// transport thread (default: 30).
     ///
-    /// A capacity of `0` creates a rendezvous channel: an envelope is accepted
-    /// only when the transport thread is currently waiting on the receiver,
-    /// otherwise it is dropped. A higher capacity reduces the chance of dropped
-    /// events in high-throughput scenarios at the cost of memory.
+    /// A capacity of `0` is clamped to `1`. A higher capacity reduces the
+    /// chance of dropped events in high-throughput scenarios at the cost of
+    /// memory.
     #[inline]
     pub fn with_channel_capacity(self, channel_capacity: usize) -> Self {
         Self {
--- a/sentry/src/transports/curl.rs
+++ b/sentry/src/transports/curl.rs
@@ -7,7 +7,7 @@ use sentry_core::TransportOptions;
 
 use super::{
     thread::{TransportThread, TransportThreadOptions},
-    RateLimiter, HTTP_PAYLOAD_TOO_LARGE, HTTP_PAYLOAD_TOO_LARGE_MESSAGE,
+    RateLimiter, DEFAULT_CHANNEL_CAPACITY, HTTP_PAYLOAD_TOO_LARGE, HTTP_PAYLOAD_TOO_LARGE_MESSAGE,
 };
 
 use crate::{sentry_debug, types::Scheme, ClientOptions, Envelope, Transport};
@@ -33,6 +33,7 @@ pub struct CurlHttpTransport {
 pub struct CurlHttpTransportOptions {
     general_options: TransportOptions,
     client: Option<CurlClient>,
+    channel_capacity: usize,
 }
 
 impl CurlHttpTransport {
@@ -86,6 +87,7 @@ impl CurlHttpTransport {
                     ..
                 },
             client,
+            channel_capacity,
         } = options;
 
         let client = client.unwrap_or_else(CurlClient::new);
@@ -222,6 +224,7 @@ impl CurlHttpTransport {
 
         let thread = TransportThreadOptions::new(send_fn)
             .with_client_report_recorder(client_report_recorder)
+            .with_channel_capacity(channel_capacity)
             .spawn_thread();
         Self { thread }
     }
@@ -236,7 +239,7 @@ impl Transport for CurlHttpTransport {
     }
 
     fn shutdown(&self, timeout: Duration) -> bool {
-        self.flush(timeout)
+        self.thread.shutdown(timeout)
     }
 }
 
@@ -246,6 +249,7 @@ impl From<TransportOptions> for CurlHttpTransportOptions {
         Self {
             general_options: value,
             client: None,
+            channel_capacity: DEFAULT_CHANNEL_CAPACITY,
         }
     }
 }
@@ -256,6 +260,21 @@ impl CurlHttpTransportOptions {
     pub fn with_client(self, client: CurlClient) -> Self {
         let client = Some(client);
         Self { client, ..self }
+    }
+
+    /// Set the capacity of the channel that queues envelopes for the background
+    /// transport thread (default: 30).
+    ///
+    /// A capacity of `0` creates a rendezvous channel: an envelope is accepted
+    /// only when the transport thread is currently waiting on the receiver,
+    /// otherwise it is dropped. A higher capacity reduces the chance of dropped
+    /// events in high-throughput scenarios at the cost of memory.
+    #[inline]
+    pub fn with_channel_capacity(self, channel_capacity: usize) -> Self {
+        Self {
+            channel_capacity,
+            ..self
+        }
     }
 
     /// Create a [`CurlHttpTransport`] using these options.
--- a/sentry/src/transports/reqwest.rs
+++ b/sentry/src/transports/reqwest.rs
@@ -6,7 +6,7 @@ use sentry_core::TransportOptions;
 
 use super::{
     tokio_thread::{TransportThread, TransportThreadOptions},
-    RateLimiter, HTTP_PAYLOAD_TOO_LARGE, HTTP_PAYLOAD_TOO_LARGE_MESSAGE,
+    RateLimiter, DEFAULT_CHANNEL_CAPACITY, HTTP_PAYLOAD_TOO_LARGE, HTTP_PAYLOAD_TOO_LARGE_MESSAGE,
 };
 
 use crate::{sentry_debug, ClientOptions, Envelope, Transport};
@@ -34,6 +34,7 @@ pub struct ReqwestHttpTransport {
 pub struct ReqwestHttpTransportOptions {
     general_options: TransportOptions,
     client: Option<ReqwestClient>,
+    channel_capacity: usize,
 }
 
 impl ReqwestHttpTransport {
@@ -87,6 +88,7 @@ impl ReqwestHttpTransport {
                     ..
                 },
             client,
+            channel_capacity,
         } = options;
 
         let client = client.unwrap_or_else(|| {
@@ -192,6 +194,7 @@ impl ReqwestHttpTransport {
 
         let thread = TransportThreadOptions::new(send_fn)
             .with_client_report_recorder(client_report_recorder)
+            .with_channel_capacity(channel_capacity)
             .spawn_thread();
         Self { thread }
     }
@@ -206,7 +209,7 @@ impl Transport for ReqwestHttpTransport {
     }
 
     fn shutdown(&self, timeout: Duration) -> bool {
-        self.flush(timeout)
+        self.thread.shutdown(timeout)
     }
 }
 
@@ -216,6 +219,7 @@ impl From<TransportOptions> for ReqwestHttpTransportOptions {
         Self {
             general_options: value,
             client: None,
+            channel_capacity: DEFAULT_CHANNEL_CAPACITY,
         }
     }
 }
@@ -226,6 +230,21 @@ impl ReqwestHttpTransportOptions {
     pub fn with_client(self, client: ReqwestClient) -> Self {
         let client = Some(client);
         Self { client, ..self }
+    }
+
+    /// Set the capacity of the channel that queues envelopes for the background
+    /// transport thread (default: 30).
+    ///
+    /// A capacity of `0` creates a rendezvous channel: an envelope is accepted
+    /// only when the transport thread is currently waiting on the receiver,
+    /// otherwise it is dropped. A higher capacity reduces the chance of dropped
+    /// events in high-throughput scenarios at the cost of memory.
+    #[inline]
+    pub fn with_channel_capacity(self, channel_capacity: usize) -> Self {
+        Self {
+            channel_capacity,
+            ..self
+        }
     }
 
     /// Create a [`ReqwestHttpTransport`] using these options.
--- a/sentry/src/transports/reqwest.rs
+++ b/sentry/src/transports/reqwest.rs
@@ -235,10 +235,9 @@ impl ReqwestHttpTransportOptions {
     /// Set the capacity of the channel that queues envelopes for the background
     /// transport thread (default: 30).
     ///
-    /// A capacity of `0` creates a rendezvous channel: an envelope is accepted
-    /// only when the transport thread is currently waiting on the receiver,
-    /// otherwise it is dropped. A higher capacity reduces the chance of dropped
-    /// events in high-throughput scenarios at the cost of memory.
+    /// A capacity of `0` is clamped to `1`. A higher capacity reduces the
+    /// chance of dropped events in high-throughput scenarios at the cost of
+    /// memory.
     #[inline]
     pub fn with_channel_capacity(self, channel_capacity: usize) -> Self {
         Self {
--- a/sentry/src/transports/thread.rs
+++ b/sentry/src/transports/thread.rs
@@ -1,31 +1,28 @@
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::mpsc::{sync_channel, SyncSender, TrySendError};
 use std::sync::Arc;
 use std::thread::{self, JoinHandle};
 use std::time::Duration;
 
+use crossbeam_channel::{bounded, select_biased, unbounded, Sender, TrySendError};
 use sentry_core::client_report::{Reason as ClientReportReason, Recorder as ClientReportRecorder};
 
 use super::ratelimit::{RateLimiter, RateLimitingCategory};
+use super::{normalize_channel_capacity, DEFAULT_CHANNEL_CAPACITY, DROP_FLUSH_TIMEOUT};
 #[cfg(doc)]
 use super::{StdTransportThread, StdTransportThreadOptions}; // so we can use pub re-exports in docs
 use crate::{sentry_debug, Envelope};
 
-#[expect(
-    clippy::large_enum_variant,
-    reason = "In normal usage this is usually SendEnvelope, the other variants are only used when \
-    the user manually calls transport.flush() or when the transport is shut down."
-)]
-enum Task {
-    SendEnvelope(Envelope),
-    Flush(SyncSender<()>),
+enum ControlTask {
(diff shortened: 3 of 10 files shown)


szokeasaurusrex:
Hi @mvanhorn, thanks for the contribution! Before I proceed to a full review, please address the CI failures and the AI review agent comments. If any of the AI review comments are inaccurate, please comment on them and mark as resolved. Thanks!

mvanhorn:
Addressed in 9c8f904:
- Added `transport_channel_capacity` to the manual Debug impl
- Clamped capacity to min 1 to prevent rendezvous channel (`try_send` would silently drop)
- Ran `cargo fmt`

szokeasaurusrex (changes requested):
Thanks again for the contribution. I think we should revise the public API shape to avoid public API breakages before merging this change.

The current implementation introduces public API breakage by:

- adding a new public field to `ClientOptions`
- changing the signatures of the public transport thread constructors

I would like to avoid those breakages here and instead expose this as additive API:

- add additive `with_capacity(...)` constructors on the public transport thread types, i.e. `StdTransportThread::with_capacity(send, channel_capacity)` and `TokioTransportThread::with_capacity(send, channel_capacity)`
- keep the existing `new(...)` signatures unchanged, but make those constructors delegate to `with_capacity(..., 30)`
- add transport-specific constructors such as `ReqwestHttpTransport::with_channel_capacity(options, channel_capacity)` (and similarly for `curl` and `ureq`), which would use those new `with_capacity(...)` thread constructors internally
- document use through `ClientOptions.transport`

That still gives users a way to override the transport queue capacity via the existing `TransportFactory` mechanism, without changing `ClientOptions` yet.

For example, initializing the SDK with a custom capacity could look like this:

```rust
let opts = ClientOptions {
    transport: Some(Arc::new(move |opts| {
        Arc::new(ReqwestHttpTransport::with_channel_capacity(opts, 256))
    })),
    ..Default::default()
};
```

If you are open to it, please refactor the PR in that direction. If not, let me know and I can take it over as a follow-up.

mvanhorn:
Thanks for the direction @szokeasaurusrex. Refactored to the additive shape you described in fbff3ea:

- `TransportThread::new(send)` restored to the original signature. New `TransportThread::with_capacity(send, channel_capacity)` is the entry point for custom capacity (clamped to a minimum of 1 to avoid the rendezvous channel). Same pattern in `tokio_thread::TransportThread`.
- `ReqwestHttpTransport::with_channel_capacity(options, channel_capacity)` added. `new(options)` / `with_client(options, client)` keep the default capacity of 30. Same for `CurlHttpTransport` and `UreqHttpTransport` (`with_agent` likewise unchanged).
- Removed `transport_channel_capacity` from `ClientOptions` (field, `Default`, and the manual `Debug` impl).

Your example works unchanged:

```rust
let opts = ClientOptions {
    transport: Some(Arc::new(move |opts| {
        Arc::new(ReqwestHttpTransport::with_channel_capacity(opts, 256))
    })),
    ..Default::default()
};
```

`cargo check --workspace --all-features`, `cargo fmt --all`, and `cargo clippy --workspace --all-features` all pass.

sentry/src/transports/curl.rs
```diff
@@ -18,15 +18,28 @@ pub struct CurlHttpTransport {
...
     /// Creates a new Transport.
     pub fn new(options: &ClientOptions) -> Self {
-        Self::new_internal(options, None)
+        Self::new_internal(options, None, 30)
     }
 
     /// Creates a new Transport that uses the specified [`CurlClient`].
     pub fn with_client(options: &ClientOptions, client: CurlClient) -> Self {
-        Self::new_internal(options, Some(client))
+        Self::new_internal(options, Some(client), 30)
```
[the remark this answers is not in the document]
mvanhorn: This finding predates the additive-API refactor in fbff3ea. Per @szokeasaurusrex's review (https://github.com/getsentry/sentry-rust/pull/1040#pullrequestreview-4105775265), the field was intentionally NOT added to `ClientOptions` to avoid a public API breakage. Custom capacity is configured via the additive `TransportThread::with_capacity` and transport `with_channel_capacity` constructors. The default `new(options)` correctly uses 30 because that's the documented default.

sentry/src/transports/reqwest.rs
```diff
@@ -21,15 +21,28 @@ pub struct ReqwestHttpTransport {
 impl ReqwestHttpTransport {
     /// Creates a new Transport.
     pub fn new(options: &ClientOptions) -> Self {
-        Self::new_internal(options, None)
+        Self::new_internal(options, None, 30)
```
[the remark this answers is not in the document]
mvanhorn: This finding is analyzing against the original design. The final shape per maintainer review (https://github.com/getsentry/sentry-rust/pull/1040#pullrequestreview-4105775265) is additive: default `new()` keeps the hardcoded 30 (the documented default); configurable capacity goes through `with_channel_capacity(options, N)`. Not a bug - it's the requested API shape.

mvanhorn:
@szokeasaurusrex - the refactor in fbff3ea matches the additive-API shape you requested. I've also replied to and resolved the two stale bot comments from Apr 15, which were analyzing against the original design. Verified locally: `cargo build --features reqwest` succeeds and the sentry crate tests pass. Ready for your full review whenever you have time.

szokeasaurusrex:
@mvanhorn I have this PR on my list of things to review. I am busy with another project at the moment but will try to review this within the next week or so

mvanhorn:
Thanks @szokeasaurusrex - no rush, whenever you get a chance.

sentry/src/transports/tokio_thread.rs
```diff
@@ -26,14 +26,29 @@ pub struct TransportThread {
...
+        Self::with_capacity(send, 30)
+    }
+
+    /// Spawn a new background thread with a custom channel capacity.
+    ///
+    /// The channel capacity bounds how many envelopes may be queued before
+    /// `send` blocks. `channel_capacity` is clamped to a minimum of 1 to
+    /// avoid a rendezvous channel, which would silently drop envelopes under
+    /// `try_send`.
+    pub fn with_capacity<SendFn, SendFuture>(mut send: SendFn, channel_capacity: usize) -> Self
```
szokeasaurusrex: m: Let's make this `pub(crate)` for now to limit public API surface. If folks want to have this as a public API in the future, we can expose it at that time.

```suggestion
    pub(crate) fn with_capacity<SendFn, SendFuture>(mut send: SendFn, channel_capacity: usize) -> Self
```

sentry/src/transports/thread.rs
```diff
@@ -26,12 +26,25 @@ pub struct TransportThread {
...
+        Self::with_capacity(send, 30)
+    }
+
+    /// Spawn a new background thread with a custom channel capacity.
+    ///
+    /// The channel capacity bounds how many envelopes may be queued before
+    /// `send` blocks. `channel_capacity` is clamped to a minimum of 1 to
+    /// avoid a rendezvous channel, which would silently drop envelopes under
+    /// `try_send`.
+    pub fn with_capacity<SendFn>(mut send: SendFn, channel_capacity: usize) -> Self
```
szokeasaurusrex: m: Let's make this `pub(crate)` for now to limit public API surface. If folks want to have this as a public API in the future, we can expose it at that time.

```suggestion
    pub(crate) fn with_capacity<SendFn>(mut send: SendFn, channel_capacity: usize) -> Self
```

sentry/src/transports/curl.rs
```diff
@@ -18,15 +18,28 @@ pub struct CurlHttpTransport {
 impl CurlHttpTransport {
     /// Creates a new Transport.
     pub fn new(options: &ClientOptions) -> Self {
-        Self::new_internal(options, None)
+        Self::new_internal(options, None, 30)
```
szokeasaurusrex: m: As we use the number `30` as the default channel capacity in all the transports, we should extract it to a constant that we reuse in all of them.

szokeasaurusrex (commented):
Thanks for your patience for my review. I added some small comments; this looks pretty good overall though!

mvanhorn:
Addressed in 6d1c9ff:
- Both `TransportThread::with_capacity` are now `pub(crate)` per your suggestion.
- Extracted the `30` default to `pub(crate) const DEFAULT_CHANNEL_CAPACITY: usize = 30` in `transports/mod.rs` and reused it across curl, reqwest, and ureq transports (plus both thread modules' `new` defaults). Open to moving the const to `transports/thread.rs` instead if that reads cleaner -- happy to follow your preference.

Verified `cargo fmt --check`, `cargo build --features reqwest`, and `cargo build --features curl` locally before pushing.

sentry/src/transports/thread.rs
```diff
@@ -26,12 +27,25 @@ pub struct TransportThread {
...
+    ///
+    /// The channel capacity bounds how many envelopes may be queued before
+    /// `send` blocks. `channel_capacity` is clamped to a minimum of 1 to
+    /// avoid a rendezvous channel, which would silently drop envelopes under
+    /// `try_send`.
+    pub(crate) fn with_capacity<SendFn>(mut send: SendFn, channel_capacity: usize) -> Self
+    where
+        SendFn: FnMut(Envelope, &mut RateLimiter) + Send + 'static,
+    {
+        let (sender, receiver) = sync_channel(channel_capacity.max(1));
```
szokeasaurusrex: I think we should honor `0` here instead of clamping it. This is an advanced, opt-in transport API, and `sync_channel(0)` has defined rendezvous/no-buffer semantics. A capacity of `1` can still drop most events under bursts; it is not a general safeguard, only a different buffering policy. If someone explicitly passes `0`, treating that as “no queued buffering” seems reasonable.

I tested this end-to-end with the clamp removed and `with_channel_capacity(..., 0)`: sending 10 rapid events accepted 1 and dropped 9. That matches the channel semantics: zero capacity does not necessarily drop everything; it accepts an envelope when `try_send` happens while the transport thread is already waiting on the receiver. If we keep support for `0`, the doc comment should describe that behavior rather than saying it would silently drop envelopes generally.

sentry/src/transports/tokio_thread.rs
```diff
@@ -26,14 +27,32 @@ pub struct TransportThread {
...
+    pub(crate) fn with_capacity<SendFn, SendFuture>(
+        mut send: SendFn,
+        channel_capacity: usize,
+    ) -> Self
+    where
+        SendFn: FnMut(Envelope, RateLimiter) -> SendFuture + Send + 'static,
+        // NOTE: returning RateLimiter here, otherwise we are in borrow hell
+        SendFuture: std::future::Future<Output = RateLimiter>,
+    {
+        let (sender, receiver) = sync_channel(channel_capacity.max(1));
```
szokeasaurusrex: Same concept applies here; I tested both transport thread variants with the clamp removed, and both accepted 1 of 10 rapid events with capacity `0` rather than dropping everything.

szokeasaurusrex (commented):
Hey, thanks for addressing those! I just have one more thought about the clamping, then I think this is good to merge

mvanhorn:
Done in c819ce9 - dropped the `.max(1)` clamp in both `transports/thread.rs` and `transports/tokio_thread.rs`, and updated the doc comments to describe the rendezvous semantics rather than implying capacity 0 drops everything. Local `cargo build --features reqwest` and `cargo test -p sentry --features reqwest --lib` pass; `cargo fmt --check` clean.

mvanhorn:
Rebased onto master and resolved the conflict with the new transport-options refactor. Two things:

1. The clamp is gone (c819ce9) - capacity 0 now creates a rendezvous channel as you asked, no `.max(1)`.

2. To fit your new options-builder pattern, I moved capacity config onto the transport options builders: `CurlHttpTransportOptions::with_channel_capacity(n)` (and the Reqwest/Ureq equivalents), backed by `TransportThreadOptions::with_channel_capacity(n)`. That drops the old transport-level `with_channel_capacity(&ClientOptions, n)` convenience in favor of the builder, which felt more consistent now that `new()`/`with_client()` are deprecated - happy to add a convenience method back if you'd rather.

Verified locally with `--all-features`: build, all tests, clippy `-D clippy::all`, and rustfmt all clean.

mvanhorn:
Pushed 58951b1. Addressed the capacity-0 flush/drop hang the bot flagged: flush() and drop() now use try_send() for their control tasks (envelope enqueue was already try_send), and the sender is wrapped in Option so Drop can take it cleanly, so a rendezvous channel never blocks the caller on flush or shutdown. Added a regression test in both the std and tokio transports asserting flush() returns false instead of blocking on a busy capacity-0 channel. Also made with_channel_capacity pub(crate) in both TransportThreadOptions to keep the public surface limited as you asked. The shared DEFAULT_CHANNEL_CAPACITY constant and honoring 0 landed in the earlier round.

mvanhorn:
Pushed 3cda078. Capacity 0 stays supported — flush and shutdown now go through a dedicated control channel rather than the envelope channel, so they can't deadlock at capacity 0, and flush drains pending envelopes instead of dropping them via `try_send`. Added a changelog entry. (ESP-IDF CI and the ureq-feature clippy expectation I couldn't run locally — the latter looks pre-existing in sentry-core, but flag it if not.)

mvanhorn:
Pushed 780d015: Drop drains pending envelopes before shutting the worker down (no more silent loss when the transport is dropped mid-queue), and the control-channel check no longer delays idle flushes. Transport suite is green locally (6/6) along with fmt.

mvanhorn:
Pushed the fix for the flagged transport channel-capacity handling — pending envelopes are now drained on `Drop` and the idle flush stays responsive (thread.rs / tokio_thread.rs). Verified locally with `cargo fmt`, the exact CI clippy, and the full all-features/all-targets test + doc-test suite. One heads-up: the branch is a fair way behind `master`, so it'll need a rebase before it can merge — I'll get that sorted separately.

mvanhorn:
This is ready for another look whenever you have time. The requested changes went in at 8c14fa8, which landed after the last review pass, and all 16 review threads are resolved.

Re-verified locally on that commit: the timeout-preserving `Drop` fix and its regression tests are present for both transport variants, and rustfmt, an offline build, Clippy, and the focused tests all pass.

No rush, just flagging that the ball is back on your side.

mvanhorn:
Addressed in afa4f65.

The two HIGH Drop findings are the same defect from two angles: Drop sent a Flush and blocked on recv() with no bound. That hangs on a stuck worker, and it also made Client::close's timed shutdown pointless, since dropping right after blocked anyway. Drop now runs a shutdown bounded by DROP_FLUSH_TIMEOUT.

I kept the join when the flush completed, so a healthy worker is still awaited and a worker panic still surfaces rather than being swallowed. Only a timed-out flush skips it, since join() has no timeout and would put the unbounded wait right back. Draining on a normal drop is unchanged, so 780d015 isn't regressed.

On the capacity finding: I'd documented 0 as a deliberate rendezvous / no-buffer back-pressure mode, but the Bugbot point stands. With send using try_send, that reads as back-pressure and behaves as near-total silent loss, and nothing warned. Clamped to a minimum of 1. Happy to reinstate the rendezvous behaviour behind something more explicit if you'd rather have it available.

Both the std and tokio transport threads. 9/9 transport tests pass locally, clippy and fmt clean.

mvanhorn:
Rebased onto master (0.49.1) and squashed to a single commit.

The branch had accumulated 16 commits, most of them fork-sync duplicates of upstream work under different SHAs, which is also why it went conflicting. The transport code is byte-identical to what you last reviewed at afa4f65 - only the history and the CHANGELOG placement changed. My entry now sits under a new `Unreleased` heading above 0.49.1 rather than inside it.

@szokeasaurusrex the clamping thought from your last pass is in, and both Bugbot `Drop` findings are covered by named regression tests that pass on this commit:

- `drop_does_not_wait_for_a_stuck_worker`
- `timed_out_shutdown_does_not_block_drop_or_drain_queued_envelopes`

Full run: 9/9 on `sentry --lib transports`.

szokeasaurusrex:
Thanks @mvanhorn, I have it on my to-do list to review this PR, might take a few days/weeks until I get around to it 

Thank you in advance for your patience 🙏

mvanhorn:
Right on both counts, and ureq was the last file still describing the old behaviour. normalize_channel_capacity does .max(1), so a capacity of 0 never produced a rendezvous channel; curl, reqwest, thread, tokio_thread and the module doc were already corrected earlier in this branch.

6192e0d. cargo check -p sentry --features transport,ureq is clean.

szokeasaurusrex:
Closing in favor of #1311. As `ClientOptions` is `#[non_exhaustive]` starting in `0.49.0`, we can now add the capacity option directly on `ClientOptions` without needing a breaking change.

PR #1026, opened 2026-03-05
feat(sentry-core): Implement trace metric capture batching

szokeasaurusrex:
Basic metrics capture functionality. Follow-up PR will implement the rest.

Stacked on #1022 

Closes #1023
Closes [RUST-168](https://linear.app/getsentry/issue/RUST-168/implement-trace-metric-capture-and-batching-in-sentry-core)

PR #1061, opened 2026-04-15
feat(metrics): Associate trace data with metrics

### LINKED ISSUE #1058: Add trace metric tracing association in sentry-core
Split from [#1024](https://github.com/getsentry/sentry-rust/issues/1024).

Add tracing association behavior in `sentry-core` so emitted metrics include tracing identifiers from the current execution context, aligned with [tracing association](https://develop.sentry.dev/sdk/telemetry/metrics/#tracing-association). Set `trace_id` from the active span when one exists, otherwise from the propagation context, and set `span_id` from the active span when one exists. Focus on the relevant metric preparation and scope application code in `sentry-core/src/client.rs`, `sentry-core/src/scope/real.rs`, and `sentry-core/src/scope/noop.rs`. Add tests for `trace_id` from propagation context when no active span exists, `trace_id` from the active span when a span exists, and `span_id` from the active span when a span exists. Do not include default attribute enrichment, user attribute enrichment, or callback processing in this issue.

If code is reused from [#997](https://github.com/getsentry/sentry-rust/pull/997), include `Co-authored-by: Joris Bayer <joris.bayer@sentry.io>` in commit trailers.

szokeasaurusrex:
Set trace and span id on metrics.

Stacked on #1026

Closes #1058
Closes [RUST-186](https://linear.app/getsentry/issue/RUST-186/add-trace-metric-tracing-association-in-sentry-core)

### DIFF
--- a/sentry-core/src/client.rs
+++ b/sentry-core/src/client.rs
@@ -12,7 +12,7 @@ use crate::protocol::SessionUpdate;
 use rand::random;
 use sentry_types::random_uuid;
 
-#[cfg(feature = "logs")]
+#[cfg(any(feature = "logs", feature = "metrics"))]
 use crate::batcher::Batcher;
 use crate::constants::SDK_INFO;
 use crate::protocol::{ClientSdkInfo, Event};
@@ -24,6 +24,8 @@ use crate::SessionMode;
 use crate::{ClientOptions, Envelope, Hub, Integration, Scope, Transport};
 #[cfg(feature = "logs")]
 use sentry_types::protocol::v7::Context;
+#[cfg(feature = "metrics")]
+use sentry_types::protocol::v7::Metric;
 #[cfg(feature = "logs")]
 use sentry_types::protocol::v7::{Log, LogAttribute};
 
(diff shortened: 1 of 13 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -399,6 +401,17 @@ impl Scope {
...
 
+    /// Applies the contained scoped data to a trace metric, setting the `trace_id` and `span_id`.
+    #[cfg(feature = "metrics")]
+    pub fn apply_to_metric(&self, metric: &mut Metric) {
+        metric.trace_id = self
+            .get_span()
+            .map(|span| span.get_trace_context().trace_id)
+            .unwrap_or(self.propagation_context.trace_id);
+
+        metric.span_id = self.get_span().map(|span| span.span_id());
```
lcian, line 412: We should fall back on the propagation context to get the span ID, as it seems the Python implementation does it that way: https://github.com/getsentry/sentry-python/blob/5d26c2f60e289c89ab260ee25607a3bea5e174c9/sentry_sdk/scope.py#L1828.

I saw that logs do the same and only consider `self.span`, so we should probably fix that too.

sentry-core/src/client.rs
```diff
@@ -524,16 +524,25 @@ impl Client {
...
+                .as_ref()
+            {
+                batcher.enqueue(metric);
+            }
         }
     }
+
+    /// Prepares a metric to be sent, setting trace association data from the scope.
+    #[cfg(feature = "metrics")]
+    fn prepare_metric(&self, mut metric: Metric, scope: &Scope) -> Option<Metric> {
```
lcian, line 542: This could also return just `Metric`. If we need to have `before_send_metric` (not sure if that exists) then I guess `Option<Metric>` makes sense though. It's an internal API anyway so it doesn't really matter.

lcian (approved):
LGTM other than the comment about `span_id`.

szokeasaurusrex:
Marking as "draft" until I have addressed @lcian's comments

szokeasaurusrex:
folded into #1073

PR #1062, opened 2026-04-16
feat(metrics): Attach default SDK attributes to metrics

szokeasaurusrex:
Set these metric attributes by default:
- `sentry.environment`
- `sentry.release`
- `sentry.sdk.name`
- `sentry.sdk.version`
- `server.address`

Preserve explicitly set metric attributes when applying those defaults.

Stacked on #1073

Co-authored-by: Joris Bayer <joris.bayer@sentry.io>
Closes #1059
Closes [RUST-187](https://linear.app/getsentry/issue/RUST-187/add-trace-metric-default-attribute-enrichment-in-sentry-core)

PR #1063, opened 2026-04-16
feat(metrics): Add metric user attributes from scope

### LINKED ISSUE #1060: Add trace metric user attribute enrichment gated by send_default_pii in sentry-core
Split from [#1024](https://github.com/getsentry/sentry-rust/issues/1024).

Add user attribute enrichment behavior in `sentry-core` so emitted metrics include selected user attributes only when `send_default_pii` is enabled, aligned with [user attributes](https://develop.sentry.dev/sdk/telemetry/metrics/#user-attributes). Attach `user.id`, `user.name`, and `user.email` only when `send_default_pii` is `true`, and preserve explicitly set metric attributes instead of overwriting them. Focus on the relevant metric scope application code in `sentry-core/src/scope/real.rs`, `sentry-core/src/scope/noop.rs`, and `sentry-core/src/client.rs`. Add tests for user attributes omitted when `send_default_pii` is `false`, user attributes attached when `send_default_pii` is `true`, and explicitly set `user.*` metric attributes not overwritten by scope user data. Do not include tracing association, default attribute enrichment, or callback processing in this issue.

If code is reused from [#997](https://github.com/getsentry/sentry-rust/pull/997), include `Co-authored-by: Joris Bayer <joris.bayer@sentry.io>` in commit trailers.

szokeasaurusrex:
Add the metric user attribute portion of the earlier enrichment work.

Pass send_default_pii into metric scope application and attach user.id, user.name, and user.email from the scope when that option is enabled. Preserve explicitly set `user.*` metric attributes instead of overwriting them.

Stacked on #1062

Closes #1060
Closes [RUST-188](https://linear.app/getsentry/issue/RUST-188/add-trace-metric-user-attribute-enrichment-gated-by-send-default-pii)

### DIFF
--- a/sentry-core/src/scope/real.rs
+++ b/sentry-core/src/scope/real.rs
@@ -9,13 +9,15 @@ use std::sync::{Arc, PoisonError, RwLock};
 #[cfg(feature = "metrics")]
 use crate::metrics::{IntoProtocolMetric, MetricTraceInfo};
 use crate::performance::TransactionOrSpan;
+#[cfg(feature = "logs")]
+use crate::protocol::Log;
+#[cfg(any(feature = "logs", feature = "metrics"))]
+use crate::protocol::LogAttribute;
 #[cfg(feature = "metrics")]
 use crate::protocol::Metric;
 use crate::protocol::{
     Attachment, Breadcrumb, Context, Event, Level, TraceContext, Transaction, User, Value,
 };
-#[cfg(feature = "logs")]
-use crate::protocol::{Log, LogAttribute};
 #[cfg(feature = "release-health")]
 use crate::session::Session;
 use crate::{Client, SentryTrace, TraceHeader, TraceHeadersIter};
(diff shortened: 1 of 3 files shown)


sentry-core/src/scope/real.rs
```diff
@@ -457,3 +469,48 @@ impl Scope {
...
+        V: Into<Value>,
+    {
+        self.attributes
+            .insert(key.into(), LogAttribute(value.into()));
+    }
+
+    fn attribute(&self, key: &str) -> Option<&LogAttribute> {
+        self.attributes.get(key)
+    }
+}
```
lcian, line 516: You could just inline this code instead of putting it in a separate extension trait. But I guess it's a matter of preference, so feel free to keep it if you like it more this way.

lcian (approved):
LGTM

sentry-core/src/scope/real.rs
szokeasaurusrex: I find it quite a bit more ergonomic tbh, especially because this extension trait also handles type conversions for me.

Why do you think it would be better to inline? Just personal preference, or is there some technical tradeoff that I am failing to consider?

lcian: Just personal preference 👍

sentry-core/src/scope/real.rs
```diff
@@ -401,15 +403,25 @@ impl Scope {
...
         metric.span_id = self.get_span().map(|span| span.span_id());
+
+        let should_add_user_attributes = send_default_pii && !metric.has_any_user_attributes();
+
+        if let Some(user) = should_add_user_attributes
+            .then_some(self.user.as_deref())
+            .flatten()
+        {
+            metric.apply_user_attributes(user);
+        }
```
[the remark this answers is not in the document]
szokeasaurusrex: Intentional

PR #1064, opened 2026-04-16
feat(metrics): Add before_send_metric callback

### LINKED ISSUE #1025: Add before_send_metric callback processing in sentry-core
Add `before_send_metric` to `ClientOptions` and execute it in the metric prepare path so metrics can be modified or dropped before batching. Callback execution should happen after scope/default attribute enrichment and before enqueue.

### Scope
Apply the relevant logic from [#997](https://github.com/getsentry/sentry-rust/pull/997) where useful, focused on [`sentry-core/src/clientoptions.rs`](https://github.com/getsentry/sentry-rust/blob/343a63d025276b0474882e2dbc2ff27bc5400743/sentry-core/src/clientoptions.rs) and [`sentry-core/src/client.rs`](https://github.com/getsentry/sentry-rust/blob/343a63d025276b0474882e2dbc2ff27bc5400743/sentry-core/src/client.rs), aligned with [processing pipeline step 4](https://develop.sentry.dev/sdk/telemetry/metrics/#processing-pipeline) and [initialization options](https://develop.sentry.dev/sdk/telemetry/metrics/#initialization-options). Include tests for callback drop and callback mutation. If code is reused from #997, include `Co-authored-by: Joris Bayer <joris.bayer@sentry.io>` in commit trailers.

szokeasaurusrex:
Run metrics through `ClientOptions::before_send_metric` after scope/default attribute enrichment and before enqueue so applications can filter or mutate metrics before batching.

Stacked on #1063. 

 Closes #1025
Closes [RUST-170](https://linear.app/getsentry/issue/RUST-170/add-before-send-metric-callback-processing-in-sentry-core)

### DIFF
--- a/sentry-core/tests/metrics.rs
+++ b/sentry-core/tests/metrics.rs
@@ -1,6 +1,7 @@
 #![cfg(all(feature = "test", feature = "metrics"))]
 
 use std::collections::HashSet;
+use std::sync::Arc;
 
 use anyhow::{Context, Result};
 
@@ -8,7 +9,7 @@ use sentry::protocol::{MetricType, Unit, Value};
 use sentry_core::protocol::{EnvelopeItem, ItemContainer};
 use sentry_core::{metrics, test};
 use sentry_core::{ClientOptions, TransactionContext};
-use sentry_types::protocol::v7::{Envelope, Metric, User};
+use sentry_types::protocol::v7::{Envelope, LogAttribute, Metric, User};
 
 /// Test that metrics are sent when metrics are enabled.
 #[test]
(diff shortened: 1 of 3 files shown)


sentry-core/tests/metrics.rs
```diff
@@ -494,6 +495,45 @@ fn metric_user_attributes_do_not_overwrite_explicit() {
...
+        })),
+        ..Default::default()
+    };
+
+    let envelopes = test::with_captured_envelopes_options(|| capture_test_metric("test"), options);
+    let metric = extract_single_metric(envelopes).expect("expected a single-metric envelope");
+
+    assert_eq!(
+        metric.attributes.get("added_by_callback"),
+        Some(&LogAttribute(Value::from("yes"))),
```
lcian, line 533: Did we intend to call this just `Attribute` (could either be a new type or just a type alias)? I don't remember what was the outcome of that discussion (if any).

lcian (approved):

sentry-core/tests/metrics.rs
szokeasaurusrex: Yeah, I think it makes sense to make that change. Likely we should do it as a type alias (rename `LogAttribute` to `Attribute`, and make `LogAttribute` an alias to `Attribute`). However, I believe we can also do this later, should be a backwards and forwards-compatible change as far as I am aware

PR #1072, opened 2026-04-21
feat(types): Add `Unit` enum

### LINKED ISSUE #1070: Add Unit constants/enum
Per the [metrics spec](<https://develop.sentry.dev/sdk/telemetry/metrics/#metrics-module>), we should offer some constants for the [units we support](<https://develop.sentry.dev/sdk/foundations/state-management/scopes/attributes/#units>), but do this in a way to avoid restricting potential values. In Rust, likely we can do an enum with a fallback option

szokeasaurusrex:
The `Unit` enum will be used in the `Metric` type to store metric units.

Closes #1070
Closes [RUST-192](https://linear.app/getsentry/issue/RUST-192/add-unit-constantsenum)

### DIFF
--- a/sentry-types/src/protocol/mod.rs
+++ b/sentry-types/src/protocol/mod.rs
@@ -17,3 +17,4 @@ mod attachment;
 mod envelope;
 mod monitor;
 mod session;
+mod unit;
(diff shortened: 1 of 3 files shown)


lcian (approved):
lgtm

giortzisg (approved):

PR #1074, opened 2026-04-22
ref(sentry-panic): `PanicInfo` → `PanicHookInfo`

### LINKED ISSUE #862: Switch to `PanicHookInfo` in `sentry-panic`
### Description

The MSRV for this crate is 1.81, we should use `PanicHookInfo` instead of `PanicInfo` which is deprecated.

mvanhorn:
Closes getsentry/sentry-rust#862.

The workspace MSRV is 1.88 (`Cargo.toml`), so `std::panic::PanicHookInfo` (stabilized in 1.81) is always available. This removes the six `#[allow(deprecated)]` annotations that were silencing the `PanicInfo` deprecation warning and uses `PanicHookInfo` directly in `sentry-panic`.

## Source compatibility

`std` keeps `PanicInfo` as a type alias for `PanicHookInfo`:

```rust
// library/std/src/panic.rs
pub type PanicInfo<'a> = PanicHookInfo<'a>;
```

So downstream extractor closures that still use `&PanicInfo<'_>` continue to match the new `&PanicHookInfo<'_>` signatures on `add_extractor`, `event_from_panic_info`, and `message_from_panic_info`.

## Scope

Only `sentry-panic/src/lib.rs` is touched. No other crates reference `PanicInfo`.

Verified with `cargo build --all`, `cargo clippy -p sentry-panic`, `cargo fmt -p sentry-panic --check`, and `cargo test -p sentry-panic`.

### DIFF
--- a/sentry-panic/src/lib.rs
+++ b/sentry-panic/src/lib.rs
@@ -8,7 +8,7 @@
 //! # Configuration
 //!
 //! The panic integration can be configured with an additional extractor, which
-//! might optionally create a sentry `Event` out of a `PanicInfo`.
+//! might optionally create a sentry `Event` out of a `PanicHookInfo`.
 //!
 //! ```
 //! let integration = sentry_panic::PanicIntegration::default().add_extractor(|info| None);
@@ -19,8 +19,7 @@
 #![warn(missing_docs)]
 #![deny(unsafe_code)]
 
-#[allow(deprecated)] // `PanicHookInfo` is only available in Rust 1.81+.
-use std::panic::{self, PanicInfo};
+use std::panic::{self, PanicHookInfo};
 use std::sync::Once;
 
 use sentry_backtrace::current_stacktrace;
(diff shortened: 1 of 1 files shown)


szokeasaurusrex:
Sorry about the bot @mvanhorn, this PR indeed looks valid. I'll review when I have time.

In general, we like issues to be opened before PRs, but for simple stuff like this change, I think the issue is not necessarily needed.

Thanks again for your contribution!

szokeasaurusrex (changes requested):
Ok, just did a quick review; I am not sure whether we should do this at all, since it would break our public API. 

Unless using `PanicInfo` is causing you some inconvenience, I would suggest we just keep the code as it is for now. Otherwise, perhaps it makes sense to open an issue to discuss the right approach after all 😅 

If using `PanicInfo` is causing trouble for you, please open an issue to explain what the concrete problem is, and we can discuss a potential solution there.

mvanhorn:
Thanks for the quick look @szokeasaurusrex. The PR body covers this but may have buried it - `std` defines `pub type PanicInfo<'a> = PanicHookInfo<'a>` (type alias, not a separate type), so downstream extractor closures using `&PanicInfo<'_>` continue to match the new `&PanicHookInfo<'_>` signatures on `add_extractor` / `event_from_panic_info` / `message_from_panic_info`. No break at the public-API boundary, just silences the six `#[allow(deprecated)]` annotations.

Happy to close if you'd still rather leave things as-is - the deprecation warning is cosmetic and the allow attrs aren't hurting anything. Just didn't want you to have to merge another six `#[allow(deprecated)]` reminders down the line.

szokeasaurusrex (approved):
Thanks for the explanation and for your patience. This looks good now!

PR #1101, opened 2026-04-29
fix: do not pull in `aws-lc-rs` by default

thomaseizinger:
### Description

Activating the `rustls` feature of `reqwest` pulls in `aws-lc-rs` by default as the crypto provider. To allow people to make their own decision about which crypto provider to use with `rustls`, we should depend on the `rustls-no-provider` feature flag instead.


szokeasaurusrex:
Superseded by #1103

PR #1106, opened 2026-05-04
fix(metrics): Enable metrics by default

### LINKED ISSUE #1105: Default `enable_metrics` to true
This is the default used in other SDKs, we should align it in the Rust SDK

szokeasaurusrex:
Align runtime metrics capture with other SDKs while preserving explicit opt-out via ClientOptions::enable_metrics.

Closes #1105
Closes [RUST-207](https://linear.app/getsentry/issue/RUST-207/default-enable-metrics-to-true)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


lcian (approved):
lgtm

PR #1108, opened 2026-05-04
docs(metrics): Improve Rust metrics API docs

szokeasaurusrex:
Add more details and links to the [Rust metrics API docs](https://docs.rs/sentry/latest/sentry/).

I expect many SDK users will find the API docs more easily than the [product docs page](https://docs.sentry.io/platforms/rust/metrics/).


lcian (approved):
lgtm

PR #1109, opened 2026-05-04
fix(types): Serialize attachment headers as JSON

szokeasaurusrex:
Use serde_json to write attachment item headers instead of formatting the JSON string manually. This keeps string fields encoded consistently with the rest of envelope serialization.


giortzisg (approved):
LGTM

PR #1125, opened 2026-05-07
ref(hub): Remove boolean to track if we are using process hub

szokeasaurusrex:
Instead, it is simpler just to set the `THREAD_HUB` to the `PROCESS_HUB` (or, more specifically, an Arc-clone of it).

Stacked on #1124

### DIFF
--- a/sentry-core/src/hub_impl.rs
+++ b/sentry-core/src/hub_impl.rs
@@ -1,4 +1,4 @@
-use std::cell::{Cell, RefCell};
+use std::cell::RefCell;
 use std::marker::PhantomData;
 use std::sync::{Arc, LazyLock, MutexGuard, PoisonError, RwLock};
 use std::thread;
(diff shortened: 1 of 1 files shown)


sentry-core/src/hub_impl.rs
```diff
@@ -164,16 +165,12 @@ impl Hub {
     where
         F: FnOnce(&Arc<Hub>) -> R,
     {
-        THREAD_HUB.with(|(hub, is_process_hub)| {
-            if is_process_hub.get() {
-                f(&PROCESS_HUB.hub)
```
szokeasaurusrex, line 169: As I understand it, this was the sole purpose of `is_process_hub`: so that when we are on the thread that created the hub, the current hub is the `PROCESS_HUB`. 

The changes here eliminate the need to track `is_process_hub` separately, since we just make the thread hub an arc-clone of the process hub on the thread that creates the process hub.

lcian (approved):
lgtm

PR #1126, opened 2026-05-08
chore(hub): Deprecate `Hub::with`

szokeasaurusrex:
Following the safety improvements in #1123, `Hub::with` no longer offers any performance benefit over using `Hub::current`. `Hub::current` provides a simpler API, and also returns an owned `Arc<Hub>` rather than providing access to a `&Arc<Hub>` via a callback (necessitating a second clone for anyone needing an owned `Arc<Hub>` in that callback).

As `Hub::current` is the arguably better API, let's deprecate `Hub::with` and encourage folks to migrate.

Stacked on #1125

### DIFF
--- a/sentry-core/src/hub.rs
+++ b/sentry-core/src/hub.rs
@@ -65,13 +65,13 @@ impl Hub {
     {
         use_without_client!(f);
         with_client_impl! {{
-            Hub::with(|hub| {
-                if hub.is_active_and_usage_safe() {
-                    f(hub)
-                } else {
-                    Default::default()
-                }
-            })
+            let hub = Hub::current();
+            if hub.is_active_and_usage_safe() {
+                f(&hub)
+            } else {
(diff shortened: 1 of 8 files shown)


sentry-core/src/hub.rs
```diff
@@ -65,13 +65,13 @@ impl Hub {
...
-                    Default::default()
-                }
-            })
+            let hub = Hub::current();
+            if hub.is_active_and_usage_safe() {
+                f(&hub)
+            } else {
+                Default::default()
+            }
+
```
szokeasaurusrex, line 74: Eventually, we might want a version of this where the closure takes `Arc` rather than `&Arc`

lcian (approved):
So, to the user we always give out an `Arc::clone`, and internally we use the direct `with_borrow` path which may panic, which we're making sure to use safely.
LGTM

Issue #1002, opened 2026-02-24
Add shared client report aggregator and send_client_reports option


szokeasaurusrex:
Part of #882. Depends on #1001.

Implement a shared, thread-safe client report aggregation primitive (for example keyed by `(reason, category)` with record and drain APIs) and add `ClientOptions::send_client_reports` defaulting to `true`. Wire the option through client initialization so bookkeeping and report emission can be fully disabled when requested.

Issue #1004, opened 2026-02-24
Record transport drops and attach client_report envelopes


szokeasaurusrex:
Part of #882. Depends on #1002.

Record transport-side drops for queue overflow, rate-limit filtering, network failures, and HTTP status handling (`2xx` success, `413` + other non-`429` `4xx/5xx` as `send_error`, no client report for `429`). Attach pending `client_report` items to outgoing envelopes (piggyback only, no recursion) and apply behavior consistently across reqwest, curl, ureq, and embedded transports.

PR #1139, opened 2026-06-08
ref(client): Create `EnvelopeSender` to wrap `Transport`

### LINKED ISSUE #1155: Centralize envelope sending logic in `sentry-core`
Tracking issue for [#1139](https://github.com/getsentry/sentry-rust/pull/1139).

This PR introduces `EnvelopeSender` as a client-side abstraction around transport sending. This refactor supports [#1002](https://github.com/getsentry/sentry-rust/issues/1002) by giving client-report aggregation a single place to attach pending `client_report` items to outgoing envelopes, instead of duplicating that behavior across the client, batchers, and session flusher.

szokeasaurusrex:
Create a new `EnvelopeSender` type, which wraps the client's transport.

`EnvelopeSender` is itself a lightweight wrapper around a `TransportSlot`, which itself is a replacement for `TransportArc` and was also introduced in this change. The `TransportSlot` abstracts the logic for managing whether the transport is present or not, while the `EnvelopeSender` abstracts away the whole transport itself.

**Why the change?** Previously, code in several different places (the client, the batchers, and the session flusher) interacted with the transport directly. This makes it difficult for us to ensure that any preprocessing we want to do on all envelopes before they go to the transport happens in all the places we use the transport.

With this change, we can perform any needed preprocessing in the `EnvelopeSender`'s `send_envelope` implementation. We intend to use this for sending client reports; we will check in `EnvelopeSender`'s `send_envelope` whether there are any pending reports, and attach them to the outgoing envelope if yes.

Contributes to [#1002](https://github.com/getsentry/sentry-rust/issues/1002)
Contributes to [RUST-154](https://linear.app/getsentry/issue/RUST-154/add-shared-client-report-aggregator-and-send-client-reports-option)
Closes [#1155](https://github.com/getsentry/sentry-rust/issues/1155)
Closes [RUST-230](https://linear.app/getsentry/issue/RUST-230)

### DIFF
--- a/sentry-core/src/client/envelope_sender.rs
+++ b/sentry-core/src/client/envelope_sender.rs
@@ -1,0 +1,164 @@
+//! Contains abstractions for sending envelopes.
+//!
+//! The most important type here is the [`EnvelopeSender`] struct, which wraps a [`Transport`] and
+//! centralizes envelope sending logic. All code in this crate should send envelopes via the
+//! [`EnvelopeSender`], not by using the [`Transport`] directly.
+
+use std::sync::Arc;
+use std::time::Duration;
+
+use self::slot::TransportSlot;
+use crate::{Envelope, Transport};
+
+/// Sends envelopes through the client's transport.
+///
+/// Cloning this sender has `Arc`-like semantics: clones share the same transport
+/// slot and send to the same underlying transport until it is shut down.
+///
+/// The [`Default`] implementation creates an [`EnvelopeSender`] without an underlying transport,
+/// effectively rendering calls a no-op.
+#[derive(Clone, Default)]
+pub(crate) struct EnvelopeSender {
+    transport_slot: TransportSlot<dyn Transport>,
+}
+
+impl EnvelopeSender {
+    /// Sends an envelope if the transport is still available.
+    pub(crate) fn send_envelope(&self, envelope: Envelope) {
+        self.transport_slot.send_envelope(envelope);
+    }
+
+    /// Creates a sender using the transport returned by the provided builder callback.
+    pub(super) fn new<F>(transport_builder: F) -> Self
+    where
+        F: FnOnce() -> Arc<dyn Transport>,
+    {
+        let transport_slot = TransportSlot::new(transport_builder());
+        Self { transport_slot }
+    }
+
+    /// Flushes the transport if it is still available.
+    pub(super) fn flush(&self, timeout: Duration) -> bool {
+        self.transport_slot.flush(timeout)
+    }
+
+    /// Shuts down and removes the transport if it is still available.
+    pub(super) fn shutdown(&self, timeout: Duration) -> bool {
+        self.transport_slot.shutdown(timeout)
+    }
+
+    pub(super) fn clone_with_new_transport_slot(&self) -> Self {
+        let transport_slot = self.transport_slot.clone_into_new_slot();
+        Self { transport_slot }
+    }
+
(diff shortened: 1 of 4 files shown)


szokeasaurusrex:
@lcian This is ready for review.

I ended up changing the [`EnvelopeSender` constructor](https://github.com/getsentry/sentry-rust/pull/1139/changes#diff-154145199ccd4a927192923aa7300bee2dbe7bef7eecc05587c91211ca22b0e4R31-R38) since yesterday. It now takes a callback which builds the transport. 

In the PR which implements #1004, I'll use the callback to provide the transport constructors with the handle the transport will need to use to record losses.

sentry-core/src/client/envelope_sender.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub(super) fn flush(&self, timeout: Duration) -> bool {
+        self.transport_slot.flush(timeout)
+    }
+
+    /// Shuts down and removes the transport if it is still available.
+    pub(super) fn shutdown(&self, timeout: Duration) -> bool {
+        self.transport_slot.shutdown(timeout)
+    }
+
+    pub(super) fn clone_with_new_transport_slot(&self) -> Self {
```
lcian, line 64: I thought about this again and I don't think we need this.

Say we have two instances of `EnvelopeSender` A and B, produced via `A.clone_with_new_transport_slot()`. They will share the same underlying `Transport`.

Let's assume we shut down the transport via `A.shutdown()`.
Then the underlying transport will shut down, and A's slot will hold `Arc<RwLock<None>>`.
B's slot will still hold `Arc<RwLock<Some(transport)>>`, but `transport` has shut down, so it doesn't really make a difference behavior wise if the slot is holding `Some(transport)` instead of `None`.
Might as well always use the same underlying `TransportSlot` via the normal `.clone()` and take the value out of the shared option?

szokeasaurusrex: I agree with what you are saying, but because `shutdown` could be an arbitrary function, one could imagine a scenario where the semantics would be different if we go with your suggestion.

Maybe we eventually deprecate/remove this, but I think it is okay for now, and preserves existing behavior

lcian: Ah yeah, I understand now!

lcian (approved):
Looks good!

sentry-core/src/client/mod.rs
```diff
@@ -414,12 +412,12 @@ impl Client {
...
     pub fn is_enabled(&self) -> bool {
-        self.options.dsn.is_some() && self.transport.read().unwrap().is_some()
+        self.options.dsn.is_some() && self.envelope_sender.is_enabled()
     }
 
     /// Captures an event and sends it to sentry.
     pub fn capture_event(&self, event: Event<'static>, scope: Option<&Scope>) -> Uuid {
-        if let Some(ref transport) = *self.transport.read().unwrap() {
+        if self.envelope_sender.is_enabled() {
             if let Some(event) = self.prepare_event(event, scope) {
```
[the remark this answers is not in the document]
szokeasaurusrex: @lcian I made some changes in [this commit](https://github.com/getsentry/sentry-rust/pull/1139/changes/40e0905bfb69991118d2d27ee2c600098bfcf194) to resolve this bug. Would appreciate if you could also review those changes, thanks! 🙏

lcian (approved):

PR #1140, opened 2026-06-08
ref(client): Move `batcher` under `client` module

szokeasaurusrex:
Move `batcher` under `client`, as it is only used in the `client`.

Follow up to #1139, and stacked on that PR.

Closes [#1156](https://github.com/getsentry/sentry-rust/issues/1156)
Closes [RUST-231](https://linear.app/getsentry/issue/RUST-231)


lcian (approved):
Looks good to me.

PR #1142, opened 2026-06-09
feat(transport): Introduce structs for transport options

### LINKED ISSUE #1157: Make transport construction options more flexible
Tracking issue for [#1142](https://github.com/getsentry/sentry-rust/pull/1142).

This PR introduces dedicated transport option structs, giving the transport construction API room to grow without passing the full `ClientOptions` shape into every transport. This supports [#1004](https://github.com/getsentry/sentry-rust/issues/1004) by making it possible to pass client-report loss-recording state to SDK transports, while also leaving a more flexible path for future transport-specific configuration unrelated to client reports.

szokeasaurusrex:
This will allow us to add additional construction options, either to all transports or to individual transports, in the future without breaking the public API. This will be useful in #1004, as we will need to provide transports with a handle to record data losses.

For users who use a custom transport/transport factory, this change may produce a minor behavior breakage, as the `&ClientOptions` received by the `create_transport` method (or the closure, in the case of transport factories which are just closures) will no longer contain all of the options set when init-ing the SDK, only those options which are also present in `TransportOptions`. As the API remains fully compatible, however, we are not considering this to be a public API breakage, and will release these changes in a minor/patch release.

References [#1004](https://github.com/getsentry/sentry-rust/issues/1004)
References [RUST-156](https://linear.app/getsentry/issue/RUST-156/record-transport-drops-and-attach-client-report-envelopes)
Fixes [#1157](https://github.com/getsentry/sentry-rust/issues/1157)
Fixes [RUST-232](https://linear.app/getsentry/issue/RUST-232)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Added [`TransportFactory::create_transport_with_options`](https://docs.rs/sentry-core/latest/sentry_core/trait.TransportFactory.html#method.create_transport_with_options), which constructs transports from [`TransportOptions`](https://docs.rs/sentry-core/latest/sentry_core/struct.TransportOptions.html) instead of full [`ClientOptions`](https://docs.rs/sentry-core/latest/sentry_core/struct.ClientOptions.html) ([#1142](https://github.com/getsentry/sentry-rust/pull/1142)).
 - Added transport-specific options types and `with_options` constructors for built-in HTTP transports, including `ReqwestHttpTransportOptions`, `CurlHttpTransportOptions`, `UreqHttpTransportOptions`, and `EmbeddedSVCHttpTransportOptions` ([#1142](https://github.com/getsentry/sentry-rust/pull/1142)).
+- Added transport worker options types and deprecated the legacy constructors for built-in background transport workers, including `StdTransportThreadOptions` and `TokioTransportThreadOptions` ([#1142](https://github.com/getsentry/sentry-rust/pull/1142)).
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,18 @@
 
 ## Unreleased
 
+### New Features
+
+- Added [`TransportFactory::create_transport_with_options`](https://docs.rs/sentry-core/latest/sentry_core/trait.TransportFactory.html#method.create_transport_with_options), which constructs transports from [`TransportOptions`](https://docs.rs/sentry-core/latest/sentry_core/struct.TransportOptions.html) instead of full [`ClientOptions`](https://docs.rs/sentry-core/latest/sentry_core/struct.ClientOptions.html) ([#1142](https://github.com/getsentry/sentry-rust/pull/1142)).
+- Added transport-specific options types and `with_options` constructors for built-in HTTP transports, including `ReqwestHttpTransportOptions`, `CurlHttpTransportOptions`, `UreqHttpTransportOptions`, and `EmbeddedSVCHttpTransportOptions` ([#1142](https://github.com/getsentry/sentry-rust/pull/1142)).
+
 ### Fixes
 
 - Fixed `ureq` transport handling for HTTP error statuses so `429` rate limits and `413` payload-too-large responses are processed correctly ([#1177](https://github.com/getsentry/sentry-rust/pull/1177)).
+
+### Behavior Changes
+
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+- Added transport-specific options types and `with_options` constructors for built-in HTTP transports, including `ReqwestHttpTransportOptions`, `CurlHttpTransportOptions`, `UreqHttpTransportOptions`, and `EmbeddedSVCHttpTransportOptions` ([#1142](https://github.com/getsentry/sentry-rust/pull/1142)).
+
+### Behavior Changes
+
+- Custom transport factories that implement [`TransportFactory::create_transport`](https://docs.rs/sentry-core/latest/sentry_core/trait.TransportFactory.html#method.create_transport) now receive [`ClientOptions`](https://docs.rs/sentry-core/latest/sentry_core/struct.ClientOptions.html) reconstructed from [`TransportOptions`](https://docs.rs/sentry-core/latest/sentry_core/struct.TransportOptions.html). The reconstructed options include only transport-relevant fields, such as DSN, user agent, proxy settings, and TLS certificate validation settings. This may affect code that reads non-transport fields in `create_transport`, but the API remains source-compatible and this change is included in a minor/patch release ([#1142](https://github.com/getsentry/sentry-rust/pull/1142)).
```
szokeasaurusrex, line 17: I think this is probably reasonable to do in a patch/minor release (i.e. in `0.48.3` rather than `0.49.0`), given that custom transports are probably not super widely used, and we anyways will still include the fields that should be relevant to transports. Although it might break folks relying on the behavior, the API remains compatible, and I think the improvement justifies the risk, since these new APIs should be a lot more flexible and forwards-compatible.

That said, I am willing to reconsider the approach if there are any concerns about the potential breakage here.

sentry/src/transports/curl.rs
```diff
@@ -15,27 +16,79 @@ pub struct CurlHttpTransport {
...
+        Self::new_internal(options)
+    }
+
+    /// Backwards-compatible method for creating a [`CurlHttpTransport`].
+    ///
+    /// Please use [`Self::with_options`] instead.
+    ///
+    /// ### Panics
+    ///
+    /// Panics if called with `options` that lack a DSN.
```
szokeasaurusrex, line 41: This panic is not new. It was previously undocumented.

sentry/src/transports/reqwest.rs
```diff
@@ -18,24 +19,81 @@ pub struct ReqwestHttpTransport {
...
+        Self::new_internal(options)
+    }
+
+    /// Backwards-compatible method for creating a [`ReqwestHttpTransport`].
+    ///
+    /// Please use [`Self::with_options`] instead.
+    ///
+    /// ### Panics
+    ///
+    /// Panics if called with `options` that lack a DSN.
```
szokeasaurusrex, line 42: Not new, just documenting it more explicitly

sentry/src/transports/ureq.rs
```diff
@@ -21,19 +22,74 @@ pub struct UreqHttpTransport {
...
+        Self::new_internal(options)
+    }
+
+    /// Backwards-compatible method for creating a [`UreqHttpTransport`].
+    ///
+    /// Please use [`Self::with_options`] instead.
+    ///
+    /// ### Panics
+    ///
+    /// Panics if called with `options` that lack a DSN.
```
szokeasaurusrex, line 47: Not new, just documenting more explicitly

sentry-core/src/transport/mod.rs
```diff
@@ -32,21 +36,62 @@ pub trait Transport: Send + Sync + 'static {
...
+    /// fallback.
+    fn create_transport_with_options(&self, options: TransportOptions) -> Arc<dyn Transport> {
+        #[expect(deprecated, reason = "need to call deprecated method for back-compat")]
+        self.create_transport(&options.into_client_options())
+    }
+
+    /// The legacy method for creating a transport.
+    ///
+    /// This method exists for backwards compatiblity with custom transport factories, which were
+    /// created before [`TransportFactory::create_transport_with_options`] was added, and thus only
```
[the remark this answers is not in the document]
szokeasaurusrex: This is explicitly documented.

CHANGELOG.md
lcian: I think it's fine to release in a patch.

sentry-core/src/transport/options.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use sentry_types::Dsn;
+
+use crate::ClientOptions;
+
+/// Options for a transport.
+#[derive(Debug)]
+#[must_use]
+#[non_exhaustive]
+pub struct TransportOptions {
```
lcian, line 13: Maybe there's a way to avoid the behavioral breakage by optionally storing the `ClientOptions` here if the transport was created via `ClientOptions` (or maybe not). Either way I think the breakage is really minimal, unlikely to affect any users.

lcian (approved):
Looks good.

sentry-core/src/transport/options.rs
lcian, line 13: So how does one build a `TransportOptions` given that this is non-exhaustive and doesn't implement `Default`?

szokeasaurusrex: It's not intended for users to construct `TransportOptions`. The SDK is responsible for building the options, users should generally only read from them.

lcian (approved):

Issue #1148, opened 2026-06-10
Add transport API for recording client-report losses


szokeasaurusrex:
Add the transport-facing API needed for SDK transports to record lost data without exposing the full client-report aggregator. This should introduce a cloneable recorder handle, pass it through `TransportOptions`, keep backwards-compatible constructors working with a no-op recorder, and document that recorded losses are aggregated into future `client_report` envelope items rather than sent immediately.

### Scope
- Transports can record `(DataCategory, DiscardReason, quantity)` via a public recorder type.
- `TransportOptions` carries the recorder for transports built by the SDK client.
- Deprecated/backwards-compatible transport constructors still compile and use a no-op recorder when no client aggregator is available.

PR #1158, opened 2026-06-10
feat(client-reports): Add transport loss recorder

szokeasaurusrex:
Add a public `ClientReportRecorder` handle for transports to record lost Sentry data without exposing the full client-report aggregator.

Pass the recorder through `TransportOptions` when the SDK client builds a transport. Recorded losses are aggregated into future `client_report` envelope items, so transports can report drops without sending extra requests.

Keep backwards-compatible transport construction paths working by using a no-op recorder when no client-report aggregator is available. Built-in transports will start calling the recorder in follow-up PRs.

Resolves [#1148](https://github.com/getsentry/sentry-rust/issues/1148)
Resolves [RUST-223](https://linear.app/getsentry/issue/RUST-223)

### DIFF
--- a/sentry-core/src/client/client_reports/mod.rs
+++ b/sentry-core/src/client/client_reports/mod.rs
@@ -89,7 +89,7 @@ impl ClientReportAggregator {
         }
     }
 
-    /// Creates a [`ClientReportRecorder`] which records into this aggregator.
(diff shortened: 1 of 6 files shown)


lcian (approved):
Seems good to me.

PR #1144, opened 2026-06-09
feat(client-reports): Client report protocol

szokeasaurusrex:
Added the client report protocol.

Resolves [#1001](https://github.com/getsentry/sentry-rust/issues/1001)
Resolves [RUST-153](https://linear.app/getsentry/issue/RUST-153/add-client-report-protocol-envelope-item-support-in-sentry-types)

PR #1166, opened 2026-06-12
ref(envelope): Reuse item types in serialization

szokeasaurusrex:
This refactor should force us to make sure we add EnvelopeItemTypes corresponding to new EnvelopeItems: basically this sort of error should become impossible: https://github.com/getsentry/sentry-rust/pull/1144#discussion_r3403773888.


lcian (approved):
LGTM

Issue #1151, opened 2026-06-10
Record HTTP send losses in the reqwest transport


szokeasaurusrex:
Update `ReqwestHttpTransport` to record losses for HTTP-level send failures after an envelope has left the transport thread. Network/request errors should be recorded as `send_error`; HTTP `413` and other non-`429` `4xx`/`5xx` responses should also be recorded as `send_error`; `429` should continue to update rate limits without recording a client-report loss for that response.

### Scope
- Request failures record losses for the attempted envelope.
- `413` responses record `send_error` losses and keep the existing debug message.
- Other non-`429` `4xx`/`5xx` responses record `send_error` losses.
- `2xx` responses and `429` responses do not record send losses.

PR #1174, opened 2026-06-17
chore(reqwest): Record envelope losses in the Reqwest transport

szokeasaurusrex:
Record lost envelopes in the `reqwest` transport. Also, add necessary loss reason enum variants.

Closes [https://github.com/getsentry/sentry-rust/issues/1151](https://github.com/getsentry/sentry-rust/issues/1151)
Closes [RUST-226](https://linear.app/getsentry/issue/RUST-226)


sentry-core/src/client/client_reports/recorder.rs
```diff
@@ -43,7 +43,23 @@ pub struct Recorder {
 /// This is a subset of [`Reason`], as defined in [`sentry_types`] because only some of those
 /// reasons may be applicable to transports.
 #[non_exhaustive]
-pub enum TransportLossReason {}
+pub enum TransportLossReason {
```
lcian, line 51: Should we just use the `Reason` directly?

sentry/src/transports/reqwest.rs
```diff
@@ -119,9 +124,17 @@ impl ReqwestHttpTransport {
...
         let send_fn = move |envelope: Envelope, mut rl: RateLimiter| {
             let mut body = Vec::new();
-            envelope.to_writer(&mut body).unwrap();
+            envelope
+                .to_writer(&mut body)
+                .inspect_err(|_| {
+                    client_report_recorder
+                        .record_lost_envelope(&envelope, LossReason::InternalError)
+                })
+                .unwrap();
```
lcian: Should we avoid this unwrap while we're at it? 😅

lcian: On the other hand though, panicking here would probably indicate a bug in the SDK, so it might be "useful" to keep it.

sentry-core/src/client/client_reports/recorder.rs
szokeasaurusrex: I think a separate enum makes sense, as it allows us to specify which reasons should be used by transports, as that will eventually only be a subset of `Reason` (e.g. `before_send` will never apply to transports)

lcian (approved):
LGTM

sentry/src/transports/reqwest.rs
szokeasaurusrex: I guess it could also happen if we run out of memory.

I agree that this `unwrap` is not ideal. I'll replace it with an `expect` for now, because at least that will give a clearer error message, but I think it is out of scope right now to figure out and implement the better behavior.

lcian (approved):
LGTM

sentry/src/transports/reqwest.rs
```diff
@@ -119,9 +124,16 @@ impl ReqwestHttpTransport {
 
         let send_fn = move |envelope: Envelope, mut rl: RateLimiter| {
             let mut body = Vec::new();
-            envelope.to_writer(&mut body).unwrap();
+            envelope
+                .to_writer(&mut body)
+                .inspect_err(|_| {
+                    client_report_recorder.record_lost_data(&envelope, LossReason::InternalError);
+                })
+                .expect("envelope should serialize successfully");
```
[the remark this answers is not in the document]
szokeasaurusrex: This matches the pre-existing behavior

PR #1177, opened 2026-06-18
fix(transport): Handle ureq HTTP error statuses

### LINKED ISSUE #1176: Ureq transport likely does not handle HTTP 429 and 413 responses
`UreqHttpTransport` likely does not run its HTTP-status-specific handling for Sentry error responses because ureq defaults `http_status_as_error` to `true`. The transport currently sends with `agent.post(...).send(...)`, then handles rate-limit headers, fallback HTTP `429`, and HTTP `413` payload-too-large logging only in the `Ok(response)` branch. With ureq's default behavior, HTTP `4xx`/`5xx` statuses can instead return `Err(ureq::Error::StatusCode(code))`, so fallback rate-limit updates for `429` and the existing payload-too-large debug message for `413` likely do not work with the default agent. See the current ureq transport response handling in [`sentry/src/transports/ureq.rs`](<https://github.com/getsentry/sentry-rust/blob/a4005496a2f4cea895843ea8cae8ef37fd392724/sentry/src/transports/ureq.rs#L145-L176>).

This also affects [#1153](<https://github.com/getsentry/sentry-rust/issues/1153>). That issue needs reliable status classification so `413` and other non-`429` `4xx`/`5xx` responses can record `send_error`, while `429` continues to update rate limits without recording a client-report loss. We should fix the ureq status handling first, then implement getsentry/sentry-rust#1153 on top of the corrected response path.

szokeasaurusrex:
Previously, ureq returned HTTP 4xx and 5xx responses as send errors, so the transport skipped the response-handling logic for those statuses. As a result, fallback `429` rate-limit handling and `413` payload-too-large logging were not reached.

Set `http_status_as_error(false)` on ureq envelope requests so HTTP error statuses return as responses and flow through the existing response handling.

Fixes [#1176](https://github.com/getsentry/sentry-rust/issues/1176)
Fixes [RUST-243](https://linear.app/getsentry/issue/RUST-243)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Fixed `ureq` transport handling for HTTP error statuses so `429` rate limits and `413` payload-too-large responses are processed correctly ([#1177](https://github.com/getsentry/sentry-rust/pull/1177)).
+
 ## 0.48.2
 
 ### New Features
--- a/sentry/src/transports/ureq.rs
+++ b/sentry/src/transports/ureq.rs
@@ -90,7 +90,13 @@ impl UreqHttpTransport {
         let thread = TransportThread::new(move |envelope, rl| {
             let mut body = Vec::new();
             envelope.to_writer(&mut body).unwrap();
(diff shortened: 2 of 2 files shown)


lcian (approved):
Good catch, this might've been missed when we bumped ureq to the new major.

PR #1178, opened 2026-06-18
chore(ureq): Record envelope losses in the Ureq transport

szokeasaurusrex:
Record lost envelopes in the `ureq` transport after the transport thread accepts them for sending. Ureq request failures are recorded as `network_error`; non-`429` HTTP `4xx`/`5xx` responses are recorded as `send_error`.

Keep existing rate-limit handling for `429` responses and the existing payload-too-large debug message for `413` responses.

Closes [#1153](https://github.com/getsentry/sentry-rust/issues/1153)
Closes [RUST-228](https://linear.app/getsentry/issue/RUST-228)


lcian (approved):
LGTM

Issue #1181, opened 2026-06-19
Allow `Envelope::filter` to take a callback to report filtered out envelopes


szokeasaurusrex:
We need a way to know what gets filtered out for getsentry/sentry-rust#1149

Issue #1112, opened 2026-05-06
Flaky test: `future_cross_thread_info_span`


szokeasaurusrex:
I have noticed [this test](<https://github.com/getsentry/sentry-rust/blob/2bebbd44e47fc9056cc04cf5f755cd5b6fc9dfd7/sentry-tracing/tests/await_cross_thread_guard.rs#L13-L51>) flaking multiple times. I am unsure if there is a bug in the SDK or in the test itself

lcian:
https://github.com/getsentry/sentry-rust/actions/runs/29086740273/job/86355230440?pr=1214

```
thread 'future_cross_thread_info_span' (3694) panicked at sentry-tracing/tests/await_cross_thread_guard.rs:189:5:
expected exactly one envelope, got 0: []
```

PR #1182, opened 2026-06-19
feat(envelopes): Observe filtered envelope items

szokeasaurusrex:
Add an `EnvelopeFilter` trait and callback container so callers can observe
items removed by `Envelope::filter`. This includes items rejected directly by
the filter and attachments removed later when their event or transaction is no
longer present.

Keep existing closure-based callers source-compatible through a blanket
`FnMut(&EnvelopeItem) -> bool` implementation. Some closure sites may need an
explicit item type annotation after this generalization.

Closes [#1181](https://github.com/getsentry/sentry-rust/issues/1181)
Closes [RUST-245](https://linear.app/getsentry/issue/RUST-245)


sentry-types/src/protocol/envelope.rs
```diff
@@ -473,30 +514,43 @@ impl Envelope {
     /// contains an [`EnvelopeItem::Event`] or [`EnvelopeItem::Transaction`].
     ///
     /// [`None`] is returned if no items remain in the Envelope after filtering.
-    pub fn filter<P>(self, mut predicate: P) -> Option<Self>
+    pub fn filter<F>(self, mut filter: F) -> Option<Self>
     where
-        P: FnMut(&EnvelopeItem) -> bool,
+        F: EnvelopeFilter,
```
szokeasaurusrex, line 519: This change can break the inference (it broke it in the `sentry` crate), but it is [considered a minor change](https://doc.rust-lang.org/cargo/reference/semver.html#generic-bounds-loosen), as it is simply a loosening of the type bounds, due to the blanket `EnvelopeFilter` implementation for `FnMut(&EnvelopeItem) -> bool`.

I think the breakage is acceptable, as this is likely not used widely outside the SDK

lcian:
CI [failed](https://github.com/getsentry/sentry-rust/actions/runs/27948985722/job/82700854955?pr=1182#step:6:303) in an unexpected way
```
thread 'future_cross_thread_info_span' (3871) panicked at sentry-tracing/tests/await_cross_thread_guard.rs:191:28:
expected exactly one envelope
```
Seems like something to look into, as this should not be flaky.
As far as I can tell, the changes from the current stack shouldn't affect this, as our `TestTransport` shouldn't be wired up to send client reports.
My clanker just goes in circles when trying to look into this.

sentry-types/src/protocol/envelope.rs
lcian: Agreed.

sentry-types/src/protocol/envelope.rs
```diff
@@ -367,6 +368,46 @@ impl Items {
...
+pub struct EnvelopeFilterCallbacks<F, C> {
+    filter: F,
+    on_filtered: C,
+}
+
+impl<F, C> EnvelopeFilterCallbacks<F, C> {
+    /// Create a new [`EnvelopeFilterCallbacks`].
+    ///
+    /// `filter` will be called to determine whether the envelope should be kept. `on_filtered`
+    /// will be called on all envelopes which are then dropped.
```
lcian:
```suggestion
    /// `filter` will be called to determine whether the envelope items should be kept. `on_filtered`
    /// will be called on all envelope items which are then dropped.
```

szokeasaurusrex:
@lcian regarding the CI failure, I have also noticed this flakiness, and have already opened an issue to investigate: https://github.com/getsentry/sentry-rust/issues/1112.

sentry-types/src/protocol/envelope.rs
szokeasaurusrex: good catch! 🙏

sentry-types/src/protocol/envelope.rs
```diff
@@ -473,30 +514,43 @@ impl Envelope {
...
-            if predicate(&item) {
+            if filter.filter(&item) {
                 filtered.add_item(item);
+            } else {
+                filter.on_filtered(item);
             }
         }
 
         // filter again, removing attachments which do not make any sense without
         // an event/transaction
```
lcian, line 540: I guess this is not true anymore as we have "trace attachments". If the SDK implemented those, you would be able to send attachments in a standalone way. But I also don't see a usecase right now to implement that.

lcian (approved):
LGTM.

Issue #1150, opened 2026-06-10
Record queue and rate-limit losses in the tokio transport thread


szokeasaurusrex:
Update the tokio-backed `TransportThread` used by the reqwest transport to record losses before an envelope reaches the HTTP transport. Queue overflow in `send` should be recorded as `queue_overflow`, and envelopes/items dropped by global or per-item rate limiting should be recorded as `rate_limit` with the correct data categories and quantities.

### Scope
- `try_send` failures record losses for the dropped envelope.
- Global rate-limit drops record losses for every item in the envelope.
- Per-item rate-limit filtering records losses only for filtered items.
- Existing async send, flush, and shutdown behavior is unchanged.

PR #1186, opened 2026-06-22
chore(transport): Record tokio thread client report losses

szokeasaurusrex:
Record envelopes dropped before the reqwest HTTP transport sees them. The tokio transport thread now reports queue overflow, disconnected worker, and rate-limit drops through the client report recorder.

Shutdown losses remain unreported because shutdown ends the opportunity to send another envelope carrying the aggregated client report.

Fixes [#1150](https://github.com/getsentry/sentry-rust/issues/1150)
Fixes [RUST-225](https://linear.app/getsentry/issue/RUST-225)

### DIFF
--- a/sentry/src/transports/reqwest.rs
+++ b/sentry/src/transports/reqwest.rs
@@ -122,17 +122,20 @@ impl ReqwestHttpTransport {
         let auth = dsn.to_auth(Some(&user_agent)).to_string();
         let url = dsn.envelope_api_url().to_string();
 
+        let send_fn_client_report_recorder = client_report_recorder.clone();
+
         let send_fn = move |envelope: Envelope, mut rl: RateLimiter| {
             let mut body = Vec::new();
             envelope
                 .to_writer(&mut body)
                 .inspect_err(|_| {
-                    client_report_recorder.record_lost_data(&envelope, LossReason::InternalError);
+                    send_fn_client_report_recorder
+                        .record_lost_data(&envelope, LossReason::InternalError);
                 })
                 .expect("envelope should serialize successfully");
             let request = client.post(&url).header("X-Sentry-Auth", &auth).body(body);
 
-            let client_report_recorder = client_report_recorder.clone();
+            let client_report_recorder = send_fn_client_report_recorder.clone();
 
             // NOTE: because of lifetime issues, building the request using the
             // `client` has to happen outside of this async block.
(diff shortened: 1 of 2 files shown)


sentry/src/transports/tokio_thread.rs
```diff
@@ -105,21 +125,25 @@ impl TransportThread {
                             }
                         };
 
-                        if let Some(time_left) =  rl.is_disabled(RateLimitingCategory::Any) {
```
lcian, line 108: Did rustfmt not run on this file? 🤔

szokeasaurusrex: no clue 🤣

sentry/src/transports/reqwest.rs
```diff
@@ -122,17 +122,20 @@ impl ReqwestHttpTransport {
         let auth = dsn.to_auth(Some(&user_agent)).to_string();
         let url = dsn.envelope_api_url().to_string();
 
+        let send_fn_client_report_recorder = client_report_recorder.clone();
```
lcian, line 125:
```suggestion
        let client_report_recorder = client_report_recorder.clone();
```

szokeasaurusrex: This would not work, as we need `client_report_recorder` [again below](https://github.com/getsentry/sentry-rust/pull/1186/changes/BASE..807212aeb344dd3d0fd150aab3795bd71c0c075d#diff-d906bba2df2a016a6c73849f2e3842f1174232f725591a7ff99caa7d8fc5cb9aR194). We need a clone with a different name unfortunately.

sentry/src/transports/tokio_thread.rs
```diff
@@ -105,21 +124,22 @@ impl TransportThread {
...
-                        match rl.filter_envelope(envelope) {
+                        match rl.filter(envelope, &handle_client_report_recorder) {
                             Some(envelope) => {
                                 rl = send(envelope, rl).await;
-                            },
+                            }
                             None => {
                                 sentry_debug!("Envelope was discarded due to per-item rate limits");
-                            },
+                            }
```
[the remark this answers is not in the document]
szokeasaurusrex: We probably should investigate this further prior to merging — marking as draft in the meantime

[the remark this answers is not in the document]
szokeasaurusrex: We decided that `ratelimit_backoff` is appropriate in this case

lcian (approved):
🚀

Issue #1187, opened 2026-06-22
Record event pipeline drops in client reports


szokeasaurusrex:
Record SDK-side event drops in [`sentry-core/src/client/mod.rs`](<https://github.com/getsentry/sentry-rust/blob/ca000031a1572eac2364cb3312e7e78d4b3a2db2/sentry-core/src/client/mod.rs>): scope event processors and `Integration::process_event` drops should record `error` / `event_processor`, `before_send` drops should record `error` / `before_send`, and error event sampling drops should record `error` / `sample_rate`. Record client report losses in accordance with [the spec](<https://develop.sentry.dev/sdk/telemetry/client-reports/>).

The implementation should record integration drops centrally in `Client::prepare_event`; individual integration crates should not need changes. Add tests that trigger each drop path and then send a follow-up envelope so the attached `client_report` item can be asserted.

PR #1191, opened 2026-06-22
chore(client-reports): Record event pipeline drops

szokeasaurusrex:
Record SDK-side error event drops in client reports for the event pipeline paths that can discard events before they reach transport. Scope processors and integration processors now report `event_processor`, `before_send` reports `before_send`, and error event sampling reports `sample_rate`.

Add integration coverage that verifies each drop path is attached to the next outgoing envelope as a client report.

Fixes [#1187](https://github.com/getsentry/sentry-rust/issues/1187)
Fixes [RUST-246](https://linear.app/getsentry/issue/RUST-246)

### DIFF
--- a/sentry-core/src/session.rs
+++ b/sentry-core/src/session.rs
@@ -484,7 +484,9 @@ mod session_impl {
             } else {
                 panic!("expected session");
             }
-            assert_eq!(items.next(), None);
+            // A client report gets captured for the dropped event.
+            assert!(matches!(items.next(), Some(EnvelopeItem::ClientReport(_))));
+            assert!(items.next().is_none());
         }
 
         #[test]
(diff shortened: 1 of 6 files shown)


sentry-core/src/session.rs
```diff
@@ -566,7 +568,12 @@ mod session_impl {
             } else {
                 panic!("expected session");
             }
-            assert_eq!(items.next(), None);
+            // The envelope may or may not contain a client report.
+            assert!(matches!(
+                items.next(),
+                None | Some(EnvelopeItem::ClientReport(_))
+            ));
```
lcian, line 575: Does this depend on timing?

lcian (approved):
LGTM

sentry-core/src/session.rs
szokeasaurusrex: It's random; we have the `sample_rate` set to `0.5` so I guess it is `None` or `Some(EnvelopeItem::ClientReport(_))` each with 50% probability, depending on whether the last error was sampled or not.

lcian: Aaah I see.

Issue #1189, opened 2026-06-22
Record log and metric callback drops in client reports


szokeasaurusrex:
Follow up on [#1003](https://github.com/getsentry/sentry-rust/issues/1003) if that issue is split into smaller client-report tasks. Record SDK-side drops caused by `before_send_log` and `before_send_metric` in [`sentry-core/src/client/mod.rs`](https://github.com/getsentry/sentry-rust/blob/ca000031a1572eac2364cb3312e7e78d4b3a2db2/sentry-core/src/client/mod.rs): dropped logs should record both `log_item` / `before_send` and `log_byte` / `before_send`, and dropped metrics should record both `trace_metric` / `before_send` and `trace_metric_byte` / `before_send`.

Use an approximate serialized byte count; serialization failure for the estimate should contribute zero bytes while still recording the item count. Do not record drops when logs or metrics are disabled via `enable_logs = false` or `enable_metrics = false`, because those paths are disabled instrumentation rather than discarded captured telemetry.

PR #1193, opened 2026-06-23
chore: Record client reports for dropped logs/metrics

szokeasaurusrex:
Resolves [#1189](https://github.com/getsentry/sentry-rust/issues/1189)
Resolves [RUST-248](https://linear.app/getsentry/issue/RUST-248)

### DIFF
--- a/sentry-types/src/protocol/client_report/envelope_losses.rs
+++ b/sentry-types/src/protocol/client_report/envelope_losses.rs
@@ -22,6 +22,7 @@ pub trait LossSource: private::Sealed {
 ///
 /// [`Report`]: super::Report
 #[non_exhaustive]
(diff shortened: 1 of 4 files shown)


sentry-types/src/protocol/client_report/envelope_losses.rs
```diff
@@ -22,6 +22,7 @@ pub trait LossSource: private::Sealed {
 ///
 /// [`Report`]: super::Report
 #[non_exhaustive]
+#[derive(Debug, Clone)]
```
lcian: Could this derive `Copy` and then use `self.iter().copied()` below?

lcian (approved):

PR #1195, opened 2026-06-24
chore(client-reports): Prevent client report loss when envelope dropped

szokeasaurusrex:
Prevent client report data from being lost when an envelope containing a client report is dropped by adding all losses from the lost client report back into the aggregator. This ensures that the next outgoing envelope will include the loss data.

This change fixes some undesirable behavior I discovered while manually testing client reports: if the transport queue overflows when sending an error, we record a lost error event. If we then immediately try to send another error, that envelope will pick up the loss, draining it from the aggregator, but if the queue is still full, that envelope will also get lost with the client report. We will record the lost error event from the second envelope, but without this change, we would not include the loss from the first envelope. If we consider the case where many envelopes get sent in rapid succession and overflow the queue, we could end up with a severe undercount of reported losses as a result.


lcian (approved):
Nice.

Issue #1201, opened 2026-06-25
Enforce strict trace continuation when starting transactions


szokeasaurusrex:
Implement the strict trace continuation decision when creating transactions. The SDK should reject incoming traces when the incoming `sentry-org_id` is incompatible with the SDK organization ID according to the [spec decision matrix](<https://develop.sentry.dev/sdk/foundations/trace-propagation/#continuing-traces>). This is the behavior-changing implementation task for getsentry/sentry-rust#1016.

Add a `CHANGELOG.md` entry because this can significantly alter incoming trace continuation behavior.

<details><summary>Implementation notes</summary>

Target files:

* `sentry-core/src/performance/mod.rs`
* `sentry-actix` tests
* `sentry-tower` tests
* `CHANGELOG.md`

Implementation:

* In `Transaction::new`, evaluate the continuation decision before sampling and before constructing `protocol::TraceContext`.
* Use:
  * `ctx.incoming_org_id`
  * `client.options().resolved_org_id()` when a client exists
  * `client.options().strict_trace_continuation`
* Add a private helper such as `fn should_continue_trace(incoming: Option<OrganizationId>, sdk: Option<OrganizationId>, strict: bool) -> bool`.
* Decision rules:
  * both IDs present and different -> start a new trace regardless of `strict_trace_continuation`
  * exactly one ID present -> start a new trace when `strict_trace_continuation` is `true`; continue when `false`
  * both IDs present and equal -> continue
  * both IDs absent -> continue
* On rejection, replace inbound trace state before sampling:
  * new random/default `trace_id`
  * `parent_span_id: None`
  * new/default `span_id`
  * `sampled: None`
  * `incoming_org_id: None`
* Ensure parent sampling decisions from rejected traces do not influence `traces_sampler` or `ctx.sampled` fallback.
* Preserve existing behavior when there is no active client, since no SDK organization ID or strict option is available.
* Do not create head-of-trace DSC in this task; that belongs to later DSC work.

Core tests:

* Cover the full spec matrix for incoming org ID, SDK org ID, and `strict_trace_continuation`.
* On rejection, assert the transaction trace ID differs from the incoming trace ID, `parent_span_id` is `None`, and parent sampling is not inherited.
* Assert explicit `ClientOptions::org_id` overrides the DSN-derived organization ID.

Integration tests:

* Add actix and tower coverage with the behavior change because these integrations exercise the automatic inbound server path.
[...]

PR #1209, opened 2026-06-29
chore(tracing): Store incoming baggage org ID

### LINKED ISSUE #1200: Parse sentry-org_id from incoming baggage for transaction continuation
Parse `sentry-org_id` from incoming `baggage` headers and store the parsed value on `TransactionContext` for later strict continuation decisions. This task should not enforce strict continuation yet and should not implement full Dynamic Sampling Context parsing or outgoing baggage propagation. This is part of getsentry/sentry-rust#1016.

The goal is to make the incoming organization ID available to transaction creation while keeping the change behavior-neutral.

<details><summary>Implementation notes</summary>

Target file:

* `sentry-core/src/performance.rs`

Implementation:

* Add an internal parsed incoming trace metadata type, for example `struct IncomingTrace { sentry_trace: SentryTrace, org_id: Option<OrganizationId> }`.
* Preserve public compatibility for `parse_headers`; either keep `parse_headers` returning `Option<SentryTrace>` and add a private `parse_incoming_trace_headers`, or add a distinct helper while leaving `parse_headers` unchanged.
* Extend `TransactionContext` with an internal `incoming_org_id: Option<OrganizationId>` field.
* Initialize `incoming_org_id` to `None` in constructors except `continue_from_headers`.
* Update `TransactionContext::continue_from_headers` to parse both `sentry-trace` and `baggage`.
* Parse baggage minimally:
  * scan comma-separated baggage members
  * trim whitespace around members
  * match only key `sentry-org_id`
  * parse the value as `OrganizationId`
  * treat malformed, empty, duplicate-unparseable, or absent values as `None`
* Do not parse, preserve, or serialize any other DSC fields.

Tests:

* `baggage: sentry-org_id=123` is parsed.
* Baggage with unrelated fields before or after `sentry-org_id` is parsed.
* Mixed-case `baggage` header name is accepted.
* Malformed org IDs are treated as absent.
* No `sentry-trace` means the returned context starts a fresh trace and does not carry an incoming org ID.
* Validate with `cargo test -p sentry-core performance`.

</details>

szokeasaurusrex:
Parse `sentry-org_id` from incoming `baggage` headers alongside `sentry-trace` and store it on `TransactionContext`.

This keeps the change behavior-neutral while making the incoming org ID available for strict trace continuation work in a getsentry/sentry-rust#1201. Malformed or missing org IDs are treated as absent, and baggage without a valid `sentry-trace` does not attach incoming org metadata to a fresh transaction context.

Closes [#1200](<https://github.com/getsentry/sentry-rust/issues/1200>)  <br>Closes getsentry/sentry-rust#1200  <br>References [#1201](<https://github.com/getsentry/sentry-rust/issues/1201>)  <br>References getsentry/sentry-rust#1201

### DIFF
--- a/sentry-core/src/performance/headers.rs
+++ b/sentry-core/src/performance/headers.rs
@@ -3,17 +3,23 @@
 use std::error::Error;
 use std::fmt::{Display, Formatter, Result as FmtResult};
 
+use sentry_types::protocol::v7::OrganizationId;
+
 use crate::protocol::{SpanId, TraceId};
 
 /// A key-value header pair.
 type Header<'h> = (&'h str, &'h str);
+
+/// The baggage key for the Sentry org ID.
+const SENTRY_ORG_ID: &str = "sentry-org_id";
 
 /// The [trace propagation] context.
 ///
 /// Contains the information necessary for propagating Sentry traces and continuing traces from
 /// incoming requests.
 ///
-/// The data stored in this struct can be parsed from and transmitted as `sentry-trace` headers.
+/// The data stored in this struct can be parsed from and transmitted as `sentry-trace` and Sentry
+/// baggage headers.
 ///
 /// Note that the Rust SDK only partially supports trace propagation, certain features such as
 /// [dynamic sampling] may be missing or incomplete.
@@ -25,6 +31,7 @@ pub struct TracePropagationContext {
     pub(crate) trace_id: TraceId,
     pub(crate) span_id: SpanId,
     pub(super) sampled: Option<bool>,
+    pub(super) org_id: Option<OrganizationId>,
 }
 
 #[derive(Debug, Clone)]
@@ -53,6 +60,7 @@ impl TracePropagationContext {
             trace_id,
             span_id,
             sampled: None,
+            org_id: None,
         }
     }
 
@@ -68,6 +76,7 @@ impl TracePropagationContext {
             trace_id,
             span_id,
             sampled,
+            org_id: _,
         } = self;
 
         let sampled_suffix = sampled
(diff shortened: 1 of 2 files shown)


sentry-core/src/performance/headers.rs
```diff
@@ -3,17 +3,23 @@
...
 
+use sentry_types::protocol::v7::OrganizationId;
+
 use crate::protocol::{SpanId, TraceId};
 
 /// A key-value header pair.
 type Header<'h> = (&'h str, &'h str);
 
+/// The baggage key for the Sentry org ID.
+const SENTRY_ORG_ID: &str = "sentry-org_id";
```
Litarnus, line 14: Maybe we can also add `sentry-trace` and `baggage` as constants for consistency

sentry-core/src/performance/headers.rs
```diff
@@ -190,6 +208,42 @@ impl std::fmt::Display for SentryTrace {
...
+    }
+
+    /// Update `self` with a key-value pair from the baggage header.
+    ///
+    /// The value is only updated if it is valid, otherwise the old value is kept.
+    fn update_from_value(&mut self, key: &str, value: &str) {
+        if key == SENTRY_ORG_ID {
+            self.org_id = value.parse().ok().or(self.org_id);
+        }
+    }
```
Litarnus, line 250: nit: I think we can just have a `filter` after `flat_map` to filter by org ID, but this solution is also fine if ever want to parse more data

Litarnus (approved):

sentry-core/src/performance/headers.rs
szokeasaurusrex: Yeah, let's keep it so that we can parse more things in the future

Issue #1208, opened 2026-06-29
Preserve Hub binding for streaming responses in sentry-tower/axum


szokeasaurusrex:
When using the `sentry-tower`/`axum` integration with streaming responses, the middleware binds the hub only for the request handler. Once the handler returns a streaming body, the middleware unwinds and `axum` drains the stream outside that hub. This causes missing context and any newly set context to be attached to the wrong hub during stream evaluation.

Track the tracing/transaction behavior separately in getsentry/sentry-rust#1053.

A possible direction is to bind the hub to the response stream, similar to getsentry/objectstore#526.

PR #1214, opened 2026-06-30
feat(core): Stream extension trait

lcian:
### Description

This adds an extension trait for `futures_core::Stream` that can be used to `.bind_hub(...)` on a stream, much like the existing machinery we have for futures.
This came up a few times already, and it happens to be one of the ways one could work around https://github.com/getsentry/sentry-rust/issues/1208 (given that the problematic response body is a stream, one can use such a stream wrapper to bind the current hub before converting that stream to a response -- I still think a proper separate fix should be introduced in the `sentry-tower` instrumentation though, as the user should not have to remember to do this manually. This is a more general construct that can be used in other scenarios too.


sentry-core/src/stream.rs
```diff
@@ -0,0 +1,131 @@
...
+impl<S> Stream for SentryStream<S>
+where
+    S: Stream,
+{
+    type Item = S::Item;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let hub = self.hub.clone();
+        // https://doc.rust-lang.org/std/pin/index.html#pinning-is-structural-for-field
+        let stream = unsafe { self.map_unchecked_mut(|s| &mut s.stream) };
```
lcian, line 38: It is possible to avoid this direct usage of `unsafe` (and the one in `SentryFuture`) by adding a dependency to `pin-project-lite`.

szokeasaurusrex:
@lcian what is the status with this PR? Is it abandoned or are you still planning to finish it and get it ready for review?

lcian:
@szokeasaurusrex it's ready for review now, please take a look whenever you have time. I don't need this at the moment but it came up a couple times, so I think it would be a nice utility to have.

sentry-core/Cargo.toml
```diff
@@ -31,6 +31,7 @@ logs = []
 metrics = []
 
 [dependencies]
+futures-core = { workspace = true }
```
szokeasaurusrex, line 34: h: We [should avoid adding dependencies](https://develop.sentry.dev/sdk/getting-started/philosophy/#dependencies-cost) when there is not a strong reason why we need the dependency.

In this case, I am not seeing a strong reason for `sentry-core` to depend on `futures-core`.

I see a few alternatives that would avoid this unconditional dependency:

1. It seems we could probably provide the extension trait in a separate crate; perhaps we could name it `sentry-futures`, kinda like we do for integrations. The crate itself could perhaps later be evolved into a full-fledged integration.
2. Alternatively, we can put the trait behind a feature-flag. Then, `futures-core` would be an optional dependency activated by that flag. If we go with this path, I would probably expose the trait in `sentry`, not `sentry-core`, unless there's a reason why it needs to be in `sentry-core`.

sentry-core/src/stream.rs
```diff
@@ -0,0 +1,131 @@
...
+/// This activates the given hub for the duration of the inner stream's `poll_next`
+/// method. Users usually do not need to construct this type manually, but
+/// rather use the [`StreamExt::bind_hub`] method instead.
+///
+/// [`StreamExt::bind_hub`]: trait.StreamExt.html#method.bind_hub
+#[derive(Debug)]
+pub struct SentryStream<S> {
+    hub: Arc<Hub>,
+    stream: S,
+}
```
szokeasaurusrex, line 20: m: If possible, I would make this type private, and adjust `SentryStreamExt::bind_hub` to return `impl Stream<Self::Item>`.

```suggestion
struct SentryStream<S> {
    hub: Arc<Hub>,
    stream: S,
}
```

sentry-core/src/stream.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn bind_hub<H>(self, hub: H) -> SentryStream<Self>
+    where
+        H: Into<Arc<Hub>>,
+    {
+        SentryStream {
+            stream: self,
+            hub: hub.into(),
+        }
+    }
+}
```
szokeasaurusrex, line 66: You need to adjust this trait a bit in order to be able to return `impl Stream`

```suggestion
/// Stream extensions for Sentry.
pub trait SentryStreamExt: Sized + Stream {
    /// Binds a hub to this stream.
    ///
    /// This ensures that the stream is polled within the given hub.
    fn bind_hub<H>(self, hub: H) -> impl Stream<Item = Self::Item>
    where
        H: Into<Arc<Hub>>,
    {
        SentryStream {
            stream: self,
            hub: hub.into(),
        }
    }
}
```

szokeasaurusrex (changes requested):
Hey @lcian, apologies that it took me so long to get around to this.

I am requesting changes because this adds a dependency on `futures-core` to `sentry-core`, and we should try to avoid that 🙏 more details on inline comment

PR #1221, opened 2026-07-02
feat(core): Introduce builder API for `ClientOptions`

### LINKED ISSUE #1220: Introduce builder API for `ClientOptions`
Doing so is needed to allow configuring `ClientOptions` once made `#[non_exhaustive]` (see parent issue)

szokeasaurusrex:
Introduce a builder API as the preferred method for constructing `ClientOptions`. In the changelog, we note that the old construction method is deprecated, but unfortunately, this is the only deprecation notice; Rust does not give us a way to warn at compile time.

As the builder methods are the new primary ways for setting things, we move the field documentation to the respective builder methods. Also, we make `ClientOptions` `#[must_use]` to ensure that folks use the options after building them.

Resolves [#1220](https://github.com/getsentry/sentry-rust/issues/1220)
Resolves [RUST-259](https://linear.app/getsentry/issue/RUST-259)

### DIFF
--- a/sentry-core/src/clientoptions.rs
+++ b/sentry-core/src/clientoptions.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use crate::constants::USER_AGENT;
-use crate::performance::TracesSampler;
+use crate::performance::{TracesSampler, TransactionContext};
 use crate::protocol::{Breadcrumb, Event, Log, Metric};
 use crate::types::Dsn;
 use crate::{Integration, IntoDsn, TransportFactory};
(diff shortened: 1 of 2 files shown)


sentry-core/src/clientoptions.rs
```diff
@@ -84,156 +87,566 @@ impl MaxRequestBodySize {
...
+    /// Sets the [DSN](field@ClientOptions::dsn) to use.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value fails to parse as a [DSN](`Dsn`).
+    #[inline]
+    pub fn dsn(self, dsn: &str) -> Self {
+        let dsn = Some(dsn.parse().expect("invalid value for DSN"));
+        Self { dsn, ..self }
+    }
```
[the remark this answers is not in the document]
szokeasaurusrex: bad clanker: people who don't want to have a `dsn` just don't call the method

sentry-core/src/clientoptions.rs
```diff
@@ -84,156 +87,566 @@ impl MaxRequestBodySize {
...
+
+    /// Sets the optional [HTTPS proxy](field@ClientOptions::https_proxy) to use.
+    ///
+    /// This defaults to the `HTTPS_PROXY` environment variable, or `http_proxy` if that one
+    /// exists.
+    #[inline]
+    pub fn https_proxy<T>(self, https_proxy: T) -> Self
+    where
+        T: Into<Cow<'static, str>>,
```
alexander-alderman-webb, line 543: Why not accept borrowed stuff as well, even if it has to allocated? (since it's just once for configuration).

```suggestion
        T: Into<String>,
```

alexander-alderman-webb (approved):

sentry-core/src/clientoptions.rs
szokeasaurusrex: Discussed this offline; also, keeping with `Cow<'static, str>` is consistent with how this value is [typed in the struct](https://github.com/getsentry/sentry-rust/pull/1221/changes/BASE..754109dfc3b67a49ac7de584e05e711d3301147f#diff-9862b73c669fb70bacec5a9890b32ead609c25699f2d38b9a81e2dd863cf1e46L166).

PR #1235, opened 2026-07-10
feat(core): Add `read_scope` for read-only scope access

bardiharborow:
`Hub::configure_scope` clones the entire scope and writes it back under the hub's write lock, which is wasteful on hot paths that only need to read scope data (e.g. integrations resolving the current span per event).

This PR adds `Hub::read_scope` which provides read-only access without the clone and write-back, and the global `sentry::read_scope` exposes the same via the active hub, mirroring `sentry::configure_scope`. It also switches existing call sites where appropriate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### New Features
+
+- Added `sentry::read_scope` and `Hub::read_scope`, read-only counterparts to [`configure_scope`](https://docs.rs/sentry-core/0.48.4/sentry_core/fn.configure_scope.html) that avoid cloning the scope and writing it back.
+
 ## 0.48.4
 
 ### New Features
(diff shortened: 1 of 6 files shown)


szokeasaurusrex:
Hi @bardiharborow, thanks for the PR.

Seems like an interesting and potentially useful proposal, but it could take some time before we can give this a proper review.

In the meantime, I would appreciate if you can open an issue linked to this PR, where you explain the concrete problem you were experiencing. I understand that the current solution can be wasteful on hot paths, but it would be helpful to know where you noticed this, and how much improvement these proposed changes generated.

Additionally, it would be super helpful if you can also document any other alternative approaches you considered. I see that you made the commits with Claude, so if you could export the entire Claude session (or a summary of it) and attach it to the issue, that would be very much appreciated. 🙏 That will help us to provide a better, and likely quicker, review.

PR #1236, opened 2026-07-10
test(tracing): Print count and debug output in assertion helpers

lcian:
I encountered the failing test described in https://github.com/getsentry/sentry-rust/issues/1112 again in https://github.com/getsentry/sentry-rust/actions/runs/29080836630/job/86323043895?pr=1214.
Given that that test failure could indicate a problem in the SDK, I'm making this change to the assertion helper to print more information, which will help understand what's going on.

Refs https://github.com/getsentry/sentry-rust/issues/1112
#skip-changelog


szokeasaurusrex (approved):
Nice, thanks! This will be super helpful to debug the flakes

If you get any debug output from this, please post on the issue. We could also include a link to the flakiness issue in the specific test's output with a request to report any debug logs on that issue when the test flakes.

PR #1253, opened 2026-07-21
feat(tower): Record http.response.status_code on HTTP transactions

lcian:
When `SentryHttpLayer` starts a transaction for an incoming request, successful responses now also set the canonical [`http.response.status_code`](https://getsentry.github.io/sentry-conventions/attributes/http/) attribute on transaction data. The existing `SpanStatus` mapping is unchanged; this adds the numeric status so Sentry can surface HTTP response codes the same way other SDKs do.

Only the tower HTTP integration is affected, and only when transactions are enabled.

Context: I was trying to debug something and find traces with a particular status code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-### New Features
+### Features
 
 - `SentryHttpLayer` now records the [`http.response.status_code`](https://getsentry.github.io/sentry-conventions/attributes/http/) attribute on transactions.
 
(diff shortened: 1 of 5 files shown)


lcian:
bugbot run
@sentry review

szokeasaurusrex (approved):
Looks good! Sorry that I missed this for some time. 

I guess we need to resolve some merge conflicts. Would you like to do that, @lcian? Otherwise, I can take care of merging

lcian:
Thanks @szokeasaurusrex, please take care of merging for me, thank you.

PR #1258, opened 2026-07-22
fix(transport): Restore reqwest protocol features

### LINKED ISSUE #1057: Drop or pass through `reqwest` feature `http2`
### Problem Statement

`http2` in `reqwest` / `hyper` is [slow](https://github.com/hyperium/hyper/issues/3071) [and](https://github.com/hyperium/hyper/issues?q=is%3Aissue%20state%3Aopen%20http2) [buggy](https://github.com/seanmonstar/reqwest/issues?q=is%3Aissue%20state%3Aopen%20http2).

We would like to be able to disable `reqwest`'s `http2` feature entirely, but `sentry` makes this impossible if the `reqwest` transport should be used.

### Solution Brainstorm

Either drop feature [from the `reqwest` dependency](https://github.com/getsentry/sentry-rust/blob/5d9787299f3629a6780fd4953f9151941f7d62a2/sentry/Cargo.toml#L78), or pass through an optional flag. (E.g. `sentry` flag `reqwest` for transport `reqwest` without `http2` – and `sentry` flag `reqwest-http` for `reqwest?/http2`.)

lcian:
In https://github.com/getsentry/sentry-rust/pull/998 we upgraded `reqwest` to `0.13.x`.
We moved from `reqwest/default-tls` to `reqwest/native-tls`. The two features are not exactly equivalent, as the second one additionally enables ALPN.

Along with that, we enabled its `http2` feature flag, which was not enabled by default in `0.12.x`.
The original author of that PR enabled `http2` as a doctest was failing.
Apparently it was failing due to this bug in reqwest https://github.com/seanmonstar/reqwest/pull/2927, where enabling only ALPN but not HTTP2 would cause a runtime panic, as the client still advertised HTTP 2 even though it was compiled without support for it.

Enabling HTTP2 is not a problem for `sentry` itself, but it can be if an application crate uses `reqwest` (independently) + `sentry`, as upgrading `sentry` might silently enable HTTP2 on the unified `reqwest` dependency.

This PR restores the previous protocol feature set by selecting `native-tls-no-alpn`, and removing `http2`.
It also bumps `reqwest` to `0.13.2` to get the fix for the aforementioned bug in. Otherwise, if a user has exactly version `0.13.1` as a direct dependency, and enables ALPN but not HTTP2, the reqwest transport will run exactly into that bug, and panic.

Close https://github.com/getsentry/sentry-rust/issues/1057

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@
 - Removed the public `ClientOptions::sample_rate` field. Use `ClientOptions::event_sampling_strategy` to inspect the configured event sampling strategy, and use the existing `ClientOptions::sample_rate(...)` builder setter to configure fixed-rate sampling ([#1228](https://github.com/getsentry/sentry-rust/pull/1228)).
 - Removed the public `ClientOptions::traces_sample_rate` and `ClientOptions::traces_sampler` fields. Use `ClientOptions::traces_sampling_strategy` to inspect the configured traces sampling strategy, and use the existing `ClientOptions::traces_sample_rate(...)` and `ClientOptions::traces_sampler(...)` builder setters to configure fixed-rate and callback-based sampling ([#1227](https://github.com/getsentry/sentry-rust/pull/1227)).
 
+### Fixes
+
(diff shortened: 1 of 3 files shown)


szokeasaurusrex (approved):

lcian:
bugbot run

szokeasaurusrex (approved):

lcian:
bugbot run
@sentry review

PR #1033, opened 2026-03-12
feat: add feedback api

zmitchell:
### Description

Implements the plumbing for the User Feedback API.

#### Issues

* resolves: getsentry/sentry-rust#743

#### Reminders

* Add GH Issue ID *&* Linear ID (if applicable)
* Add an entry to [CHANGELOG.md](<http://CHANGELOG.md>), following the format of existing entries
* The PR title should use [Conventional Commits](<https://develop.sentry.dev/engineering-practices/commit-messages/#type>) style (`feat:`, `fix:`, `ref:`, `meta:`, etc.)
* Useful links for external contributors: [Sentry SDK development docs](<https://develop.sentry.dev/sdk/>), [Discord community](<https://discord.gg/sentry>)

PR #1259, opened 2026-07-22
feat: add feedback API

### LINKED ISSUE #743: Implementing a capture_feedback function
I forked the repo to attempt to implement this function based off of the corresponding function in the Javascript SDK: https://github.com/getsentry/sentry-javascript/blob/136370c8d4f0c0cd9ec80b47ba2292406d0b4ba7/packages/core/src/feedback.ts#L8

From what I gather, the necessary steps are:
- Add a `Feedback` variant to the `Context` enum.
- Add a `FeedbackContext` struct that holds `contact_email`, `name`, and `message`.
- I'm not sure what to do about the `type` field in the Javascript `Event`. I see it in the Go SDK as well for their event type, but I don't see it in the Rust SDK's `Event` type.
- Add the `Context::Feedback` to the `contexts` field of the `Event`
- Call `capture_event` with this new `Event`.

I went ahead and implemented that locally, but it's showing up as an "unlabeled event" (and not displaying in the User Feedback section), even though the "Feedback" context is populated as shown in the screenshot below. I've also checked that the project this event is sent to has User Feedback enabled.

Image

You can see what I've implemented so far here (there's no `capture_feedback` function, but I'm creating the event and calling `capture_event` manually in my own project): https://github.com/zmitchell/sentry-rust/commit/248de09e183481ae8c0186b246aec8fffcf22979

xremming:
Continues the work from [#1033](<https://github.com/getsentry/sentry-rust/issues/1033>).

I personally need this feature so though I might as well finish the partial PR. Largely done together with Claude.

~~Splitted the change of the ~~`~~\~\~type\~\~~~` field from a `~~\~\~String\~\~~~` to en enum into its own commit. Based on my investigations, the whole `type` field is not actually needed.

## Issues

* Resolves getsentry/sentry-rust#743
* Resolves [RUST-32](https://linear.app/getsentry/issue/RUST-32/implementing-a-capture-feedback-function)

### DIFF
--- a/sentry-types/src/protocol/client_report/mod.rs
+++ b/sentry-types/src/protocol/client_report/mod.rs
@@ -87,6 +87,8 @@ indexed_enum! {
         Attachment,
         /// A monitor check-in.
         Monitor,
+        /// A user feedback event.
+        Feedback,
         /// A log item.
         ///
         /// Dropped logs should also be counted as dropped [`LogByte`]s so client reports include
--- a/sentry-types/src/protocol/envelope.rs
+++ b/sentry-types/src/protocol/envelope.rs
@@ -5,10 +5,12 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use uuid::Uuid;
 
+use super::{
+    feedback::{Feedback, FeedbackEvent},
+    v7 as protocol,
+};
 use crate::Dsn;
 use crate::{protocol::v7::ClientReport, utils::ts_rfc3339_opt};
-
-use super::v7 as protocol;
 
 use protocol::{
     Attachment, AttachmentType, ClientSdkInfo, DynamicSamplingContext, Event, Log, Metric,
@@ -135,6 +137,9 @@ enum EnvelopeItemType {
     /// A client report.
     #[serde(rename = "client_report")]
     ClientReport,
+    /// A User Feedback Item type.
+    #[serde(rename = "feedback")]
+    Feedback,
 }
 
 /// An Envelope Item Header.
@@ -188,6 +193,13 @@ pub enum EnvelopeItem {
     ClientReport(ClientReport),
     /// A container for a list of multiple items.
     ItemContainer(ItemContainer),
+    /// A User Feedback item.
+    ///
+    /// Feedback is transmitted as an [`Event`] carrying a `feedback` context, wrapped in a
+    /// [`FeedbackEvent`] which guarantees that context is present. Construct it via
+    /// `EnvelopeItem::from(feedback)`, and use [`EnvelopeItem::as_feedback`] to recover the
+    /// feedback.
+    Feedback(FeedbackEvent),
     /// This is a sentinel item used to `filter` raw envelopes.
     Raw,
     // TODO:
@@ -276,7 +288,19 @@ impl EnvelopeItem {
             Self::MonitorCheckIn(_) => Some(EnvelopeItemType::MonitorCheckIn),
             Self::ClientReport(_) => Some(EnvelopeItemType::ClientReport),
             Self::ItemContainer(container) => Some(container.item_type()),
+            Self::Feedback(_) => Some(EnvelopeItemType::Feedback),
             Self::Raw => None,
+        }
+    }
+
+    /// Returns the [`Feedback`] carried by this item.
+    ///
+    /// Returns `None` for any item that is not a feedback item, or whose wrapped event is missing
+    /// its `feedback` context.
+    pub fn as_feedback(&self) -> Option<&Feedback> {
+        match self {
+            Self::Feedback(feedback) => Some(feedback.feedback()),
+            _ => None,
         }
     }
 }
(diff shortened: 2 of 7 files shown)


sentry-types/src/protocol/v7.rs
```diff
@@ -1730,6 +1734,54 @@ pub struct Event<'a> {
     /// SDK metadata
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub sdk: Option<Cow<'a, ClientSdkInfo>>,
+    /// The event type, which determines how Sentry classifies and routes the event.
+    ///
+    /// Defaults to `None`, which the backend treats as a regular error event. Specific values such
+    /// as [`EventType::Feedback`] route the event to a different product surface.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub r#type: Option<EventType>,
```
xremming: Based on the spec and Claude's opinion, it appears that the `type` field might not actually be required?

---

Question: is the event-body type: "feedback" field needed, or should feedback be identified only by the envelope item header?

Per the User Feedback spec, a feedback is sent as a dedicated envelope item whose item header is {"type":"feedback"}, and the example event payload body carries no top-level type field — the event attributes table lists only event_id, timestamp, and platform as required, and type is not among them.

This implementation currently sets it in both places:
- the envelope item header, via EnvelopeItemType::Feedback (this matches the spec ✅), and
- the event body, via Event.r#type = "feedback" — which is what the EventType enum in the second commit exists to support.

If the backend identifies feedback purely from the envelope item type (as the spec implies), then the event-body type field is redundant and the ref(types): Model the event type field as an enum commit could be dropped, simplifying to_new_event. The original bug report on #1033 claimed the payload requires "type":"feedback" (else it shows as an "unlabeled event"), which contradicts the spec — so I'd like to confirm which is actually true against a live ingest before we keep or remove it.

Spec: https://develop.sentry.dev/sdk/telemetry/feedbacks/

Does the backend rely on the event-body type, or is the envelope item type sufficient?

sentry-types/src/protocol/client_report/mod.rs
```diff
@@ -87,6 +87,8 @@ indexed_enum! {
         Attachment,
         /// A monitor check-in.
         Monitor,
+        /// A user feedback event.
+        Feedback,
```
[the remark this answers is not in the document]
xremming: Claude:

> This is a false positive. Relay's DataCategory::UserReportV2 has canonical name "feedback", not "user_report_v2" — see relay-base-schema/src/data_category.rs: name() returns "feedback" (Self::UserReportV2 => "feedback"), serde serializes it to "feedback", and from_name() accepts both "feedback" and "user_report_v2". The variant is named UserReportV2 internally only to avoid clashing with the legacy UserReport; the wire string is "feedback". So Category::Feedback serializing to "feedback" is correct and outcomes for dropped feedback are recognized.

xremming:
Tested to be working.

image

xremming:
@szokeasaurusrex Since you had reviewed the previous PR would you be the correct person to ping?

sentry-types/src/protocol/envelope.rs
```diff
@@ -188,6 +190,12 @@ pub enum EnvelopeItem {
     ClientReport(ClientReport),
     /// A container for a list of multiple items.
     ItemContainer(ItemContainer),
+    /// A User Feedback item.
+    ///
+    /// Feedback is transmitted as an [`Event`] carrying a `feedback` context, so this variant
+    /// wraps an `Event` rather than a [`Feedback`]. Construct it via `EnvelopeItem::from(feedback)`
+    /// rather than the raw variant, and use [`EnvelopeItem::as_feedback`] to recover the feedback.
+    Feedback(Event<'static>),
```
szokeasaurusrex: Given that Feedback events are meant to have a "feedback" context, I would be in favor of introducing a special FeedbackEvent type which forces the "feedback" context to be included.

This would eliminate the code you have which checks for the presence of a "feedback" context at runtime

sentry-types/src/protocol/v7.rs
```diff
@@ -27,6 +27,7 @@ use crate::utils::{display_from_str_opt, ts_rfc3339_opt, ts_seconds_float};
 pub use self::client_report::Report as ClientReport;
 pub use super::attachment::*;
 pub use super::envelope::*;
+pub use super::feedback::*;
```
szokeasaurusrex: I know we have a lot of existing `*` imports throughout the codebase, but this is really something I would like us to avoid going forward, as it makes it too easy to accidentally introduce unintended public API changes.

Please therefore replace this with an explicit export of only the types which need to be re-exported.

szokeasaurusrex (commented):
Hi, thanks for the PR!

I admittedly have only had time to skim this briefly so far; to do a proper review, I would first need to familiarize myself a bit more with the spec, which would take some time.

In the meantime, I have given some suggestions for simplifying and improving the code. Would appreciate if you can address these 🙏

szokeasaurusrex:
Please also check and address the CI failures before the next review round

xremming:
Forgot to re-request the review. 😅

I think the issues should now be fixed.

szokeasaurusrex:
Hey @xremming, thank you! As I mentioned before, doing a proper review would take some time. It is on my to-do list, but it may take several weeks until I have time to do a full review.

Issue #1127, opened 2026-05-17
Allow overriding sampling rate for events


darrell-roberts:
### Problem Statement

Most event capture will use the sampling rate for the client, however there are some events we consider high priority we want to always send. This would require a user defined function that allows use to set something on the event that we could check for by-passing the sampling rate.

### Solution Brainstorm

Add a user defined function to `ClientOptions` that receives an event and checks if it qualifies for ignoring the configured sampling rate.
Have `prepare_event` check for user defined sampling override and check if sampling rate should be ignored.

darrell-roberts:
Created draft PR: https://github.com/getsentry/sentry-rust/pull/1128

szokeasaurusrex:
Hi @darrell-roberts, thanks for the issue and the draft PR!

Rather than adding a function that returns a boolean to fully ignore the sample rate, I would suggest we instead add support for an [`error_sampler` function, like we have in the Python SDK](https://docs.sentry.io/platforms/python/sampling/#dynamically-sampling-error-events), which returns a floating-point value instead of a boolean. That version allows sampling different events at different sample rates, in addition to supporting your use case. 

However, as adding fields to the `ClientOptions` is a breaking change, we will need to carefully consider if/when we should make this change.

darrell-roberts:
> Hi [@darrell-roberts](https://github.com/darrell-roberts), thanks for the issue and the draft PR!
> 
> Rather than adding a function that returns a boolean to fully ignore the sample rate, I would suggest we instead add support for an [`error_sampler` function, like we have in the Python SDK](https://docs.sentry.io/platforms/python/sampling/#dynamically-sampling-error-events), which returns a floating-point value instead of a boolean. That version allows sampling different events at different sample rates, in addition to supporting your use case.
> 

Hi @szokeasaurusrex 

Thanks. The approach in the python SDK is a better approach.  I've updated the draft PR with the suggestion. 

> However, as adding fields to the `ClientOptions` is a breaking change, we will need to carefully consider if/when we should make this change.

This may be ok, in the sense if no user defined sampling rate override function is provided it falls back to the default behavior.

szokeasaurusrex:
> This may be ok, in the sense if no user defined sampling rate override function is provided it falls back to the default behavior.

[...]

PR #1284, opened 2026-08-08
feat: Allow overriding the sampling rate.

darrell-roberts:
Allow for a user defined function for setting the sampling rate. Similar to Python SDK https://docs.sentry.io/platforms/python/sampling/#dynamically-sampling-error-events

### Description
Allow defining a function that overrides the sampling rate. For example you have a sampling rate set below 100% but you have certain events that should always be 100%. This allows for a user defined function to provide a sample rate override otherwise falls back to the defined sampling rate.

#### Issues

https://github.com/getsentry/sentry-rust/issues/1127

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.49.2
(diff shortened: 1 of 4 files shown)


szokeasaurusrex:
We have decided not to implement this feature at this time. For more information, see https://github.com/getsentry/sentry-rust/issues/1127#issuecomment-5254655840.

PR #1100, opened 2026-04-28
feat: Add WASM support

darrell-roberts:
### Description
This adds WASM support to the `sentry-*` crates.

### New crate: `sentry-time`
- Brand-new internal crate (`sentry-time/`) exposing `now_system_time() -> SystemTime`.
- Non-WASM: delegates to `SystemTime::now()`.
- WASM (`target_arch = "wasm32"`): builds `SystemTime` from `chrono::Utc::now()` (since `SystemTime::now()` panics on `wasm32-unknown-unknown`).
- Added to workspace members in root `Cargo.toml`.

### Replace `SystemTime::now()` call sites
All call sites that produced a current timestamp now go through `sentry_time::now_system_time` (or are `#[serde(default = "sentry_time::now_system_time")]`):
- `sentry-types/src/auth.rs`, `protocol/session.rs`, `protocol/v7.rs` (Breadcrumb, Event, Span, Transaction defaults + `Span::finish`/`Transaction::finish`).
- `sentry-core/src/lib.rs` (re-exports `sentry_time` as `#[doc(hidden)] pub use`), `logger.rs` (4 macro arms now use `$crate::sentry_time::now_system_time()`), `performance.rs`, `session.rs`.
- `sentry-log/src/converters.rs`.

### `Instant` → `chrono` for elapsed time (WASM-incompatible `Instant`)
- `sentry-core/src/batcher.rs`: `Instant`/`std::time::Duration` swapped for `chrono::{DateTime<Utc>, Duration}`. `FLUSH_INTERVAL` now `Duration::seconds(5)`; `last_flush.elapsed()` becomes `Utc::now() - last_flush`; `.checked_sub(...).to_std()` bridges back to `std::time::Duration` for `cvar.wait_timeout`.
- `sentry-core/src/session.rs`: `Session.started: Instant` → `DateTime<Utc>`; flusher loop converted analogously, `chrono::Duration::from_std(FLUSH_INTERVAL)`.

### Platform-gated deps
- `sentry-contexts/Cargo.toml`: `hostname` moved under `cfg(not(target_arch = "wasm32"))`; `uname` gated on `cfg(all(not(windows), not(target_arch = "wasm32")))`.
- `sentry-contexts/src/utils.rs`: `server_name()` returns `None` on WASM; `os_context()` on WASM returns `OsContext { name: Some("WASM"), ..default() }`.

### Cargo updates
- `sentry-core`, `sentry-types`, `sentry-log` add `sentry-time` path dep.
[...]

Issue #226, opened 2020-06-03
Wasm support?


MTRNord:
Hi wasm based rust currently fails in all kind of ways.

Especially this often happens when compiling:

```
error[E0599]: no method named `as_sock` found for reference `&udp::UdpBuilder` in the current scope
    --> /home/marcel/.cargo/registry/src/github.com-1ecc6299db9ec823/net2-0.2.34/src/ext.rs:1558:22
     |
1558 |         get_opt(self.as_sock(), SOL_SOCKET, SO_ERROR).map(int2err)
     |                      ^^^^^^^ method not found in `&udp::UdpBuilder`
```

Is it possible to get wasm support in the future?

Swatinem:
We actually talked about this internally a few days ago. So far we haven’t tried it, and I don’t think it will be a priority any time soon. However, the new modularized `sentry-core` should be a much better starting point for this.

MTRNord:
It would be awesome at some point :)

Diggsey:
Has there been any movement on this issue? :) I see there are some Wasm related types in the documentation.

GeeWee:
I know that leaving +1's are generally a bad practice, but I wanted to mention that particularly now that serving API's via WASM through Fastly or CloudFlare, this is a missing piece that would be _very nice_ to have.

hut8:
I'll be digging through this shortly to see if anything has changed, but has anyone even tried recently? If I add these lines to my wasm project, I don't get any errors: 
```
sentry = {version = "0.29.1", default-features = false}
getrandom = { version = "0.2.8", features = ["js"] }
```

The second line is necessary because `sentry` requires `getrandom` but does not set the `js` feature, which is necessary to get random numbers from a wasm (in JS) context. If I don't set `default-features = false` in `sentry` then the first error I get is about not having OpenSSL headers, and obviously that is not happening in a WASM context. But I see that `reqwest` *does* support WASM, so that stuff shouldn't be necessary...

omarabid:
@hut8 I got it working. However, using the `sentry` package, you have to do all the serializing of data manually (and I think this will be platform specific too, I'm working on CloudFlare workers). If this is of interest to you and you are interested to working on it, reach me by email.

hut8:
It looks like no matter what, I can't use the sentry crate at all, and I believe that's @omarabid 's experience too.  I get a panic here:

```
 at std::time::SystemTime::now::h69200a4ca613d8a5
[...]

PR #1285, opened 2026-08-08
Add WASM support

darrell-roberts:
### Description
The SDK today can not be used with WASM targets. These changes work for my current utilization of the SDK but there are other things that need to change. The transport for example. I have a custom transport for our needs along with a WASM transport. I can add the WASM transport to this PR but for now I'd like to submit this to get started on the discussion/work to get WASM support in the SDK.


szokeasaurusrex:
Hey @darrell-roberts 👋

I would like to reiterate our position from [the previous discussion in #226](https://github.com/getsentry/sentry-rust/issues/226#issuecomment-4335473261) and [my follow-up on #1100](https://github.com/getsentry/sentry-rust/pull/1100#issuecomment-4554375914).

As I explained on #1100, that PR was intentionally kept closed because we do not currently have the capacity to review this work. Since this PR covers substantially the same feature, I am going to close it as well.

More generally, adding WASM support affects several core parts of the SDK. We want to ensure that the resulting implementation is something we can maintain and support, so we plan to investigate, design, and implement a solution ourselves. We are therefore **not accepting external PRs for WASM support**, so please refrain from opening any further WASM-support PRs.

Thanks again for your interest and for sharing your work 🙏

PR #1286, opened 2026-08-10
chore(core): Implement disabled tracing semantics

szokeasaurusrex:
PR https://github.com/getsentry/sentry-rust/pull/1227 enabled us to represent disabled tracing differently from a 0.0 sample rate when the SDK is initialized but left the behavior for the two cases largely the same.

This change fundamentally alters the semantics around how disabled tracing is handled.

Spans/Transactions now keep track of a so-called `TracingState` which keeps track of both whether tracing is enabled and what the sampling decision is. When tracing is enabled, there will always be a sampling decision (including a corresponding sample rate); when tracing is disabled, a sampling decision may be available if the trace was continued from an incoming trace with a sampling decision. Traces started in the SDK when tracing is disabled never have a sampling decision.

We then separate the concepts of sampling decision and whether to capture traces. The sampling decision now can correctly represent the case where the decision has been deferred. Span/Transaction tracing headers will always include this sampling decision, regardless of whetehr the SDK is capturing the corresponding the Span/Transaction. For whether to capture traces, we now use a `FinishAction` with three possibilities: `Send` and `Discard` are both used when tracing is enabled and are used for sampled and unsampled traces, respectively, while `Ignore` is used for all traces when tracing is disabled. Both `Discard` and `Ignore` result in spans/transactions not being sent to Sentry, but only `Discard` sends a client report.

### DIFF
--- a/sentry-core/src/performance/headers.rs
+++ b/sentry-core/src/performance/headers.rs
@@ -132,6 +132,11 @@ impl TracePropagationContext {
         })
     }
 
+    /// Set the `sampled` field, accepting `Option` values.
+    pub(crate) fn with_maybe_sampled(self, sampled: Option<bool>) -> Self {
+        Self { sampled, ..self }
+    }
+
     /// Attempts to construct a [`TracePropagationContext`] from the given Sentry trace header.
     ///
     /// Returns [`None`] if the header cannot be parsed.
--- a/sentry-core/src/performance/mod.rs
+++ b/sentry-core/src/performance/mod.rs
@@ -10,8 +10,11 @@ use sentry_types::protocol::v7::client_report::Reason as ClientReportReason;
 use sentry_types::protocol::v7::OrganizationId;
 use sentry_types::protocol::v7::SpanId;
 
+use self::sampling::TracingState;
 #[cfg(feature = "client")]
 use crate::clientoptions::TracesSamplingStrategy;
+#[cfg(feature = "client")]
+use crate::performance::sampling::FinishAction;
 use crate::{protocol, Hub};
 
 #[cfg(feature = "client")]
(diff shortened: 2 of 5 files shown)


sentry-core/src/performance/mod.rs
```diff
@@ -974,46 +975,42 @@ impl Transaction {
...
+                        let opts = client.options();
+                        transaction.release.clone_from(&opts.release);
+                        transaction.environment.clone_from(&opts.environment);
+                        transaction.sdk = Some(std::borrow::Cow::Owned(client.sdk_info.clone()));
+                        transaction.server_name.clone_from(&opts.server_name);
+
+                        let mut dsc = protocol::DynamicSamplingContext::new()
+                            .with_trace_id(inner.context.trace_id)
+                            .with_sample_rate(sample_rate)
+                            .with_sampled(true);
```
szokeasaurusrex, line 1003: If we are sending the transaction, that means that the transaction is sampled.

szokeasaurusrex:
@giortzisg in case you took at this PR earlier, just a heads up that I have fundamentally changed it.

The previous implementation conflated sampling decision and tracing enabled/disabled by assuming that a deferred sampling decision is equivalent to tracing being disabled. This is true when we are the SDK starting the trace, but if we are continuing an incoming trace, we need to be able to correctly represent that tracing is disabled, but that there is still a sampling decision that needs to be propagated outwards.

sentry-core/src/performance/headers.rs
```diff
@@ -132,6 +132,11 @@ impl TracePropagationContext {
         })
     }
 
+    /// Set the `sampled` field, accepting `Option` values.
+    pub(crate) fn with_maybe_sampled(self, sampled: Option<bool>) -> Self {
```
giortzisg, line 136: l: found the name a bit confusing. Maybe `with_sampled_option` is a better one.

giortzisg (approved):

sentry-core/src/performance/headers.rs
szokeasaurusrex: We've used such a naming pattern elsewhere I believe; it is not a public API, and documentation is clear, so I will keep it like this for now. Can always adjust later

PR #1299, opened 2026-08-13
chore: Deprecate the `enable_logs` option

### LINKED ISSUE #1297: Deprecate and no-op-ify `enable_logs`
We will only partially no-op-ify `enable_logs`; the option should affect integrations, but not logging APIs that are manually called

szokeasaurusrex:
Disabling log sending via `ClientOptions::enable_logs` is somewhat of a footgun, as this flag can prevent all Sentry logs from being sent. This change deprecates the option and restricts its effect: setting `enable_logs` to `false` now only disables automatic log capture by the log-capturing integrations (currently `tracing` and `log`, both with the `logs` feature). Logs captured manually via the logging APIs we provide (`Hub::capture_log` and the `logger_*` macros) are always sent, regardless of the option's value.

We also deprecate the option, as we will remove it in the next breaking release. The migration path is to configure the relevant integration via its own options to not send logs; each integration's documentation describes how to do this.

Resolves [#1297](https://github.com/getsentry/sentry-rust/issues/1297)
Resolves [RUST-274](https://linear.app/getsentry/issue/RUST-274/deprecate-and-no-op-ify-enable-logs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Deprecations
 
-- Deprecated `ClientOptions::enable_logs`. The option is now a no-op; logs are always enabled. To stop sending logs, stop calling the logging APIs or disable the log-capturing integrations (`tracing`, `log`, or `slog`) ([#1299](https://github.com/getsentry/sentry-rust/pull/1299)).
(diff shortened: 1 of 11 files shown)


giortzisg (approved):

szokeasaurusrex:
@giortzisg would appreciate re-review, as I made it now so `enable_logs` is not fully a no-op. Rather, it only gates integration-captured logs, not manually-captured ones

giortzisg (approved):

PR #1301, opened 2026-08-13
ref(client): Make `logs_batcher` non-Option

szokeasaurusrex:
Following the changes in the previous PR, `logs_batcher` is always `Some`. This change makes the field non-`Option` so we don't need to check for a never-occuring `None`.


szokeasaurusrex:
Actually, option is still necessary on close

PR #1315, opened 2026-09-11
feat(minidump): Add `sentry-minidump` integration

### LINKED ISSUE #1316: Add sentry-minidump integration to Rust SDK
Add a `sentry-minidump` crate that captures native crashes in a separate process and uploads minidumps as attachments. Support Linux, macOS, and Windows via the `minidump` feature.

Port the existing standalone [`sentry-rust-minidump`](<https://github.com/timfish/sentry-rust-minidump>) integration, including manual scope synchronization across the crash reporter process boundary.

timfish:
Adds a new `sentry-minidump` crate that captures native crashes as minidumps in a separate process and sends them to Sentry as attachments. Enable it with the `minidump` feature on `sentry`. Linux, macOS and Windows only.

The code is ported from the standalone [`sentry-rust-minidump`](https://github.com/timfish/sentry-rust-minidump) crate.

## How it works

The integration re-executes the current binary as a crash reporter process. The app process spawns the reporter and connects to it; the reporter attaches a native crash handler and waits. On a crash it writes a minidump, attaches it to a `Fatal` event, and uploads it. Scope does not cross the process boundary on its own. To give the crash event context (user, tags, extra, breadcrumbs), the app sends updates to the reporter explicitly through methods on the integration; each call is forwarded over a socket and applied to the reporter's scope.

## Difference from `sentry-rust-minidump`

In `sentry-rust-minidump` the process model is explicit: you call `init(&client)`, which re-executes the binary, and you then have to pass the client in by hand, keep the returned `Handle` alive for the life of the program (or `leak()` it), and know that everything before `init` runs in both processes.

Here the process work moves inside `Integration::setup`, which runs inside `Client::with_options` before `sentry::init` binds the client to the hub. That removes most of the ceremony:

- In the app process, `setup` spawns the reporter and keeps the handle inside the integration, which the client owns for the life of the process. No `Handle`, no `leak()`, no client to pass in.
- In the reporter process, `setup` never returns. It builds its own client from the cloned `ClientOptions`, runs the minidump server loop, and exits. `sentry::init` is the last line of `main` that runs there.

Scope sync is still manual, reached via `sentry::with_integration` instead of a method on a `Handle`. `inherit_args` now defaults to `true` since I suspect this will be more useful to most without causing pain for others.

The one caveat is inherent to re-executing the binary: code before `sentry::init` still runs in both processes. `is_crash_reporter_process()` stays public so apps can gate on that.

Resolves #1316


sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,427 @@
...
+/// hub, then runs the minidump server. `spawn` exits the process with 0 in
+/// server mode. Any error means the reporter could not start, so capture it
+/// and exit with 1 rather than fall through into app code.
+fn run_crash_reporter(child: MinidumperChild, options: &ClientOptions, flush_timeout: Duration) -> ! {
+    let mut reporter_options = options.clone();
+    // Drop this integration from the second client so its setup does not
+    // recurse into the reporter path.
+    reporter_options
+        .integrations
+        .retain(|i| i.as_any().downcast_ref::<MinidumpIntegration>().is_none());
```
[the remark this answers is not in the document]
timfish: Not correct. `i.as_any()` resolves through the `dyn Integration` trait object (via its `AsAny` supertrait), not the `Arc` wrapper, so `downcast_ref::<MinidumpIntegration>()` matches and `retain` drops it. The reporter's second client is built without this integration and never re-enters `setup`, so there's no recursion. The end-to-end test confirms it: the reporter spawns and uploads a minidump.

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,427 @@
...
+
+        static SPAWNED: Once = Once::new();
+        SPAWNED.call_once(|| match child.spawn() {
+            Ok(handle) => {
+                let _ = self.handle.set(handle);
+            }
+            Err(err) => {
+                sentry_debug!("could not start crash reporter: {err}");
+            }
+        });
```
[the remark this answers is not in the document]
timfish: By design. The crash reporter is a process-level resource, started by re-executing the binary, and is spawned once per process, the same way the panic hook is installed once. Re-initializing the client must not fork another reporter, so the `Once` guard is intentional rather than a bug.

sentry-minidump/src/lib.rs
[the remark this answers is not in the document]
timfish: Hardened in 2f5d794: switched to `i.as_ref().as_any()` so the downcast always dispatches through the `dyn Integration` vtable to the concrete type. The previous `i.as_any()` already matched here (because `AsAny` wasn't in scope, so the blanket impl on `Arc` didn't apply), but relying on that was fragile, a stray `use sentry_core::AsAny` would have flipped it to downcast the `Arc` wrapper. This matches how `sentry-core` filters integrations in `Client::with_options`.

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+use std::time::Duration;
+
+use minidumper_child::{ClientHandle, MinidumperChild};
+use sentry_core::protocol::{Attachment, AttachmentType, Breadcrumb, Event, User, Value};
+use sentry_core::{sentry_debug, Client, ClientOptions, Hub, Integration, Level, Scope};
+
+/// The environment variable that marks the crash reporter process.
+///
+/// Change it with [`MinidumpIntegration::server_env_var`].
+pub const DEFAULT_SERVER_ENV_VAR: &str = "_CRASH_REPORTER_SERVER";
```
szokeasaurusrex: l: to make collisions less likely, perhaps prefixing with `SENTRY` makes sense

```suggestion
pub const DEFAULT_SERVER_ENV_VAR: &str = "_SENTRY_CRASH_REPORTER_SERVER";
```

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+
+/// The default time to wait for the crash event to upload.
+pub const DEFAULT_FLUSH_TIMEOUT: Duration = Duration::from_secs(5);
```
szokeasaurusrex: l: Unless we have a strong reason to make these public API, I would rather keep them private. We can document the defaults in the relevant setter methods.

```suggestion
/// The environment variable that marks the crash reporter process.
///
/// Change it with [`MinidumpIntegration::server_env_var`].
const DEFAULT_SERVER_ENV_VAR: &str = "_CRASH_REPORTER_SERVER";

/// The default time to wait for the crash event to upload.
const DEFAULT_FLUSH_TIMEOUT: Duration = Duration::from_secs(5);
```

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+///
+/// This checks [`DEFAULT_SERVER_ENV_VAR`]. If you set a custom name with
+/// [`MinidumpIntegration::server_env_var`], use
+/// [`MinidumpIntegration::is_crash_reporter_process`] instead.
+///
+/// It is safe to call this before the Sentry client exists, for example to
+/// skip log output in the crash reporter process.
+pub fn is_crash_reporter_process() -> bool {
+    std::env::var_os(DEFAULT_SERVER_ENV_VAR).is_some()
+}
```
szokeasaurusrex: m/question: I find it quite confusing that we have two different functions named `is_crash_reporter_process` that do different things. Do we really need both, or could we get rid of one of them to make the API a bit clearer?

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    /// Sets the directory where minidumps are written before upload.
+    ///
+    /// Defaults to `Crashes` inside the system temp directory.
+    #[must_use]
+    pub fn crashes_dir(mut self, dir: impl Into<PathBuf>) -> Self {
+        self.crashes_dir = Some(dir.into());
+        self
+    }
```
szokeasaurusrex, line 165: m: We typically prefer named generics in the SDK.

```suggestion
    pub fn crashes_dir<P>(mut self, dir: P) -> Self
    where
        P: Into<PathBuf>,
    {
        self.crashes_dir = Some(dir.into());
        self
    }
```

In https://github.com/getsentry/sentry-rust/pull/1319, I have enabled a lint to enforce this going forward.

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// environment variable is set after this callback, so `env_clear`
+    /// does not break process detection.
+    #[must_use]
+    pub fn on_process<F>(self, f: F) -> Self
+    where
+        F: FnOnce(&mut Command) + Send + Sync + 'static,
+    {
+        *self.on_process.lock().unwrap() = Some(Box::new(f));
+        self
+    }
```
szokeasaurusrex: l/nit: we can avoid the `unwrap` by just using a new mutex

```suggestion
    pub fn on_process<F>(mut self, f: F) -> Self
    where
        F: FnOnce(&mut Command) + Send + Sync + 'static,
    {
        self.on_process = Mutex::new(Some(Box::new(f)));
        self
    }
```

sentry-minidump/tests/e2e.rs
```diff
@@ -0,0 +1,108 @@
...
+        let _ = tx.send(body);
+    });
+
+    Command::new(env!("CARGO"))
+        .args(["run", "--quiet", "--example", "minidump"])
+        .current_dir(env!("CARGO_MANIFEST_DIR"))
+        .spawn()
+        .expect("spawn example")
+        .wait()
+        .expect("wait for example");
```
szokeasaurusrex: If I understand correctly, this runs `sentry-minidump/examples/app.rs`? 

I do find it slightly confusing that the test is under `examples`. Perhaps this is common practice, and I am just not yet familiar with it, in which case, this is fine. But, I do think it may make sense to either comment this somewhere or to symlink the example into tests.

sentry-minidump/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[dependencies]
+sentry-core = { workspace = true, features = ["client"] }
+minidumper-child = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+
+[dev-dependencies]
+sentry = { workspace = true, features = ["test", "ureq"] }
+sadness-generator = "0.7"
```
szokeasaurusrex: m: Please add this to the workspace dependencies list, as we are trying to unify all dependency definitions there.

sentry-minidump/README.md
szokeasaurusrex, line 1: h: We generate these `README.md` files with `cargo readme` in the version-bumping script, using this command:

https://github.com/getsentry/sentry-rust/blob/2f5d7948812f9eb8f6f1927b5e0bdac7e3d56c64/scripts/generate-readme.sh#L8

The files are generated from the crate-level documentation in `src/lib.rs`. It seems that this `README` contains some information not included in `src/lib.rs`, so you should move that information into `src/lib.rs` so that the `README` survives regeneration.

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+        if let Some(timeout) = self.client_connect_timeout {
+            child = child.with_client_connect_timeout(timeout);
+        }
+        if let Some(timeout) = self.server_stale_timeout {
+            child = child.with_server_stale_timeout(timeout);
+        }
+
+        let inherit_args = self.inherit_args;
+        let process_name = self.process_name.clone();
+        let on_process = self.on_process.lock().unwrap().take();
```
szokeasaurusrex: l: Should be safe to ignore `PoisonError` here

```suggestion
        let on_process = self.on_process.lock().unwrap_or_else(|e| e.into_inner()).take();
```

szokeasaurusrex (commented):
For the most part, this looks good, but I left a couple suggestions I think we should implement before merging.

I tested it out locally and the crashes are showing up nicely in the UI.

sentry-minidump/tests/e2e.rs
timfish: Yes, it runs the example for the test. Added some comments!

I don't think this is common practice, it's just what I came up a while back. The most idiomatic way is probably to add `[[bin]]` target but then `sadness-generator` and `sentry` would need to be moved from dev deps to full dependencies. 

If `sentry-minidump` took a normal dependency on `sentry`, that's a `sentry` ↔ `sentry-minidump` cycle in the normal dependency graph, which Cargo rejects. However, Cargo permits a cycle when an edge is a **dev-dependency** and examples can use dev-dependencies!

The only other way I can think of solving this would be to add a private bin crate simply for the test binary.

timfish:
@szokeasaurusrex do you think we should set the default process name as `sentry crash reporter`, `crash reporter` or similar?

sentry-minidump/tests/e2e.rs
szokeasaurusrex: Ok! I suppose that's fine then if it's documented

szokeasaurusrex:
> @szokeasaurusrex do you think we should set the default process name as `sentry crash reporter`, `crash reporter` or similar?

Yes, I think that's a good idea, but I'd also include some indication that the process is part of the Rust SDK. 

I think something like "Sentry Rust SDK Crash Reporter" or "Crash Reporter (Sentry Rust SDK)" could work well.

sentry-minidump/examples/app.rs
```diff
@@ -0,0 +1,36 @@
+use std::time::Duration;
```
szokeasaurusrex, line 5: l: would also mention that this is part of the end to end test here

```suggestion
//! Example of the `sentry-minidump` crate.
//!
//! This is also executed by the end to end test.

use std::time::Duration;
```

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+//! ```no_run
+//! let _guard = sentry::init((
+//!     "__DSN__",
+//!     sentry::ClientOptions::new().add_integration(
+//!         sentry_minidump::MinidumpIntegration::new()
+//!             .crashes_dir("/var/lib/my-app/crashes"),
+//!     ),
+//! ));
+//! // Only the app process reaches here.
+//! ```
```
szokeasaurusrex: l: we can remove `no_run` by supplying a string that parses as a DSN.

```suggestion
//! ```rust
//! let _guard = sentry::init((
//!     "https://your-dsn@sentry.io/0",
//!     sentry::ClientOptions::new().add_integration(
//!         sentry_minidump::MinidumpIntegration::new()
//!             .crashes_dir("/var/lib/my-app/crashes"),
//!     ),
//! ));
//! // Only the app process reaches here.
//! ```
```

Although I guess it is not that important to run this, because we don't assert anything, I suppose it is still good to run the code and make sure it does not panic

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+//! reporter re-executes the current binary. Build the integration and call
+//! [`MinidumpIntegration::is_crash_reporter_process`] on it to skip work
+//! that should run only in the app process.
+//!
+//! # Scope sync
+//!
+//! Scope changes do not cross the process boundary on their own. Send them
+//! to the crash reporter through the integration:
+//!
+//! ```no_run
```
szokeasaurusrex: l: this example already runs successfully 

```suggestion
//! ```rust
```

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+//! });
+//! ```
+//!
+//! # Platforms
+//!
+//! `sentry-minidump` builds on Linux, macOS and Windows only.
+
+#![doc(html_favicon_url = "https://sentry-brand.storage.googleapis.com/favicon.ico")]
+#![doc(html_logo_url = "https://sentry-brand.storage.googleapis.com/sentry-glyph-black.png")]
+#![warn(missing_docs)]
```
szokeasaurusrex: l: We can remove this; I have centralized the missing_docs lint warning in the workspace in #1321 

```suggestion
```

szokeasaurusrex (approved):
Looks like you addressed my previous feedback 🙏 

I identified a couple more small items but lgtm in general; please ping me when this is ready to be merged

szokeasaurusrex:
@timfish if I understand correctly, after this change is released, the existing `sentry-rust-minidump` crate will continue to exist separately, and this will be released as a new `sentry-minidump` crate?

The one thing I am not sure about is whether Craft will be able to create a new crate when releasing, or if it will error because the crate does not exist yet. I haven't released a new crate at Sentry before so this may be worth looking into

timfish:
> the existing `sentry-rust-minidump` crate will continue to exist separately, and this will be released as a new `sentry-minidump` crate?

At some point I will publish a release of `sentry-rust-minidump` with deprecation and readme notices pointing to the new officially supported integration. 

`sentry-minidump` using the integration API makes `sentry-rust-minidump` redundant!

sentry-minidump/tests/e2e.rs
```diff
@@ -0,0 +1,114 @@
...
+        assert!(n > 0, "connection closed before body end");
+        buf.extend_from_slice(&chunk[..n]);
+    }
+
+    buf[header_end..body_end].to_vec()
+}
+
+#[test]
+fn captures_minidump_from_crash() {
+    let listener = TcpListener::bind(("127.0.0.1", 8123)).expect("bind listener");
```
szokeasaurusrex: This test fails on my local machine at this line:

```
thread 'captures_minidump_from_crash' (2421475) panicked at sentry-minidump/tests/e2e.rs:52:59:
bind listener: Os { code: 48, kind: AddrInUse, message: "Address already in use" }
```

Do you think it would be possible to make this test more robust against the possibility that the address/port is already in use?

sentry-minidump/tests/e2e.rs
```diff
@@ -0,0 +1,140 @@
...
+        .args(["run", "--quiet", "--example", "minidump"])
+        .current_dir(env!("CARGO_MANIFEST_DIR"))
+        .env("SENTRY_DSN", format!("http://dsn@127.0.0.1:{port}/0"))
+        .spawn()
+        .expect("spawn example");
+
+    wait_for_timeout(example_process, Duration::from_secs(5));
+
+    let body = rx
+        .recv_timeout(Duration::from_secs(5))
```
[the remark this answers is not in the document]
szokeasaurusrex: True, forgot about this because it was already cached for me

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+            child = child.with_server_stale_timeout(timeout);
+        }
+
+        let inherit_args = self.inherit_args;
+        let process_name = self.process_name.clone();
+        let on_process = self
+            .on_process
+            .lock()
+            .unwrap_or_else(|e| e.into_inner())
+            .take();
```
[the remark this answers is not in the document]
szokeasaurusrex: @timfish I believe this report is valid, even if reusing an integration across multiple client initializations is rare. 

If it is easy to fix, it might be worth it to do so. One possible solution imo would be to restrict `on_process` to `FnMut`; then we could run the callback on each initialization. Or we could restrict further to `Fn`; I believe that would also allow us to get rid of the mutex lock. 

Is there any reason we need to accept `FnOnce`?

sentry-minidump/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+            }
+            match self.build_child().spawn() {
+                Ok(handle) => {
+                    let _ = self.handle.set(handle);
+                }
+                Err(err) => {
+                    sentry_debug!("could not start crash reporter: {err}");
+                }
+            }
+        });
```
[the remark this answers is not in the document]
szokeasaurusrex: I think this is ok, because we would have already panicked, and multiple initializations are rare

PR #1329, opened 2026-09-22
fix: Use strict trace continuation for `continue_from_span`

szokeasaurusrex:
The function `continue_from_span` allows continuing a distributed trace from a span that is spawned on another thread.

Typically, both spans in this case are from the same Client, but if they are attached to different Clients it is also possible that they are associated with different org IDs.

Anyways, to keep the handling of distributed traces consistent, it makes sense to populate `incoming_trace` in the `continue_from_span` method. As we will be adding more things to `IncomingTrace`, we should be using it consistently.


sentry-core/src/performance/mod.rs
```diff
@@ -271,24 +271,53 @@ impl TransactionContext {
...
+        /// Helper struct for the match statement; this allows us to have `incoming_trace` be
+        /// guarded by the client feature.
+        struct IncomingSpanData {
+            trace_id: TraceId,
+            span_id: SpanId,
+            #[cfg(feature = "client")]
+            incoming_trace: IncomingTrace,
+            sampled: Option<bool>,
+        }
+
```
szokeasaurusrex, line 283: The changes here look more complicated than they really are; we needed to refactor the tuple used in the match statement to a named structure as `incoming_trace` is only available under the `"client"` feature. 

Any other solution I could think of was more complicated

alexander-alderman-webb (approved):

giortzisg:
Is there a reason we introduce an `IncomingSpanData` struct instead of storing the information in the already existing `TracePropagationContext`?

szokeasaurusrex:
> Is there a reason we introduce an IncomingSpanData struct instead of storing the information in the already existing TracePropagationContext? 

`IncomingSpanData` is a local struct just for this function, it cannot be used outside the function. It helps us pass the data through the `match` statement in such a way that allows `IncomingTrace` to be `cfg`-ed out on non-client builds. We don't store anything in it.

It's essentially just replacing the tuple we had been using until now. However, tuples don't allow configuring out a field based on compile-time features.

See also https://github.com/getsentry/sentry-rust/pull/1329#discussion_r4072346888

giortzisg (approved):

